PR #68, opened 2019-10-15
Fix transparency parameter and window size

kHarshit:
* Fixed missing alpha in instance masks of panoptic segmentation output
*(When changing alpha in demo/demo.py, the instance masks' alpha doesn't change otherwise)*
* Add namedWindow to handle the overlage window size

### DIFF
--- a/demo/demo.py
+++ b/demo/demo.py
@@ -13,6 +13,9 @@ from detectron2.utils.logger import setup_logger
 
 from predictor import VisualizationDemo
 
+# constants
+WINDOW_NAME = "COCO detections"
+
 
 def setup_cfg(args):
     # load config from file and command-line arguments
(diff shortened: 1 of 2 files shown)


demo/demo.py
```diff
@@ -92,13 +92,15 @@ def get_parser():
...
+                cv2.namedWindow("COCO detections", cv2.WINDOW_NORMAL)
                 cv2.imshow("COCO detections", visualized_output.get_image()[:, :, ::-1])
                 if cv2.waitKey(0) == 27:
                     break  # esc to quit
     elif args.webcam:
         assert args.input is None, "Cannot have both --input and --webcam!"
         cam = cv2.VideoCapture(0)
         for vis in tqdm.tqdm(demo.run_on_video(cam)):
+            cv2.namedWindow("COCO detections", cv2.WINDOW_NORMAL)
             cv2.imshow("COCO detections", vis)
```
ppwwyyxx: Could you make this string a file-level constant `WINDOW_NAME`?

kHarshit: Okay, I'll do it

PR #177, opened 2019-10-26
Fix a bug not ignoring the IGNORE_VALUE label when visualizing semantic annotation dataset

xiongzubiao:
An IndexError occurred when visualizing the semantic annotations (stuffs) via:
`python tools/visualize_data.py --source annotation --config-file configs/Misc/semantic_R_50_FPN_1x.yaml --show`

It is due to a bug not respecting `cfg.MODEL.SEM_SEG_HEAD.IGNORE_VALUE` in draw_dataset_dict.

### DIFF
--- a/detectron2/utils/visualizer.py
+++ b/detectron2/utils/visualizer.py
@@ -463,14 +463,12 @@ class Visualizer:
 
         return self.output
 
-    def draw_dataset_dict(self, dic, ignore_label=255):
+    def draw_dataset_dict(self, dic):
         """
         Draw semantic segmentation predictions/labels.
 
         Args:
             dic (dict): Metadata of one image, in Detectron2 Dataset format.
-            ignore_label (int): value in semantic segmentation ground truth. The corresponding
-                pixels are not drawn.
 
         Returns:
             output (VisImage): image object with visualizations.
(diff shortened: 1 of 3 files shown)


detectron2/utils/visualizer.py
```diff
@@ -463,7 +463,18 @@ def draw_panoptic_seg_predictions(
 
         return self.output
 
-    def draw_dataset_dict(self, dic):
+    def draw_dataset_dict(self, dic, ignore_label=255):
```
ppwwyyxx: `None` (no labels are ignored) is a more reasonable default

ppwwyyxx: It's not a good idea to add an argument that's only specific to semantic segmentation here.
A reasonable fix would be to modify `draw_sem_seg` so that it does not draw anything that does not have a class name. I think that's already done in this PR as well so the change in this method seems unnecessary.

A even better fix is to include ignored labels as the metadata for semantic segmentation datasets.

xiongzubiao: I was thinking that IGNORE_VALUE can be one of the registered classes. But if this corner case is not something the codebase considers, yeah, filtering by the class name is enough. Let me know how you want to proceed.

I like the idea with metadata. Will you come with a commit to fix this? I am not sure where the best place is to pass  IGNORE_VALUE from cfg to metadata. One place I can think about is in the main function of `visualize_data.py`, but it also looks too specific.

ppwwyyxx: I think only keeping the filtering is the best solution for now. Metadata is attribute of the dataset and should be set when registering the dataset. This can be done in later commits.

xiongzubiao: Cool, I will proceed with this.

PR #197, opened 2019-10-30
'nvidia-docker' has deprecated.

eungbean:
Update comment about docker building.

```nvidia-docker``` has deprecated and was substituted by ```docker (..) --gpus (..) ```.
check [this link](https://github.com/NVIDIA/nvidia-docker).

Thanks for your contribution!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,17 @@
 FROM nvidia/cuda:10.1-cudnn7-devel
 # To use this Dockerfile:
-# 1. `nvidia-docker build -t detectron2:v0 .`
-# 2. `nvidia-docker run -it --name detectron2 detectron2:v0`
+# 1. `docker build -t detectron2:v0 .`
+# 2. `docker run  --gpus all -it --name detectron2 detectron2:v0`
+
+#### Test nvidia-smi with the latest image
+# $ docker run --gpus all detectron2:v0 nvidia-smi
+
+# Start a GPU enabled container on two GPUs
+# $ docker run --gpus 2 detectron2:v0 nvidia-smi
+
+# Starting a GPU enabled container on specific GPUs
+# $  docker run --gpus '"device=1,2"' detectron2:v0 nvidia-smi
+# $ docker run --gpus '"device=UUID-ABCDEF,1"' detectron2:v0 nvidia-smi
 
 
 ENV DEBIAN_FRONTEND noninteractive

### CODE REVIEW

File: Dockerfile [hunk 1]
```diff
@@ -1,7 +1,17 @@
 FROM nvidia/cuda:10.1-cudnn7-devel
 # To use this Dockerfile:
-# 1. `nvidia-docker build -t detectron2:v0 .`
-# 2. `nvidia-docker run -it --name detectron2 detectron2:v0`
```
ppwwyyxx: It seems the new feature is supported only after docker 19.03, which is still quite new. It would be better to keep both old and new instructions.

ahrnbom:
Why not just use Singularity? You get GPU and GUI access out of the box, and it works fine with existing Dockerfiles.

ppwwyyxx:
If it's already working out of the box, you're welcome to contribute additional instructions to use singularity.

ahrnbom:
> If it's already working out of the box, you're welcome to contribute additional instructions to use singularity.

Sure. I couldn't find your Dockerfile on Dockerhub "officially" (hosted by any of the developers here) so I just took one that someone else made. If you want to recommend this strategy to people, I suggest you host it yourself. 

Here's my `detectron2.def` file:
```Bootstrap: docker
From: hakanardo/detectron2

%environment
    export LC_ALL=C
    
%post
    apt-get update --fix-missing && apt-get install -y libcanberra-gtk-module libcanberra-gtk3-module
    
%runscript
    export PROMPT_COMMAND="echo -n \[\ Singularity \]\ "
    exec bash
```

which I run by simply `sudo singularity build detectron2.sif detectron2.def` followed by `singularity run --nv detectron2.sif`. I was able to run the demo with GPU and GUI support.

ppwwyyxx:
I don't think we would spend the extra maintenance effort on dockerhub. 
If there are simple ways to use the `Dockerfile` directly with singularity we would be happy to add them.

ahrnbom:
As far as I am aware, there is no "direct" method for doing this, but rather several options. If you want a low-maintenance method, you can let Docker itself build the image and then let Singularity convert the binary (see https://sylabs.io/guides/3.5/user-guide/singularity_and_docker.html under "Locally Available Images: Cached by Docker") but then you need both Docker and Singularity so I'm not sure it would be considered a "clean" solution. It does remove the need for any hacks for GUI/GPU support, so there's that.

ppwwyyxx:
instructions that use latest docker have been added

PR #225, opened 2019-11-03
setup.py : missing fvcore

jeremyfix:
I was missing fvcore (provided by FAIR)  when pip installing detectron2.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -11,7 +11,6 @@ also installs detectron2 with a few simple commands.
 - [torchvision](https://github.com/pytorch/vision/) that matches the PyTorch installation.
 	You can install them together at [pytorch.org](https://pytorch.org) to make sure of this.
 - OpenCV, needed by demo and visualization
-- [fvcore](https://github.com/facebookresearch/fvcore/): `pip install 'git+https://github.com/facebookresearch/fvcore'`
 - pycocotools: `pip install cython; pip install 'git+https://github.com/cocodataset/cocoapi.git#subdirectory=PythonAPI'`
 - GCC >= 4.9
 
--- a/setup.py
+++ b/setup.py
@@ -79,6 +79,7 @@ setup(
         "matplotlib",
         "tqdm>4.29.0",
         "tensorboard",
+        "fvcore @ git+https://github.com/facebookresearch/fvcore.git"
     ],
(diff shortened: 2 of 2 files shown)


ppwwyyxx:
Thanks for the fix. Could you also remove the mention of fvcore from https://github.com/facebookresearch/detectron2/blob/master/INSTALL.md#requirements, since it will now be automatically installed?

jeremyfix:
@ppwwyyxx done.

ppwwyyxx:
I confirmed that the URL syntax is not supported when running `setup.py` directly (e.g., `python setup.py install`) -- it still searches on pypi instead. I'll revert this change. Feel free to follow up with new PRs if you found the correct syntax for it.

jeremyfix:
You are right ; I gave a try in a [Dockerfile](https://github.com/jeremyfix/detectron2/blob/dockerfileCPU/Dockerfile.cpu) (a CPU dockerfile for this test) and `pip install -e`  works but `python3 setup.py install` fails with a "couldn't find index page...." ; I will have a look to see if I can fix this.

jeremyfix:
@ppwwyyxx :  I think this is is fixed with an example [Dockerfile](https://github.com/jeremyfix/detectron2/blob/dockerfileCPU/Dockerfile.cpu) to be tested on CPU. 
```
git clone https://github.com/jeremyfix/detectron2
cd detectron2
git checkout dockerfileCPU
docker build -t detectron2:cpuv0 -f Dockerfile.cpu .
docker run -it detectron2:cpuv0
sh test.sh
```
And then you can comment/uncomment  in Dockerfile.cpu  to use `pip install -e`


This is still a kind of black magic for me, I'm not familiar enough with the setuptools and pip installations; But it seems that for `setup.py install` to be working, the git links need to be specified in the dependency_links ; These dependency_links are deprecated for [pip install](https://setuptools.readthedocs.io/en/latest/setuptools.html#dependencies-that-aren-t-in-pypi) but that does not seem to be an issue. This way, you can install both using `pip install -e` or `python3 setup.py install`

After some cleanup, keeping only the modified setup.py , does it make sense to create a new pull request ?

ppwwyyxx:
Thanks for digging into this. We would appreciate a new PR!

PR #226, opened 2019-11-03
setup.py missing pycocotools

jeremyfix:
When testing the demo script with : 

> python3 demo/demo.py --config-file configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_3x.yaml  --input sample.jpg --opts MODEL.WEIGHTS detectron2://COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_3x/137849600/model_final_f10217.pkl

it was complaining pycocotools was missing. 

I propose to add to the setup.py script

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,7 +79,8 @@ setup(
         "matplotlib",
         "tqdm>4.29.0",
         "tensorboard",
-        "pycocotools"
+        "cython",
+        "pycocotools @ git+https://github.com/cocodataset/cocoapi.git#subdirectory=PythonAPI"
     ],
     extras_require={"all": ["shapely", "psutil"]},
     ext_modules=get_extensions(),
--- a/setup.py
+++ b/setup.py
@@ -79,6 +79,7 @@ setup(
         "matplotlib",
         "tqdm>4.29.0",
         "tensorboard",
+        "pycocotools"
     ],
     extras_require={"all": ["shapely", "psutil"]},
     ext_modules=get_extensions(),
--- a/setup.py
+++ b/setup.py
@@ -70,6 +70,9 @@ setup(
     "platform for object detection and segmentation.",
     packages=find_packages(exclude=("configs", "tests")),
     python_requires=">=3.6",
+    setup_requires=[
+        "cython"
+    ],
     install_requires=[
         "termcolor>=1.1",
         "Pillow",


ppwwyyxx:
You should simply follow [INSTALL.md](https://github.com/facebookresearch/detectron2/blob/master/INSTALL.md) to install detectron2.

Adding it to setup.py will not work.

ppwwyyxx:
The reason it does not work is because pycocotools from pypi is too outdated. It might work if you could use the github link directly as in #225 .
Another gotcha for pycocotools is that cython is required at build time but it still can install even without cython. If we're to install it automatically we need to check for cython in our setup.py.

jeremyfix:
@ppwwyyxx  : I did the fix as requested but, it seems to me that [pycocotools on pypi](https://pypi.org/project/pycocotools/) is 2.0.0 whichs seems to be the up to date [pycocotools on github](https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/setup.py)

bendidi:
`Cython` is needed to build pycocotools, so you can't have them both in `setup.py` at the same, `Cython` need to installed before installing `pycocotools` (there is installation order in `setuptools` I think)

ppwwyyxx:
pycocotools on pypi has not been updated for a while and is not new enough for our use.

Adding cython as a dependency may not work as expected. If I remember it correctly, cython has to be installed prior to executing `setup.py` to make it take effect. That's why it's perhaps better to `import Cython` as a check, though it is not elegant. I'm not sure what's the best way.

jeremyfix:
@ppwwyyxx  @bendidi , according to the doc, a package can be mentioned in setup_requires so that it gets available for later processing the dependencies, but if we need it to be installed, the package needs to be both in setup_requires and install_requires. 
The excerpt from the [doc](https://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords) when dealing with setup_requires:

> (Note: projects listed in setup_requires will NOT be automatically installed on the system where the setup script is being run. They are simply downloaded to the ./.eggs directory if they’re not locally available already. If you want them to be installed, as well as being available when the setup script is run, you should add them to install_requires and setup_requires.)

I'm actually not sure if getting it into the ./eggs directory is sufficient or if we need to get it installed as well.

RaymondKirk:
In my personal projects I have solved this issue by forking the cocoapi repository and applying the relevant cython/unicode fixes (https://github.com/RaymondKirk/cocoapi/tree/python-api) then in my `setup.py` I can add 

```python
install_requires=[
    # ...
     'pycocotools @ https://github.com/RaymondKirk/cocoapi/archive/python-api.zip',
    # ...
],
```

to install it, perhaps something similar could be done here?

ppwwyyxx:
Closing due to no update.

PR #242, opened 2019-11-05
ensure bbox always remap to original image size

githubgsq:
hi, dear all:
  Sometimes, people may use their own dataset, but they may not set the width and height attributes for dataset_dict, then when func _detector_postprocess_ is called, bbox would map to resized scale instead of the original image size. This PR can help to address this issue.
  Thanks!

### DIFF
--- a/docs/notes/compatibility.md
+++ b/docs/notes/compatibility.md
@@ -49,7 +49,7 @@ The major differences regarding inference are:
 	can improve mask AP on COCO by ~0.5% absolute.
 
 There are some other differences in training as well, but they won't affect
-model-level compatibility. The majors ones are:
+model-level compatibility. The major ones are:
 
 - We fixed a [bug](https://github.com/facebookresearch/Detectron/issues/459) in
   Detectron, by making `RPN.POST_NMS_TOPK_TRAIN` per-image, rather than per-batch.


detectron2/data/dataset_mapper.py
```diff
@@ -73,6 +73,12 @@ def __call__(self, dataset_dict):
         dataset_dict = copy.deepcopy(dataset_dict)  # it will be modified by code below
         # USER: Write your own image loading if it's not from a file
         image = utils.read_image(dataset_dict["file_name"], format=self.img_format)
+        # To ensure bbox always remap to original image size
+        if "width" not in dataset_dict:
+            dataset_dict["width"] = image.shape[1]
+        if "height" not in dataset_dict:
+            dataset_dict["height"] = image.shape[0]
+
         utils.check_image_size(dataset_dict, image)
```
ppwwyyxx: could you implement it inside this function instead?

githubgsq: hi, yuxin, which function do you mean please?

githubgsq: @ppwwyyxx hi, yuxin, new commit is here: https://github.com/facebookresearch/detectron2/pull/249/commits/9374908e079947749d41f0f2c03ba49b910c93b4

PR #262, opened 2019-11-07
Fix exif image orientation

higumachan:
hi.

I tried demo.py our data. 
I found small problem for image have exif information.

I fixed it. please check it. thanks.

### DIFF
--- a/detectron2/data/detection_utils.py
+++ b/detectron2/data/detection_utils.py
@@ -9,7 +9,7 @@ import logging
 import numpy as np
 import torch
 from fvcore.common.file_io import PathManager
-from PIL import Image
+from PIL import Image, ImageOps
 
 from detectron2.structures import (
     BitMasks,
@@ -24,17 +24,6 @@ from detectron2.structures import (
 from . import transforms as T
 from .catalog import MetadataCatalog
 
-convert_image_with_exif = {
-    1: lambda img: img,
-    2: lambda img: img.transpose(Image.FLIP_LEFT_RIGHT),                              # horizontal flip
-    3: lambda img: img.transpose(Image.ROTATE_180),                                   # 180 degree rotate
-    4: lambda img: img.transpose(Image.FLIP_TOP_BOTTOM),                              # vertical flip
-    5: lambda img: img.transpose(Image.FLIP_LEFT_RIGHT).transpose(Image.ROTATE_90),  # horizontal flip & counter clock 90 degree rotate
-    6: lambda img: img.transpose(Image.ROTATE_270),                                   # counter clock wise 270 degree rotate
-    7: lambda img: img.transpose(Image.FLIP_LEFT_RIGHT).transpose(Image.ROTATE_270), # horizontal flip & counter clock wise 270 degree rotate
-    8: lambda img: img.transpose(Image.ROTATE_90),                                    # counter clock wise 90 degree rotate
-}
-
 
 class SizeMismatchError(ValueError):
     """
@@ -45,6 +34,7 @@ class SizeMismatchError(ValueError):
 def read_image(file_name, format=None):
     """
     Read an image into the given format.
+    If image has exif information then rotate and flipping with that information.
 
     Args:
         file_name (str): image file path
@@ -56,9 +46,7 @@ def read_image(file_name, format=None):
     with PathManager.open(file_name, "rb") as f:
         image = Image.open(f)
 
-        exif = image._getexif()
-        orientation = exif.get(0x112, 1)
-        image = convert_image_with_exif[orientation](image)
+        image = ImageOps.exif_transpose(image)
 
         if format is not None:
             # PIL only supports RGB, so convert to RGB and flip channels over below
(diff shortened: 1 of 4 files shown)


detectron2/data/detection_utils.py
```diff
@@ -24,6 +24,17 @@
...
+    1: lambda img: img,
+    2: lambda img: img.transpose(Image.FLIP_LEFT_RIGHT),                              # horizontal flip
+    3: lambda img: img.transpose(Image.ROTATE_180),                                   # 180 degree rotate
+    4: lambda img: img.transpose(Image.FLIP_TOP_BOTTOM),                              # vertical flip
+    5: lambda img: img.transpose(Image.FLIP_LEFT_RIGHT).transpose(Image.ROTATE_90),  # horizontal flip & counter clock 90 degree rotate
+    6: lambda img: img.transpose(Image.ROTATE_270),                                   # counter clock wise 270 degree rotate
+    7: lambda img: img.transpose(Image.FLIP_LEFT_RIGHT).transpose(Image.ROTATE_270), # horizontal flip & counter clock wise 270 degree rotate
+    8: lambda img: img.transpose(Image.ROTATE_90),                                    # counter clock wise 90 degree rotate
+}
+
```
ppwwyyxx: according to https://stackoverflow.com/questions/13872331/rotating-an-image-with-orientation-specified-in-exif-using-python-without-pil-in we can require PIL >=6.0 and use
```
from PIL import ImageOps

image = ImageOps.exif_transpose(image)
```

rbgirshick:
I think there are two cases to consider:

1) Usage of `read_image` from `demo.py`
2) Usage of `read_image` from training and inference code that operates on a dataset

In the case of (1), this change makes sense. But for (2), I am worried that it might cause subtle inconsistencies (e.g., image is rotated while annotations are not) that impact training and/or inference for some datasets. We have to consider both builtin datasets and also unknown datasets that users have started supporting in their own code.

Given this, I'd propose to add an option to `read_image` to support exif-based transformations. The default would be `False` (i.e., ignore exif metadata) and `demo.py` can explicitly set it to `True`. This avoids the BC-breaking change while making `demo.py` more robust.

What do you think?

ppwwyyxx:
For (2) if the dataset has such images and we do not rotate them, there would be more issues to worry about:
1. training on images without respecting its rotation is probably not desired.
2. If the annotations are not rotated either, the annotation tool may need to be fixed to respect rotation.
3. inconsistency with `cv2.imread` and possibly other libraries
 
So we did break compatibility, but for a scenario that's already quite broken. It seems to be a scenario that's not practically good/useful anyway. That's why I tend to think this seems to be a positive bc-breaking change (or call it a bugfix).

Btw in COCO train2017 and VOC I found no images with orientation.

rbgirshick:
I agree that the condition of a dataset is unknown regardless of what we decide to do; for any given dataset, it may or may not have used exif transformations when the annotation when used.

I'm ok with the following:
- Verify that all built-in datasets are ok (i.e., either free of exif transformations, or, if any exist, that the annotations are not negatively impacted by implementing this PR)
- Add documentation about the new default (maybe here https://detectron2.readthedocs.io/tutorials/datasets.html and in the change log?)
- Merge this PR (without applying my previous suggestion)

detectron2/data/detection_utils.py
higumachan: Thanks! I will refactor near future.

higumachan:
@rbgirshick @ppwwyyxx 

Thank you for disscussion for my PR.

In conclusion, how can I merge this PR? (I think only refactor comment from @ppwwyyxx)

ppwwyyxx:
@higumachan just implementing my comment is OK. Besides that could you add the comment that we will rotate the image based on its exif information, in the docstring of `read_image`, as well as in https://github.com/facebookresearch/detectron2/blob/master/docs/tutorials/datasets.md and https://github.com/facebookresearch/detectron2/blob/master/docs/notes/changelog.md?

higumachan:
@ppwwyyxx 

I refactored and add commented docstring, datasets.md  and changelog.md.
Could you check it?

detectron2/data/detection_utils.py
```diff
@@ -34,6 +34,7 @@ class SizeMismatchError(ValueError):
 def read_image(file_name, format=None):
     """
     Read an image into the given format.
+    If image has exif information then rotate and flipping with that information.
```
ppwwyyxx:
```suggestion
    Will apply rotation and flipping if the image has such exif information.
```

docs/tutorials/datasets.md
```diff
@@ -160,3 +160,9 @@ NOTE: For background on the difference between "thing" and "stuff" categories, s
 In detectron2, the term "thing" is used for instance-level tasks,
 and "stuff" is used for semantic segmentation tasks.
 Both are used in panoptic segmentation.
+
+
+### EXIF
+
+If your images have [EXIF](https://en.wikipedia.org/wiki/Exif) information, transform according to that information when loading images.
+When you annotate images having exif information, please you annotate from after transform point space.
```
ppwwyyxx: Please just add `Will apply rotation and flipping if the image has such exif information.` in the doc for `file_name` in line 45.

higumachan: @ppwwyyxx 
I fixed it. thanks.

PR #266, opened 2019-11-07
Initialise setup logger color parameter with string instead of boolean

RaymondKirk:
Allows separate loggers to use the same interface as detectron2 but still be easily distinguishable.

Purely additive PR, extends functionality of default logger behaviour.

### DIFF
--- a/detectron2/utils/logger.py
+++ b/detectron2/utils/logger.py
@@ -6,7 +6,7 @@ import sys
 from collections import Counter
 from fvcore.common.file_io import PathManager
 from tabulate import tabulate
-from termcolor import colored
+from termcolor import colored, COLORS
 
 
 class _ColorfulFormatter(logging.Formatter):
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
Different loggers can be distinguished by the module name.

RaymondKirk:
@ppwwyyxx Very true, however this is purely additive functionality.

Allows you to colour code different loggers easily using pure detectron2 code.

ppwwyyxx:
It only colors part of the message that are originally green.

If you do need such a behavior, you do not have to call `setup_logger` and you can call your own `custom_setup_logger`.

RaymondKirk:
Yes, this is my point. 

Currently if you want to create a logger with the exact same behaviour as the default detectron logger you can call `setup_logger`, this PR just adds additional functionality to that function. In the event you don't need a custom logger but you would like to see the difference between log types you can do so with this PR.

Currently to get this behaviour you have to clone the setup_logger function and change one line. I thought it would be more pythonic to give the colour parameter an actual meaning rather than it being just a flag, feel free to close this PR if you don't agree.

ppwwyyxx:
It is true that you need a clone of the function to change one line. But this is also true for almost every line in every function.
I think we can reconsider this change of API should there be a more prevalent need for it. Now it doesn't seem to be the case to me and there are already ways to achieve the same behavior with some more code.

PR #274, opened 2019-11-08
Detectron2 Model Zoo API (Proposal for #270).

### LINKED ISSUE #270: Torchvision-like model zoo
## 🚀 Feature

How about adding support to initialize standard models (such as Faster-RCNN, Mask-RCNN, RetinaNet etc.) with pre-trained weights officially released with Detectron2 (in `MODEL_ZOO.md`)?

## Motivation

Initializing a model with pretrained weights takes a lot of lines of code:

```python
from detectron2.config import get_cfg
from detectron2.modeling import build_model
from detectron2.checkpoint import DetectionCheckpointer

cfg = get_cfg()
cfg.merge_from_file("config_for_model_i_want_to_use.yaml")

model = build_model(cfg)

# Set model weights as a part of url by looking up MODEL_ZOO.md file.
cfg.MODEL.WEIGHTS = "detectron2://COCODetection/faster_rcnn_R_50_C4_1x/{model_id}/model_final_{commit_sha}.pkl"
DetectionCheckpointer(model).resume_or_load(cfg.MODEL.WEIGHTS, resume=False)
```

I wish to use pre-trained detectors for my project, and detectors are a small part in the whole pipeline. I have my own config system/checkpointer etc. and would want to get this boilerplate code to be reduced.

Apart from a lot of boilerplate code, I need to have the config file downloaded separately, and also keep a track of the URL of model weights (to override `cfg.MODEL.WEIGHTS`).

## Pitch

- `detectron2` as a package should include all the config files. This can be easily done using setuptools. 
- There should be a mapping between config file/model name to S3 URL.

With these two in place, creating a pre-trained model would be much simpler.

To make it fancy, the API can be extended and might look something like:

```python
from detectron2.model_zoo import faster_rcnn_R_50_C4_3x
model = faster_rcnn_R_50_C4_3x(pretrained=True)
```

cc: @jcjohnson

kdexd:
Fixes #270.

### Introducing Model Zoo API:

- All configs in top-level `config` directory will be copied over to `detectron2/model_zoo` during `pip install` or `python setup.py install/develop`.

The API looks like:

```python
import detectron2.model_zoo as d2mz

model = d2mz.get("COCO-Detection/mask_rcnn_R_50_FPN_1x.yaml", trained=True)
```


ppwwyyxx:
These look great! I think what's left uncertain are:
1. ideally, installing detectron2 (e.g. with `pip install`) should not create another directory named "configs" in parallel to one named `detectron2`. But the way the code currently looks for `CONFIGS_DIR` seems to suggest so. I think there are more standard ways to access resource files packaged in a python package, e.g. with `pkg_resources`.

2. `model_zoo.py` should generate function definition (e.g., with `globals()[..]`) for every model to avoid duplication.

3. About the API, `faster_rcnn_R_50_C4_1x` does not include dataset information (it could be trained on COCO or LVIS, for example). I think we need some better way to identify a model.

Would you like to proceed on polishing it or is this just for showing a proof-of-concept?

kdexd:
@ppwwyyxx:  I made a small addition here and kept this as a draft until we converged to a decent solution. Your suggestions are great! I will proceed to polish this. Completely agree on 1 & 2.

About 3, instead of a `pretrained=True` boolean argument, we can:
1. do `pretrained="dataset_name"`.
2. we can simply generate function definitions for each model separately. 

I think for 2, would multiple functions lead to the same model architecture (with `pretrained=False`)? If so, is that an issue?

ppwwyyxx:
`pretrained="dataset_name"` sounds good to me. also cc @rbgirshick to see if there are better suggestions about the API.

detectron2/checkpoint/model_zoo.py
```diff
@@ -1,132 +1,60 @@
...
-    C2_IMAGENET_MODELS = {
-        "MSRA/R-50": "ImageNetPretrained/MSRA/R-50.pkl",
-        "MSRA/R-101": "ImageNetPretrained/MSRA/R-101.pkl",
-        "FAIR/R-50-GN": "ImageNetPretrained/47261647/R-50-GN.pkl",
-        "FAIR/R-101-GN": "ImageNetPretrained/47592356/R-101-GN.pkl",
-        "FAIR/X-101-32x8d": "ImageNetPretrained/20171220/X-101-32x8d.pkl",
-        "FAIR/X-101-64x4d": "ImageNetPretrained/FBResNeXt/X-101-64x4d.pkl",
-        "FAIR/X-152-32x8d-IN5k": "ImageNetPretrained/25093814/X-152-32x8d-IN5k.pkl",
-    }
+MODEL_ZOO_FUNCTION_STRING = """
```
ppwwyyxx: rather than using exec and string manipulation, you can use a function creator which takes some arguments and returns a function to be used as model zoo function.

kdexd:
I had to include the config directory within the package and gitignore it (during `setup.py develop/install` or `pip install .`) to keep things clean. The API now looks like this:

image

detectron2/checkpoint/model_zoo.py
kdexd: Great idea! Making this modification.

rbgirshick:
> `pretrained="dataset_name"` sounds good to me. also cc @rbgirshick to see if there are better suggestions about the API.

This looks good to me as well. While catching up on the thread, I was going to suggest the same.

Minor question / suggestion: should we make the import accessible like `from detectron2.model_zoo import ...` (i.e., removing `.checkpoint`)? If this API will mainly be used by high-level detection consumers then the notion of a "checkpoint" may be strange/foreign when simply wanting to load a detectron2 model zoo model, so removing it might be a cleaner interface. But this is very minor.

setup.py
```diff
@@ -69,6 +90,7 @@ def get_extensions():
     description="Detectron2 is FAIR's next-generation research "
     "platform for object detection and segmentation.",
     packages=find_packages(exclude=("configs", "tests")),
+    data_files=get_model_zoo_configs(exclude=["quick_schedules", "Detectron1-Comparisons"]),
```
ppwwyyxx: I tried `package_data={"detectron2": ["checkpoint/configs/*.yaml", "checkpoint/configs/**/*.yaml"]},` and it works as expected, installing the config files to `detectron2/checkpoint/configs` after installation. It's much easier since the directory structure inside `checkpoint/configs/` will be kept automatically.

setup.py
```diff
@@ -61,6 +59,29 @@ def get_extensions():
     return ext_modules
 
 
+def get_model_zoo_configs(exclude: Sequence[str] = []) -> Sequence[Tuple[str, Sequence[str]]]:
```
ppwwyyxx: With `package_data=...`, I found that a symlink suffices, and no need to copy. How about just creating one symlink and copy only when this fails (e.g. on windows or some windows file system)?
It will then also include configs for quick_schedules, etc, but that should be fine.

setup.py
```diff
@@ -12,8 +14,7 @@
 
 
 def get_extensions():
-    this_dir = os.path.dirname(os.path.abspath(__file__))
-    extensions_dir = os.path.join(this_dir, "detectron2", "layers", "csrc")
+    extensions_dir = os.path.join("detectron2", "layers", "csrc")
```
ppwwyyxx: does this cause any issue? seems working to me locally

kdexd: Aah yes, this caused a problem while using `MANIFEST.in` file and using `include_package_data=True`>.

I used an alternative approach, will revert this back. Good catch, thanks!

setup.py
kdexd: What if I `pip install .` and delete the repo later? I haven't tried this but it might fail.

Not sure if symlink gets occupied to `site-packages` or the actual file. Will try and rrport back.

setup.py
kdexd: Right, this hack was also from a previous trial where configs were still out of the package after build. Will fix.

kdexd: Did you mean to suggest using `package_data` instead of `data_files` or this exact snippet?

Because I haven't considered Detectron1-comparisons and quick_schedules as a part of Model Zoo.

setup.py
ppwwyyxx: When using `package_data` I found that the symlink'ed directory gets copied during installation.

setup.py
ppwwyyxx: Yes. Having the extra configs doesn't hurt. In fact, our internal end-to-end tests can benefit from having quick_schedules as part of the package.

kdexd:
I have added some documentation to generate sphinx docs. Let me know if anything can be added. I think this is a good aid to see all the functions (generated symbols) at one place, along with supported values of input arguments.

Only one model supports two "string" values of `pretrained` for now, so I added a `pretrained=True` support for flexibility.

image

image

docs/requirements.txt
```diff
@@ -13,5 +13,3 @@ tabulate
 cloudpickle
 Pillow
 git+git://github.com/facebookresearch/fvcore.git
-https://download.pytorch.org/whl/nightly/cpu/torch-1.3.0.dev20191010%2Bcpu-cp37-cp37m-linux_x86_64.whl
```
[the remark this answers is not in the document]
kdexd: I'm sorry, didn't mean to commit this. I was using my personal laptop to build docs and had limited bandwidth to download a bulky library like PyTorch. Will add it back.

kdexd:
Thanks for the detailed review! I am going to address these minor concerns and collapse the diffs, so we can focus on the higher level discussion. These points are great, I think they will improve the API significantly once we have converged on them.

kdexd:
Worked on all the suggestions and resolved merge conflict. Now `detectron/checkpoint/catalog.py` is back to what it was on `master` before. All the discussions here would potentially alter the content of `ModelZooUrls` class.

I think one solution can be to have a mapping like:
```
MODEL_NAME: { "coco": URL1, "lvis": URL2 ...}
```

We can explicitly specify the members and control the symbols in our API, this fixes:

1. `pretrained="coco"` instead of `pretrained="COCO-Detection` or something.
2. We can only release `3x` models whenever available and have symbol (function) names without `_1x`, `_3x` etc. So we could have `faster_rcnn_R_50_C4_3x(pretrained="coco or pascalvoc")`
3. We can consider using `def model_name(dataset: str =..., load_weights: bool = ...)`.

We will not need to read all configs during load, and explicity written mapping helps readers to know what's offered and what's not.

kdexd:
> 3. We can consider using def model_name(dataset: str =..., load_weights: bool = ...).

Also, in cases where one would not care which dataset (for example, one only needs to rip out the backbone and not do anything with `num_classes`), we can provide a module level function like `def get_supported_datasets(model_name: str)`. After this, `dataset=get_supported_datasets(model_name)[0]` should do.

rbgirshick:
My preference is for an API that requires specifying a config file, like:

```python

model = model_zoo.get("COCO-InstanceSegmentation/mask_rcnn_R_101_DC5_3x.yaml", trained=True)
```

This doesn't have the syntactic sugar of the current design, but in my opinion that adds little value (other than maybe looking cool) and leads to tricky decisions about how to map call to a config file. I find specifying the exact config file to be the clearest, most unambiguous option. It has no schema about datasets and tasks, which I view as a plus, because who knows what the future will bring? (E.g., if we define a schema and then I train a model on some strange combination of datasets and tasks, it might break the schema.)

Another issue that arises when thinking about this. The `pretrained` flag can be ambiguous in other ways. When specifying a config, the number of classes is now known, but does `pretrained=False` mean random init for the whole network or init with the weights in the config (likely from ImageNet pre-training)? One option is to call this `trained=` and if it's `True` then we load the model listed in the model zoo .md on github, otherwise we return a model with random weights (not ImageNet or otherwise specified by the config). Other options are possible.

kdexd:
> This doesn't have the syntactic sugar of the current design, but in my opinion that adds little value (other than maybe looking cool) and leads to tricky decisions about how to map call to a config file.

Completely agree, losing the syntactic sugar is okay to have a simple and unambiguous API. I think it still reduces a lot of friction for external users:
- Now the config files are a part of the package (upon pip/setup.py install), so users need not keep a copy of the config files separately - one can remove the github repo after installing detectron2 as a package in their conda env, for example.
- Also, users need not keep a config->weights URL (S3) mapping at their end, this is hidden behind the detectron2 curtain.

Also, this still reduces a lot of code to initialize a model, as I mentioned in #270.

> I find specifying the exact config file to be the clearest, most unambiguous option.

There are some drawbacks, affect external users more due to less familiarity with detectron2, but they might still be okay:

1. They will have to change their function calls when the config files are renamed/re-arranged, which is similar to having directory name as a part of the API call.
  - It would help to specify the exact input string in `MODEL_ZOO.md`, and upgrade detectron2 (semantic) version even on renaming/re-arranging config files. I am not sure if this is already done, or the configs have not been changed since the release.. but it is worth noting that while `configs` directory lies outside the source directory, it is _still_ a part of the package. So, two versions of detectron2 may have no changes in the source code, but different contents in `configs` directory.

2. Not all config files would have their model weights released (?), like `quick_schedules`, and it feels a bit counter-intuitive that `model_zoo.get()` errors for some file under `configs directory.
  - It would help to have a function return a list of possible values which `.get()` can accept. It can also be used to display the error message.

> It has no schema about datasets and tasks, which I view as a plus, because who knows what the future will bring? (E.g., if we define a schema and then I train a model on some strange combination of datasets and tasks, it might break the schema.)

I would still suggest on adding very few standard models trained on standard datasets as a part of the model zoo, for example Faster R-CNN, Mask R-CNN, Keypoint-RCNN trained on COCO. These are commonly compared against, used as baselines, and more recently as feature extractors in several vision and language models (especially Faster R-CNN, the common one is trained on Visual Genome, still). Detectron2 as a library need not support **all** the models with the syntactic sugar, and `model_zoo.get()` should be encouraged otherwise. Likewise, any model not in the model zoo (with the fancy API call) should only be added if it has gained enough popularity already — this might have been under discussion for adding Mask R-CNN if today was sometime in late 2017-early 2018 :-)

> Another issue that arises when thinking about this. The `pretrained` flag can be ambiguous in other ways. ... One option is to call this `trained=` ...

I like your suggestion. Another suggestion would be to draw inspiration from [torchvision](https://pytorch.org/docs/stable/torchvision/models.html#torchvision.models.detection.fasterrcnn_resnet50_fpn). There are `pretrained` and `pretrained_backbone` flag, where `pretrained_backbone` is ignored when `pretrained=True`.

ppwwyyxx:
I feel that having the configs packaged is a good feature. I can imagine other use cases other than using the pre-trained models.

As for using the pre-trained models, I also feel that `model_zoo.get(config_name)` is the most clear and unambigous API and the model name is more like syntax sugar. I think we can start from this API first. Adding the syntax sugar (and giving them stable APIs regardless of config names) for a few well-established models sounds reasonable as well, we can add them after the underlying API is ready.

kdexd:
Alright, it looks like we have converged to a solution. This seems like a good closure on the scope of this PR. More can follow later.. making necessay changes now.

kdexd:
@ppwwyyxx @rbgirshick : Tested this API, works fine for me!

detectron2/model_zoo/model_zoo.py
```diff
@@ -0,0 +1,104 @@
...
+
+def get(config_path, trained: bool = False):
+    """
+    Get a model specified by relative config path.
+
+    Parameters
+    ----------
+    trained: bool, optional (default = False)
+        Whether to initialize with pre-trained weights of full model. If False, only the backbone
+        is initialized with ImageNet pre-trained weights.
```
rbgirshick: > "initialized with ImageNet pre-trained weights"

This is typically true, but more precisely the model will be initialized based on the setting of `MODEL.WEIGHTS`. I recommend changing to something like:

"Whether to initialize with the trained model zoo weights. If False, the initialization weights specified in the config file's `MODEL.WEIGHTS' key are used instead; this will typically (though not always) initialize a subset of weights using an ImageNet pre-trained model, while randomly initializing the other weights."

kdexd: Sounds much better, thanks!

kdexd:
Rebased with latest `master`.

detectron2/model_zoo/model_zoo.py
```diff
@@ -0,0 +1,104 @@
...
+    def get(cls, config_path):
+        if config_path in cls.CONFIG_PATH_TO_URL_SUFFIX:
+            name = config_path.replace(".yaml", "")
+            return cls.S3_PREFIX + name + "/" + cls.CONFIG_PATH_TO_URL_SUFFIX[config_path]
+        raise RuntimeError("{} not available in Model Zoo!".format(name))
+
+
+def get(config_path, trained: bool = False):
+    """
+    Get a model specified by relative config path.
```
ppwwyyxx: may need to clarify that it's "relative config path w.r.t. detectron2's official `configs` directory" to avoid confusion (it cannot be users' own config file). Or perhaps provide an example as well.

detectron2/model_zoo/model_zoo.py
```diff
@@ -0,0 +1,104 @@
...
+
+    Parameters
```
ppwwyyxx: We use the style like this (https://sphinxcontrib-napoleon.readthedocs.io/en/latest/)

```
Args:
    trained (bool): Whether to ...
```

setup.py
```diff
@@ -79,6 +108,7 @@ def get_extensions():
         "matplotlib",
         "tqdm>4.29.0",
         "tensorboard",
+        "fvcore @ git+https://github.com/facebookresearch/fvcore.git",
```
ppwwyyxx: remove this since this syntax does not always work (and should be addressed by #304)

kdexd:
For adding a test, we will have to add a config file and make it a part of the Model Zoo, right? Would that be okay? 

I am looking at options to monkeypatch the configs directory, but it doesn't seem to be too flexible with unittest.

PR #304, opened 2019-11-12
Setuptools dependencies

jeremyfix:
This is a follow-up of the pull requests #225 and #225 that can be closed if this PR is accepted. 

This PR installs the dependencies within the setup.py for both fvcore and pycocotools. 

I tested it and the installation (seems to (see below)) work when installing with `pip install -e` or `python3 setup.py install` . The Dockerfile and the documentation are modified accordingly.

For testing the modification, I did the following :
```
nvidia-docker system prune -af
nvidia-docker build -t detectron2:v0 .
nvidia-docker run -it --name detectron2 detectron2:v0
wget http://images.cocodataset.org/val2017/000000439715.jpg -O input.jpg
mkdir outputs
python3 demo/demo.py  ......  <-- as mentioned at the end of the dockerfile
```
I wanted to ensure nothing was cached by docker before testing. In both `pip install -e` and `setup.py install`, the demo.py script ran and produced an image as outputs/input.jpg without raising errors; the scripts says it detected 15 instances. I did not take the time to check the image though (I need to find the command to get it out from the docker image).

**Strange things remain** 

with `pip install -e`, during the installation, [this linked error](https://github.com/cocodataset/cocoapi/issues/172) is raised which, according to the link, indicates cython is not installed. However, few lines later, the script is saying : 

`Successfully installed PyYAML-5.1.2 absl-py-0.8.1 .... cython-0.29.14 detectron2 fvcore-0.1 google-auth-1.7.0 google-auth-oauthlib-0.4.1 ... pycocotools-2.0 .. yacs-0.1.6`

and the demo script executes without complaining.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,13 +1,3 @@
-[build-system]
-; Handling the cython dependency for working both with pip and setup.py install 
-requires = ['cython']
-
-[options]
-; For broader compatibility, cython is also mentioned in setup_requires
-; see https://setuptools.readthedocs.io/en/latest/setuptools.html#distributing-extensions-compiled-with-cython
-setup_requires=
-	cython
-
 [isort]
 line_length=100
 multi_line_output=4
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,10 @@
 [build-system]
+; Handling the cython dependency for working both with pip and setup.py install 
 requires = ['cython']
 
 [options]
+; For broader compatibility, cython is also mentioned in setup_requires
+; see https://setuptools.readthedocs.io/en/latest/setuptools.html#distributing-extensions-compiled-with-cython
 setup_requires=
 	cython
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,10 @@
+[build-system]
+requires = ['cython']
+
+[options]
+setup_requires=
+	cython
+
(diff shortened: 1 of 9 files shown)


ppwwyyxx:
Looks great. Regarding the cython issue, it seems like the cython dependency may be required before it is installed. It's perhaps best to still install cython manually.

jeremyfix:
Hummm.... you might be right indeed.

It is probably that pycocotools gets installed but without being cythonized I suppose.

And I read that [pip does not use setup_requires](https://pip.pypa.io/en/stable/reference/pip_install/#controlling-setup-requires) but [cython can be mentioned in setup.cfg](https://setuptools.readthedocs.io/en/latest/setuptools.html#distributing-extensions-compiled-with-cython)

I will have a look to this.

setup.cfg
```diff
@@ -1,3 +1,10 @@
+[build-system]
```
ppwwyyxx: thanks for getting a solution. I suppose now it's finally working? Please include a reference to https://setuptools.readthedocs.io/en/latest/setuptools.html#distributing-extensions-compiled-with-cython.

jeremyfix: Ok ; I'm not full time on it; I still need to test it. I'll let you know when this is done.

PR #318, opened 2019-11-14
display filename on image size mismatch

yanicklandry:


ppwwyyxx:
Thanks. It makes sense to print more information so user knows which image has such issue, however:
1. This changes the API. We should make it default to None, and when it's None it will not print the information.
2. It assumes there is a "file name" for every image, but this may not be true. The image may be accessed by other identifications (e.g. a key of a key-value store, etc) in users' own data loaders. It would be better to call it "name" instead.

yanicklandry:
@ppwwyyxx : Thanks for your comments, I changed it.

ppwwyyxx:
Ah, sorry but I just realized that `dataset_dict` is already given to the function. So you can actually just check whether it has "file_name" key and print the information. This would make the API simpler and achieve the same functionality for most use cases. Could you make this change instead?

Also please run `./dev/linter.sh` prior to commit. It should unify the coding style.

yanicklandry:
@ppwwyyxx : Sorry about that, I fixed it. I linted only the file I changed, I can lint the rest in a separate pull request if you prefer.

detectron2/data/detection_utils.py
```diff
@@ -73,7 +73,9 @@ def check_image_size(dataset_dict, image):
         expected_wh = (dataset_dict["width"], dataset_dict["height"])
         if not image_wh == expected_wh:
             raise SizeMismatchError(
-                "mismatch (W,H), got {}, expect {}".format(image_wh, expected_wh)
+                "mismatch (W,H), got {}, expect {} with name {}".format(
+                    image_wh, expected_wh, dataset_dict["file_name"]
+                )
```
ppwwyyxx:
```
    name = dataset_dict.get("file_name")

    ....

                "Mismatched (W,H){}, got {}, expect {} with name {}".format(
                    " for image " + name if name is not None else "", image_wh, expected_wh
                )
```
so that it handles cases of no file_name.

yanicklandry:
@ppwwyyxx : It should handle the absence of file_name now and it's also linted.

yanicklandry:
Can I know which test is failing so that I can help ?

PR #323, opened 2019-11-15
linted codebase #322

yanicklandry:
Related to issue : #322


ppwwyyxx:
Thanks for fixing these. However some of the changes disagree with the linters we use. It seems related to different behavior in different versions of `black`. I'll take a look at how to make linter more consistent.

yanicklandry:
@ppwwyyxx : Thanks, do you plan in using a requirements.txt or Pipfile ? That would solve the different versions problem.

Issue #349, opened 2019-11-19
Saving data dict to COCO format


donnydonny123:
Thanks for the excellent work! I've encountered some issues around saving my own annotations into COCO format.

## How To Reproduce the Issue
``` python
from detectron2.data.datasets.coco import convert_to_coco_json
from detectron2.data import MetadataCatalog, DatasetCatalog
def data_dict():
    return [{
        'file_name': 'image.png',
        'image_id': 'image',
        'height': 100,
        'width': 100,
        'annotations': [{
            'bbox': [70, 30, 100, 70],
            ## area should be 30*40=1200
            'bbox_mode': BoxMode.XYXY_ABS,
            'category_id': 0,
            'iscrowd': 0,
        }]
    }]

def to_json():
    DatasetCatalog.register('test', data_dict)
    MetadataCatalog.get('test').set(thing_classes=["first"])
    convert_to_coco_json('test', output_folder='./output', allow_cached=False)

if __name__ == '__main__':
    to_json()
```
After running the script above, in `./output/test_coco_format.json` I get: (formatted)
```json
{
    "info": {
        "date_created": "2019-11-19 23:23:08.401029",
        "description": "Automatically generated COCO json file for Detectron2."
    },
    "images": [
        {
            "id": "image",
            "width": 100,
            "height": 100,
            "file_name": "image.png"
        }
    ],
    "annotations": [
        {
            "id": 1,
            "image_id": "image",
            "bbox": [
                70.0,
                30.0,
                30.0,
                40.0
            ],
            "area": -400.0,
            "category_id": 0,
            "iscrowd": 0
        }
    ],
    "categories": [
        {
            "id": 0,
            "name": "first"
        }
    ],
    "licenses": null
}
```
## Expected behavior
The area in annotation should be positive (1200)
I think the problem is [here](https://github.com/facebookresearch/detectron2/blob/master/detectron2/data/datasets/coco.py#L334), but I'm not sure how to correct it. The `.area()` function isn't considering its BoxMode.  
BTW, Is saving my own annotation into COCO format and registering it by `register_coco_instances` the best practice when using `COCO_Evaluator`?  

## Environment
```
------------------------  --------------------------------------------------
sys.platform              linux
Python                    3.7.3 (default, Mar 27 2019, 22:11:17) [GCC 7.3.0]
Numpy                     1.17.4
Detectron2 Compiler       GCC 7.4
Detectron2 CUDA Compiler  10.0
[...]

PR #355, opened 2019-11-20
Fix area calculation error in convert_to_coco_dict

donnydonny123:
Fixes: #349  

The constructor of `Boxes`  needs the bbox in `BoxMode.XYXY_ABS` mode.
This error will cause incorrect area calculation in `convert_to_coco_dict`.   
I added a line converting the original bbox to XYXY_ABS before constructing `Boxes`.

### DIFF
--- a/detectron2/data/datasets/coco.py
+++ b/detectron2/data/datasets/coco.py
@@ -331,7 +331,8 @@ def convert_to_coco_dict(dataset_name):
                 area = polygons.area()[0].item()
             else:
                 # Computing areas using bounding boxes
-                area = Boxes([bbox]).area()[0].item()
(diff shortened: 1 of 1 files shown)


detectron2/data/datasets/coco.py
```diff
@@ -331,7 +331,8 @@ def convert_to_coco_dict(dataset_name):
                 area = polygons.area()[0].item()
             else:
                 # Computing areas using bounding boxes
-                area = Boxes([bbox]).area()[0].item()
+                bbox_xy = BoxMode.convert(bbox, bbox_mode, BoxMode.XYXY_ABS)
```
ppwwyyxx: this is incorrect. Since `bbox` already becomes `XYWH_ABS` mode due to line 323, the "from mode" in this line should be "XYWH_ABS", instead of `bbox_mode`.

PR #363, opened 2019-11-21
Updated proposal loading to avoid sorting

bradezard131:
Updated the code and docstring for loading precomputed proposals. The docstring now mentions the format of the proposal dict so that it is easier for users to add their own region proposal files. The loading code no longer requires the proposal dict be sorted in the same order as the dataset.

### DIFF
--- a/detectron2/data/build.py
+++ b/detectron2/data/build.py
@@ -103,6 +103,13 @@ def load_proposals_into_dataset(dataset_dicts, proposal_file):
     """
     Load precomputed object proposals into the dataset.
 
+    The proposal file should be a pickled dict[str->list] with the following
+    keys:
+    - "image_id": list[int], the image ids
+    - "boxes": list[np.ndarray], an Nx4 array of boxes corresponding to the ids
+    - "objectness_logits": list[np.ndarray], a N sized array of objectness scores
+    - "bbox_mode": the BoxMode of the boxes array
+
     Args:
         dataset_dicts (list[dict]): annotations in Detectron2 Dataset format.
         proposal_file (str): file path of pre-computed proposals, in pkl format.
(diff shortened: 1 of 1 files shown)


detectron2/data/build.py
```diff
@@ -103,6 +103,13 @@ def load_proposals_into_dataset(dataset_dicts, proposal_file):
     """
     Load precomputed object proposals into the dataset.
 
+    The proposal file should be a pickled dict[str->list] with the following
+    keys:
+    - "image_id": list[int], the image ids
```
ppwwyyxx: should be "ids" instead of "image_id"

detectron2/data/build.py
```diff
@@ -103,6 +103,13 @@ def load_proposals_into_dataset(dataset_dicts, proposal_file):
...
+    - "boxes": list[np.ndarray], an Nx4 array of boxes corresponding to the ids
+    - "objectness_logits": list[np.ndarray], a N sized array of objectness scores
+    - "bbox_mode": the BoxMode of the boxes array
```
ppwwyyxx:
```suggestion
    - "bbox_mode": the BoxMode of the boxes array. Defaults to ``BoxMode.XYXY_ABS``
```

detectron2/data/build.py
```diff
@@ -103,6 +103,13 @@ def load_proposals_into_dataset(dataset_dicts, proposal_file):
     """
     Load precomputed object proposals into the dataset.
 
+    The proposal file should be a pickled dict[str->list] with the following
```
ppwwyyxx:
```suggestion
    The proposal file should be a pickled dict with the following
```
bbox_mode is not a list

detectron2/data/build.py
```diff
@@ -103,6 +103,13 @@ def load_proposals_into_dataset(dataset_dicts, proposal_file):
...
+    keys:
+    - "image_id": list[int], the image ids
+    - "boxes": list[np.ndarray], an Nx4 array of boxes corresponding to the ids
```
ppwwyyxx:
```suggestion
    - "boxes": list[np.ndarray], each is an Nx4 array of boxes corresponding to the image id
```

detectron2/data/build.py
```diff
@@ -103,6 +103,13 @@ def load_proposals_into_dataset(dataset_dicts, proposal_file):
...
+    - "objectness_logits": list[np.ndarray], a N sized array of objectness scores
```
ppwwyyxx:
```suggestion
    - "objectness_logits": list[np.ndarray], each is an N sized array of objectness scores corresponding to the boxes
```

PR #372, opened 2019-11-22
Save the planet by not reading images for size

botcs:
```
In [7]: %timeit Image.open("datasets/coco/train2017/000000000009.jpg").size                           
34.9 µs ± 59.6 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [8]: %timeit imagesize.get("datasets/coco/train2017/000000000009.jpg")                             
9.43 µs ± 64.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

https://images.app.goo.gl/Msptv7xcCVsXJ5zh8

### DIFF
--- a/detectron2/data/datasets/coco.py
+++ b/detectron2/data/datasets/coco.py
@@ -6,6 +6,7 @@ import os
 import datetime
 import json
 import numpy as np
+import imagesize
 
 from PIL import Image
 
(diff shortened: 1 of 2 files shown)


detectron2/data/datasets/coco.py
```diff
@@ -262,12 +263,10 @@ def file2id(folder_path, file_path):
 
     dataset_dicts = []
     for (img_path, gt_path) in zip(input_files, gt_files):
+        w, h = imagesize.get(gt_path)
```
bendidi: with this we probably lose the flexibility of `PathManager` ?
For me for example I've registred multiple new handlers for PathManager and my images are all online.

maybe add `local_path = PathManager. get_local_path(gt_path)` before ?

botcs: well that sounds good :)

botcs: @bendidi Updated in a31f70b

ppwwyyxx:
Could you add `imagesize` as a dependency in `setup.py`.
btw in the future with some dataloader changes I hope to remove the need to include width/height in the dicts.

botcs:
`imagesize` was already installed somehow, probably it was a dependency for a pip package.
Do you want me to still add it to `setup.py` explicitly?

ppwwyyxx:
Perhaps it was already a dependency, but it's still good to declare it explicitly.

botcs:
Done in 84ac5f6

PR #379, opened 2019-11-22
GUI support for docker.

maxfrei750:
Thanks for the great repo! I very much like the easy set up via docker. However, one of the first things I did, was to figure out a way to display stuff like images or plots, without having to save them as images. I thought this might be interestering to others as well.

Kind regards.


ppwwyyxx:
We'd like to avoid extra files if not needed, especially the ubuntu-specific setup script.
On my machine, starting `xhost local:root` and then `nvidia-docker run -it --name detectron2 --net=host --env="DISPLAY" --volume="/tmp/.X11-unix:/tmp/.X11-unix:rw" detectron2:v0` is enough to use GUI from the docker. Then just adding these instructions in the comment should be sufficient.

maxfrei750:
It makes perfect sense to not add too much clutter. To be honest, I never expected the pull request to be merged right away, but wanted to use it as a starting point.

I like the idea to include the information as instructional comment in the dockerfile.

However, concerning `xhost local:root`, I'd still prefer
``` xhost +local:`docker inspect --format='{{ .Config.Hostname }}' detectron2` ```, 
due to security reasons (see https://wiki.ros.org/docker/Tutorials/GUI).

As soon as we have settled on a solution, I will prepare another small commit for you to cherrypick.

ppwwyyxx:
The safer way sounds good to me.

maxfrei750:
That's the approach that I originally used (thats also the reason for the bloated docker-compose.yml that I commited). However, it caused problems due to the changed user. Generally, it would be nice if the docker user was not root. However, that does not work at the moment and is an issue for itself.

maxfrei750:
Ok, I tried to clean up the PR a little. I think you should be able to just squash and merge.

ppwwyyxx:
Thanks!

maxfrei750:
If #383 was merged, then GUI support could be achieved via approach "2. The safer way" (https://wiki.ros.org/docker/Tutorials/GUI#The_safer_way). This would have the benefit, that the fiddeling with `xhost` would become obsolete.

ividovic:
How can I show image from docker? I have run image using:

docker run --gpus all -it \
	--shm-size=8gb --env="DISPLAY" --volume="/tmp/.X11-unix:/tmp/.X11-unix:rw" \
	--name=detectron2 detectron2:v0

and before that Grant docker access to host X server to show images using:

xhost +local:`docker inspect --format='{{ .Config.Hostname }}' detectron2`

PR #383, opened 2019-11-23
Create and switch to non-root user.

### LINKED ISSUE #381: Use a non-root docker user.
## 🚀 Feature
Use a non-root docker user for security reasons (see https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b).

## Motivation
Running everything in the docker container as root imposes a security risk. Even without possible security concerns, it can be tedious that everything that is written to the disk by the container belongs to root and can only be deleted by a root user (e.g. with sudo).

## Pitch
Fix the Dockerfile, so that the container can be run with `--user=$UID` or have a user with known UID and limited rights inside the docker container.

At the moment, doing the former resutls in errors like the following:
> FileNotFoundError: [Errno 2] No such file or directory: '/.torch/fvcore_cache/detectron2/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_3x/137849600/model_final_f10217.pkl.lock'

maxfrei750:
Fixes #381 
The container does not operate as root anymore, but as a new user `appuser` with limited rights, which improves security.

However, this does not fix the fact that outputfiles cannot be deleted without `sudo` (because they belong to `appuser`. To prevent that, docker can be called with the parameter `--user=$UID`. Doing so results in files that belong to the host user.

This could be added as an instructional comment int the Dockerfile. However, the Dockerfile becomes more and more bloated with instructions. It might be preferable, to introduce a docker-compose file, which includes default settings, like `user: $UID` or the GUI support, discussed in #379.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,12 @@ RUN pip install -e /detectron2_repo
 
 WORKDIR /detectron2_repo
 
+# create and switch to non-root user
+ENV FVCORE_CACHE="/tmp"
+RUN groupadd -g 999 appuser && \
+    useradd -r -u 999 -g appuser appuser
+USER appuser
+
 # run it, for example:
 # wget http://images.cocodataset.org/val2017/000000439715.jpg -O input.jpg
 # python3 demo/demo.py  \


maxfrei750:
Closing, because this will be implemented along with other changes in #385.

Issue #384, opened 2019-11-23
RuntimeError: DataLoader worker exited unexpectedly


maxfrei750:
I add this issue for documentation purposes.

When running a training in the docker container, I received the following error: 
`RuntimeError: DataLoader worker (pid(s) 25) exited unexpectedly`

maxfrei750:
**Solution:** 
Adding `--shm-size="8g"` to the docker command.

Issue #382, opened 2019-11-23
Save the planet by not downloading pretrained models over and over again, when using docker.


maxfrei750:
## 🚀 Feature
It would be great to find a way that the use of docker does not result in models being downloaded over and over again.

## Motivation
Models are quite large. Their repetitive download wastes precious resources, such as energy, time, bandwidth and money.

## Pitch
There should be a solution to avoid the repetitive model download, when using docker.

Any ideas?

bendidi:
just mount the FVCORE_CACHE volume into your container at run time

maxfrei750:
Thanks for the advice. I already thought about that and it is definitively the easiest solution. However, I was not sure whether it is a wise idea to reduce the isolation of the container and the host. On the other hand, the only other solution that I see would be to include the models into the docker image, which would severely bloat it.

PR #385, opened 2019-11-23
Added docker compose file with useful tweaks.

maxfrei750:
From my perspective a docker-compose file has several benefits. On the one hand, it increases the comfort and on the other hand it is a way to supply users with useful tweaks. The commited docker-compose file addresses several issues and tweaks:
1. It fixes potential problems with dataloaders (see #384).
2. It includes Multi-GPU and performance tweaks as suggested by NVIDIA (see https://docs.nvidia.com/deeplearning/frameworks/user-guide/index.html#caffeovr).
3. It adds GUI support (see #379).
4. It enables caching of downloaded models (see #382).
5. It makes docker run with the UID of the host user.

Of course, all this can be accomplished with a long docker command. However, a docker-compose file give a central place to gather all recommendations for running detectron2 with docker, without bloating the Dockerfile with comments.


docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+services:
+  detectron2:
+    build: .
+    runtime: nvidia  # TODO: Exchange with "gpu: all" in the future (see https://github.com/facebookresearch/detectron2/pull/197/commits/00545e1f376918db4a8ce264d427a07c1e896c5a).
+    shm_size: "8gb"
+    ulimits:
+      memlock: -1
+      stack: 67108864
+    volumes:
+      - /tmp:/tmp
```
ppwwyyxx: how about "/tmp/.X11-unix"?

ppwwyyxx:
Since we're adding a new file and I expect a new dockerfile in the future for caffe2 integration, could you move them to a dedicated directory `docker/`, and move the instructions to `docker/README.md`? This would be more friendly to users.

docker-compose.yml
maxfrei750: I'm glad you like it. You are absolutely right about the model caching. The best solution depends on, whether you would like to enable caching of downloaded models or not.
If yes, then there are two options:
1. Wait for #383 to be merged. Then the current version would enable model caching.
2. Add
```
  environment:
    FVCORE_CACHE=/tmp
```
to the docker-compose file, to support model caching without #383 being merged before.

If not, then we should make the change you suggest.

maxfrei750:
Seems like agood idea. What do you want to have included in the readme? Only the commands to spin up docker compose or also the commands from the dockerfile to run the container without docker compose?

docker-compose.yml
ppwwyyxx: The models are quite large (depend on how many you'll use) so I'm not sure using `/tmp` is the best option to cache models. Is it feasible to mount `$HOME/.torch/fvcore_cache`?

ppwwyyxx:
My thought was that the README would become a better place to show usage instructions than comments in the Dockerfile.

maxfrei750:
I got that, but do you still want to include the pure docker commands or does the docker compose command suffice?

ppwwyyxx:
A command that uses docker-compose suffices

maxfrei750:
I implemeneted your suggestions. However, it needed some fiddeling, to use the host user inside the docker container so that the models can be stored in the home directory of the host user. Due to the need for the creation of the cache folder by the host user and the environment variables, I thought a small script to execute docker (run_docker.sh) would be reasonable.

ppwwyyxx:
It seems like now detectron2 & python dependencies will be installed as root, and therefore cannot be modified by the docker user.

maxfrei750:
I think that's correct. However, I thought that it is intended to be that way (see https://medium.com/faun/set-current-host-user-for-docker-container-4e521cef9ffc, section "Extra: Grant user sudo privileges"). If you wanted to, we could add the sudo package as described in the linked article.

ppwwyyxx:
I think we should install them with user's permission, so users can download images inside detectron2 directory or make changes to code.

maxfrei750:
Sound reasonable but I think these changes are not related to this PR and should be made in #383 
Apart from that, I think I should add a section in the readme on how to rebuild the docker image, if necessary. What do you think?

ppwwyyxx:
It seems #383 will be made obsolete due to what's changed here. Could you update the users in this PR instead?

More instructions in README sounds good. In fact since there are only 4 lines in the script (and 3 are environment setup that are necessary for both running and rebuilding the image), I would rather see them in README instructions.

maxfrei750:
I will integrate the changes of #383 and improve the readme. 

Concerning the access rights: I think that it might be the easiest solution to change the access rights of the `detectron2_repo` folder to `chmod 777`. Then users can do to the code whatever they want. They would, however, still not be able to install new packages. However, if they want to do that, IMHO they should alter the dockerfile and rebuild. What do you think?

ppwwyyxx:
I suppose it's also common to use Docker also as a temporary development environment so users should be able to install/modify packages easily. If this can't be addressed I would prefer the flexibility of development over the security concerns of root user.

maxfrei750:
Then I think the best solution is to provide `sudo` for the users. Like that, they can use their own credentials to temporarily install whatever they may desire. I'll add that later today or tomorrow.

maxfrei750:
I think, unless you have further improvements, that should be it. 😃

ppwwyyxx:
It doesn't seem like this will solve the problem. It looks like the code of detectron2 will still be owned by root, which means that a user will need sudo for every command including running the minimal demo in https://github.com/facebookresearch/detectron2/blob/master/GETTING_STARTED.md. In that case this is probably not more secure than running everything as root from the beginning.

`chmod 777` everything is also hacky and it will be recognized as a change by git. 

I think the ideal case is to have detectron2 cloned & installed as non-root (not necessarily the host username), and also have sudo access to use `apt` -- because this is exactly what a normal host development environment would look like. As I said I think this flexibility wins over the security concerns of using root inside docker.

maxfrei750:
> It doesn't seem like this will solve the problem. It looks like the code of detectron2 will still be owned by root, which means that a user will need sudo for every command including running the minimal demo in https://github.com/facebookresearch/detectron2/blob/master/GETTING_STARTED.md. In that case this is probably not more secure than running everything as root from the beginning.
> 
> `chmod 777` everything is also hacky and it will be recognized as a change by git.

Makes perfect sense.

> I think the ideal case is to have detectron2 cloned & installed as non-root (not necessarily the host username), and also have sudo access to use `apt` -- because this is exactly what a normal host development environment would look like. As I said I think this flexibility wins over the security concerns of using root inside docker.

I tried following your advice but, unfortunately, it did not work properly. It may also be that I did not quite get what you suggested. However, I went with making the host user the owner of /detectron2_repo after installation. Now the image should have the requested properties.

ppwwyyxx:
I got 
```
Step 15/17 : RUN if [ ${USER_ID:-0} -ne 0 ] && [ ${GROUP_ID:-0} -ne 0 ]; then     if getent passwd ${USER_NAME}; then userdel --force ${USER_NAME}; fi &&    if getent group ${USER_NAME}; then groupdel ${USER_NAME}; fi &&    groupadd
 --gid ${GROUP_ID} ${USER_NAME}&&    useradd --no-log-init --uid ${USER_ID} --gid ${GROUP_ID} ${USER_NAME} &&    chown --changes --silent --no-dereference --recursive         ${USER_ID}:${GROUP_ID}         /detectron2_repo ;fi
 ---> Running in a246bb6e6151
groupadd: GID '100' already exists
```
when running the latest version.

maxfrei750:
Oh dear... 😅 I think I was able to fix it. Please test the current version.

ppwwyyxx:
This is working fine on one machine. However when tested on another machine, I saw this error:
```
╰─$docker-compose build detectron2                                                                                                                      
WARNING: The DISPLAY variable is not set. Defaulting to a blank string.                                                                              
ERROR: The Compose file './docker-compose.yml' is invalid because:                                                                                   
Unsupported config option for services.detectron2: 'runtime'    
```
```
╰─$docker --version; docker-compose --version
Docker version 18.06.1-ce, build e68fc7a
docker-compose version 1.17.1, build unknown
```

I'm not a docker expert -- the version seems high enough for the option so I'm not sure what else might went wrong.

ppwwyyxx:
Also, the uid hack is not working reliably. After removing runtime I'm able to run it, but the username does not map properly into the docker. Mounting "/etc/{group,passwd,shadow}" only works for a standard small linux environment, in large clusters where accounts are managed by non-standard pam module, these files do not contain much information about accounts.
What is the reason for having the same username or uid? Just using a dummy account like "appuser" might be better.

Also, mounting home is not working for me either. I got errors like:
```
ERROR: Cannot start service detectron2: error while creating mount source path 'XXX/.torch/fvcore_cache': mkdir XXX/.torch: permission denied
```
the reason seems to because my home is on an NFS filesystem and docker is unable to mount an NFS into the container.

Due to all these it seems like we should really minimize the connection between host and the container - after all that's what a container is for.

maxfrei750:
> However when tested on another machine, I saw this error:
> ...

I guess we should address that after we have sorted out the rest.

> I'm not a docker expert

As you might have noticed by now.... me neither. 😅

> Due to all these it seems like we should really minimize the
connection between host and the container - after all that's what a
container is for.

I totally agree. This PR grew quite a lot in the process and I would
definitively prefer are more elegant solution.

> What is the reason for having the same username or uid? Just using a
dummy account like "appuser" might be better.

I couldn't figure out a way to let a dummy user write files to mounted volumes, because it lacks the necessary permissions. On the other hand, a root user can write files to mounted volumes but then they can only be deleted by using sudo on the host. However, it might very well be that there is a better solution for this problem then to recreate the host user in the docker container. Input would be more than welcome.

> the reason seems to because my home is on an NFS filesystem and docker is unable to mount an NFS into the container.

To be honest, I didn't like the idea to store the downloaded models in the host users home folder in the first place. However, it seemed to be the most sensible option apart from /tmp. Can you think of some better alternative? At the moment I would prefer /tmp.

ppwwyyxx:
> I couldn't figure out a way to let a dummy user write files to mounted volumes, because it lacks the necessary permissions. 

I suppose to achieve this you only need the same uid, but not the username?

> To be honest, I didn't like the idea to store the downloaded models in the host users home folder in the first place. However, it seemed to be the most sensible option apart from /tmp. Can you think of some better alternative? At the moment I would prefer /tmp.

Since this is an interaction with host it would be better to make it explicit than implicit, also to avoid saving data to "/tmp" without user's awareness, or to have container's garbage file appear in `host:/tmp`. 
I suggest using an envvar, or an argument (if docker-compose supports) that maps a user-provided cache dir to a fixed location inside the container. The default behavior should be no volume mapping.

maxfrei750:
> I suppose to achieve this you only need the same uid, but not the username?

Correct. However, to let the user install packages, i.e. have `sudo` working, I still need to mount `/etc/group`, `/etc/passwd`, `/etc/shadow` and `/etc/sudoers.d`, which will still lead to the problems that you explained in https://github.com/facebookresearch/detectron2/pull/385#issuecomment-558868315. I think there might be three possible solutions to this (ordered by my preference):
1. We find an elegant solution to support non-root users on non-standard linuxes. Unfortunately, I won't of big help with this, because I have no experience with such setups. Can you think of a way?
2. Users of non-standard linuxes have to specify `user: "root"` in the docker-compose file or run `docker-compose run --user "root" detectron2` to run the container as root.
3. We abandon the idea of a non-root user and accept the security drawbacks and file ownership issues.

> Since this is an interaction with host it would be better to make it explicit than implicit, also to avoid saving data to "/tmp" without user's awareness, or to have container's garbage file appear in host:/tmp.
I suggest using an envvar, or an argument (if docker-compose supports) that maps a user-provided cache dir to a fixed location inside the container. The default behavior should be no volume mapping.

There is a way to have a base-docker-compose-file, which can be enhanced with a second file (see https://docs.docker.com/compose/extends/#example-use-case). I think this might be a way to realise the desired behavior.

I commited some changes to reflect the changes and proposals that I explained above. I thought this might make it easier to discuss them. Please be aware that they were not fully tested yet.

ppwwyyxx:
```
RUN apt-get update && apt-get install -y \
        libpng-dev libjpeg-dev python3-opencv ca-certificates \
        python3-dev build-essential pkg-config git curl wget automake libtool sudo

RUN useradd -r -u 1000 appuser -g sudo
RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers

USER appuser
```
can enable sudo for a new user.

maxfrei750:
> `RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers` 

That's correct and I already thought about that solution. However, security-wise, disabling the password protection comes very close to using a root user. It has however the benefit that files written by the container should be manageable (e.g. deletable) by the host user.

> `RUN useradd -r -u 1000 appuser -g sudo`

I think a fixed UID might lead to problems, especially in a cluster environment with a lot of users.

ppwwyyxx:
I don't think there is a robust way to use password from host. So this is probably the best way to go.
UID can come from the environment variable as you did now. 

Is is possible to use cache directory from environment variable as well, but also make it not mount any volume when the envvar is empty? If that's hard to achieve let's do it explicit via the command line.

ppwwyyxx:
I'd say the default should be something that always works: i.e. use appuser with no password, and mount no cache directory. If a user want more fancy features that are less reliable, they should enable them with extra arguments -- in fact I would rather call these users "power users".

maxfrei750:
The current code should now meet your requirements.

However:
I'd say the default should be something that ~~always~~ works *for the large majority of users and is secure*: i.e. use appuser with ~~no~~ password, and mount no cache directory. If a user wants more fancy features *(e.g. support for non-standard pam)* that are less ~~reliable~~ *secure*, they should enable them with extra arguments *(e.g. `--user "root"`)* -- in fact *I* would rather call *these* users "power users".

My main concern is, that users now cannot easily disable root access to their host, while before it was 
disabled by default and was very easy to activate, if desired or necessary.

IMHO, the decision comes down to whether you want ease-of-use for each and every one (i.e. in this case cluster users) or security by design.

All that said, as I said before, the decision is obviously up to you. You can merge the code as-is or revert the last commit, if I should have actually convinced you.

docker/README.md
```diff
@@ -0,0 +1,45 @@
...
+# Customization
+USER root
+RUN apt-get update && apt-get install -y \
+  nano vim emacs
+RUN pip install pandas
+USER appuser
+```
+
+### Temporary
+Use sudo (with your usual credentials) inside of the container. Changes will be lost when the container is restarted.
```
ppwwyyxx: could you correct these instructions and the title?

.gitignore
```diff
@@ -38,3 +38,6 @@ dist/
 # project dirs
 /datasets
 /models
+
+# docker compose files
+/docker/.env
```
ppwwyyxx: looks like it is no longer used. could you remove it?

docker/README.md
maxfrei750: I corrected the instructions. However, I am not sure what has to be corrected about the title.

.gitignore
maxfrei750: Done. Thanks for reminding me.

ppwwyyxx:
Could you resolve the merge conflicts?

maxfrei750:
I think that should do it.

PR #395, opened 2019-11-25
Expose select_foreground_proposals and select_proposals_with_visible_keypoints.

maxfrei750:
Expose select_foreground_proposals and select_proposals_with_visible_keypoints, as proposed in the densepose example.

If this was to be merged, it would be sensible to update other code passages (e.g. the densepose example) to reflect this change.


ppwwyyxx:
Thanks for pointing this out. We'd be conservative on what to expose in `__init__.py`. At this time it seems reasonable to add `select_foreground_proposals` but we'll hold the other one. I'll commit this together with a change on the densepose side, and also fix the lint.

PR #420, opened 2019-11-28
Another "avoid `torch/extension.h` in ops to be friendly to nvcc9.0"

bernhardschaefer:
Follow-up to #414: by replacing this additional torch/extension.h include I am able to successfully build detectron2 with nvcc9.0.
Disclaimer: I haven't tested applying deformable convolution with this change, I can only verify that this fixes the build problem for me.

### DIFF
--- a/detectron2/layers/csrc/deformable/deform_conv_cuda.cu
+++ b/detectron2/layers/csrc/deformable/deform_conv_cuda.cu
@@ -8,7 +8,7 @@
 // https://github.com/chengdazhi/Deformable-Convolution-V2-PyTorch/blob/mmdetection/mmdet/ops/dcn/src/deform_conv_cuda.c
 // Original license: Apache 2.0
 
-#include <torch/extension.h>
+#include <torch/types.h>
 
 #include "deform_conv.h"
 


bernhardschaefer:
Update: I just went through the build logs, and I get a lot of `tensor is not on GPU!` warnings.
Not sure how critical those warnings are. 
Examples:
```
/home/ubuntu/detectron2/detectron2/layers/csrc/deformable/deform_conv.h:286:5: note: in expansion of macro ‘AT_CHECK’
     AT_CHECK(offset.type().is_cuda(), "offset tensor is not on GPU!");
```
```
/home/ubuntu/detectron2/detectron2/layers/csrc/deformable/deform_conv.h:285:5: note: in expansion of macro ‘AT_CHECK’
     AT_CHECK(bias.type().is_cuda(), "bias tensor is not on GPU!");
```
```
/home/ubuntu/detectron2/detectron2/layers/csrc/deformable/deform_conv.h:136:5: note: in expansion of macro ‘AT_CHECK’
     AT_CHECK(weight.type().is_cuda(), "weight tensor is not on GPU!");
```

ppwwyyxx:
That seems to be a warning for `AT_CHECK` only and should be safe to ignore.

PR #454, opened 2019-12-03
Update detection_utils.py

Bekovmi:
To except following error https://stackoverflow.com/questions/12984426/python-pil-ioerror-image-file-truncated-with-big-images.

Thanks for your contribution!

If you're sending a large PR (e.g., >50 lines),

### DIFF
--- a/detectron2/data/detection_utils.py
+++ b/detectron2/data/detection_utils.py
@@ -10,6 +10,8 @@ import numpy as np
 import torch
 from fvcore.common.file_io import PathManager
 from PIL import Image, ImageOps
+from PIL import ImageFile
+ImageFile.LOAD_TRUNCATED_IMAGES = True
 
 from detectron2.structures import (
     BitMasks,


bendidi:
for me this should be an env var that controles the behavior with the default being throwing an error (because not everyone may want to train on truncated images)

so the 3 options would be:
- load truncated
- ignore truncated
- throw truncated image error

the default can also be to ignore it

ppwwyyxx:
Making the simple logic of "reading images" more complicated will make it hard for us to switch the image processing backend in the future (since PIL is quite slow). We'd like to keep it simple and I think truncated images should be avoided most of the time.

bendidi:
so can't they just be ignored in the mapper ? it'll be a shame to kill a training because of only having one truncated image in the dataset

ppwwyyxx:
If you really want, you can return `None` in your mapper to skip such images. https://detectron2.readthedocs.io/modules/data.html#detectron2.data.MapDataset will retry a different image if `None` is found.

Bekovmi:
> for me this should be an env var that controles the behavior with the default being throwing an error (because not everyone may want to train on truncated images)
> 
> so the 3 options would be:
> 
> * load truncated
> * ignore truncated
> * throw truncated image error
> 
> the default can also be to ignore it

I  delete those images from train and val dataset. Bug was disappeared.   Thanks for advice.

Bekovmi:
> Making the simple logic of "reading images" more complicated will make it hard for us to switch the image processing backend in the future (since PIL is quite slow). We'd like to keep it simple and I think truncated images should be avoided most of the time.
I was confusion, whick I saw pil jpeg decoder. For fast decoding I recommend PyTurboJPEG https://pypi.org/project/PyTurboJPEG/.

PR #488, opened 2019-12-09
anchor_offsets

### LINKED ISSUE #378: Are 'x_shifts' and 'y_shifts' in '_create_grid_offsets' correct?
Hi, first of all, thanks for making this excellent project open source.

My colleague @aoikonomop and I have been recently looking at the matching between `gt_boxes` and `anchors` for the `RetinaNet` model. In doing so, we have observed that `grid-anchors` seem to be centered at the position [(0, 0)](https://github.com/facebookresearch/detectron2/blob/fbd447d4de532311b8a93495e118785d9e904b95/detectron2/modeling/anchor_generator.py#L43-L46) relative to the `cells` in the response maps where the final predictions are made (`p3`, `p4`, ...). We were expecting these `grid-anchors` to be centered at the position `(stride/2, stride/2)` relative to the `cells`. Because `cell-anchors` themselves are centered at [(0, 0)](https://github.com/facebookresearch/detectron2/blob/fbd447d4de532311b8a93495e118785d9e904b95/detectron2/modeling/anchor_generator.py#L166) it seems to us that all `grid-anchors` are unnecessarily shifted toward the upper left of the cells. This, in turn, seems to produce some unnexpected (nonoptimal?) matchings for some simple cases, see this [gist](https://gist.github.com/jalabort/2b181e36d5af29fce4e49f3f79d405b1) for an example. Is this behaviour intended?

We have managed to get the matching results we were expecting by simply changing the `0`s in [here](https://github.com/facebookresearch/detectron2/blob/fbd447d4de532311b8a93495e118785d9e904b95/detectron2/modeling/anchor_generator.py#L43-L46) for `stride/2` which seems to make sense given our prior intutition.

ArutyunovG:
Fixes #378

### DIFF
--- a/detectron2/config/defaults.py
+++ b/detectron2/config/defaults.py
@@ -177,7 +177,9 @@ _C.MODEL.ANCHOR_GENERATOR.ASPECT_RATIOS = [[0.5, 1.0, 2.0]]
 # list[float], the angle in degrees, for each input feature map.
 # ANGLES[i] specifies the list of angles for IN_FEATURES[i].
 _C.MODEL.ANCHOR_GENERATOR.ANGLES = [[-90, 0, 90]]
-
+# Offset of anchor relave to tl corner of the pixel
+# Allowed values are floats in [0, 1] range inclusive
+_C.MODEL.ANCHOR_GENERATOR.OFFSET = 0.0
 
 # ---------------------------------------------------------------------------- #
 # RPN options
--- a/detectron2/config/defaults.py
+++ b/detectron2/config/defaults.py
@@ -177,7 +177,8 @@ _C.MODEL.ANCHOR_GENERATOR.ASPECT_RATIOS = [[0.5, 1.0, 2.0]]
 # list[float], the angle in degrees, for each input feature map.
 # ANGLES[i] specifies the list of angles for IN_FEATURES[i].
 _C.MODEL.ANCHOR_GENERATOR.ANGLES = [[-90, 0, 90]]
-# Offset of anchor relave to tl corner of the pixel
+# Offset of anchor relative to top-left corner of the pixel
+# Units: fraction of feature map stride (e.g., 0.5 means half stride)
 # Allowed values are floats in [0, 1] range inclusive
 _C.MODEL.ANCHOR_GENERATOR.OFFSET = 0.0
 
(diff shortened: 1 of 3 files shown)


detectron2/config/defaults.py
```diff
@@ -177,7 +177,9 @@
 # list[float], the angle in degrees, for each input feature map.
 # ANGLES[i] specifies the list of angles for IN_FEATURES[i].
 _C.MODEL.ANCHOR_GENERATOR.ANGLES = [[-90, 0, 90]]
-
+# Offset of anchor relave to tl corner of the pixel
+# Allowed values are floats in [0, 1] range inclusive
```
rbgirshick: - Please expand "tl" to "top-left".
- Specify the units for this option: "fraction of feature map stride (e.g., 0.5 means half stride)"

detectron2/modeling/anchor_generator.py
```diff
@@ -62,6 +66,10 @@ def __init__(self, cfg, input_shape: List[ShapeSpec]):
         sizes         = cfg.MODEL.ANCHOR_GENERATOR.SIZES
         aspect_ratios = cfg.MODEL.ANCHOR_GENERATOR.ASPECT_RATIOS
         self.strides  = [x.stride for x in input_shape]
+        self.offset = cfg.MODEL.ANCHOR_GENERATOR.OFFSET
```
rbgirshick: nit: align on "="

detectron2/modeling/anchor_generator.py
```diff
@@ -201,7 +209,11 @@ def __init__(self, cfg, input_shape: List[ShapeSpec]):
         sizes         = cfg.MODEL.ANCHOR_GENERATOR.SIZES
         aspect_ratios = cfg.MODEL.ANCHOR_GENERATOR.ASPECT_RATIOS
         angles        = cfg.MODEL.ANCHOR_GENERATOR.ANGLES
-        self.strides  = [x.stride for x in input_shape]
+        self.strides  = [x.stride for x in input_shape]        
+        self.offset = cfg.MODEL.ANCHOR_GENERATOR.OFFSET
```
rbgirshick: alignment

detectron2/config/defaults.py
ArutyunovG: Fixed

detectron2/modeling/anchor_generator.py
ArutyunovG: Fixed alignments

PR #510, opened 2019-12-12
Use pep517 to build install detectron2

bendidi:
This MR contains two main changes: 

1- add support for `pep517` by adding a `pyproject.toml` with `torch` as a requirement for the `build-system`, and in the same way add torch and torchvision to the `install_requires` of `setup.py`

2- (optional) set a fixed version of torch and torchvision (quite important in my point of view haha )


ppwwyyxx:
We prefer to not put torch and torchvision as dependencies, because the one that will be automatically downloaded from pypi and tied to cuda 10.1 is often not the version a user would use.

PR #520, opened 2019-12-13
Some refactoring

olramde:
Normal Class changed to Abstract Class.
As previously distributed, we changed 'if' to 'elif'.

Thank you very much to Facebook Research for providing the best project.

### DIFF
--- a/detectron2/evaluation/evaluator.py
+++ b/detectron2/evaluation/evaluator.py
@@ -5,7 +5,7 @@ import time
 from collections import OrderedDict
 from contextlib import contextmanager
 import torch
-from abc import *
+from abc import ABCMeta, abstractmethod
 
 from detectron2.utils.comm import is_main_process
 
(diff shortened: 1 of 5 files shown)


ppwwyyxx:
could you run `./dev/linter.sh` on the PR? I think `import *` is disallowed by flake8.

olramde:
@ppwwyyxx 
Thank you for pointing out the problem. The previous code didn't fit the flake8 style. I fixed the problem.

ppwwyyxx:
1. Your PR has conflicts with current master. That may be because you've modified `train_net` in another PR already.

2. Don't make `HookBase` abstractclass because a subclass is not required to implement any of its methods.

detectron2/evaluation/evaluator.py
```diff
@@ -20,13 +21,15 @@ class DatasetEvaluator:
     and produce evaluation results in the end (by :meth:`evaluate`).
     """
 
+    @abstractmethod
```
ppwwyyxx: remove this because a subclass may very well leave it as is.

olramde:
@ppwwyyxx  I'll solve all the problems you've pointed out, and I'll do PR again. Thank you

PR #543, opened 2019-12-17
Fix multiple processes writing coco json to same file

corysaildrone:
When CocoEvaluator with multiple workers using a custom loader, it appears that each worker converts the Detectron2 dataset to COCO json and then dump to disk. This can cause the output file to be corrupted. Here my fix is to only convert/dump on the main process/

### DIFF
--- a/detectron2/data/datasets/coco.py
+++ b/detectron2/data/datasets/coco.py
@@ -12,6 +12,7 @@ from PIL import Image
 
 from fvcore.common.timer import Timer
 from detectron2.structures import BoxMode, PolygonMasks, Boxes
+from detectron2.utils import comm
 from fvcore.common.file_io import PathManager
 
 
@@ -408,7 +409,8 @@ def convert_to_coco_json(dataset_name, output_folder="", allow_cached=True):
     PathManager.mkdirs(output_folder)
     if os.path.exists(cache_path) and allow_cached:
         logger.info(f"Reading cached annotations in COCO format from:{cache_path} ...")
-    else:
+    elif comm.is_main_process():
+        # Only convert/dump in the main process to avoid multiple processes writing to file
         logger.info(f"Converting dataset annotations in '{dataset_name}' to COCO format ...)")
         coco_dict = convert_to_coco_dict(dataset_name)
 
--- a/detectron2/data/datasets/coco.py
+++ b/detectron2/data/datasets/coco.py
@@ -398,26 +398,22 @@ def convert_to_coco_json(dataset_name, output_file, allow_cached=True):
             must be registered in DatasetCatalog and in detectron2's standard format
         output_file: path of json file that will be saved to
         allow_cached: if json file is already present then skip conversion
-    Returns:
-        cache_path: path to the COCO-format json file
     """
 
     # TODO: The dataset or the conversion script *may* change,
     # a checksum would be useful for validating the cached data
 
     PathManager.mkdirs(os.path.dirname(output_file))
-    if os.path.exists(output_file) and allow_cached:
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
Thanks for finding this issue! It would be better to fix it using the pattern described in https://github.com/facebookresearch/fvcore/blob/8694adf300c4e47d575ad1583bfb9d646fe9c12c/fvcore/common/file_io.py#L46-L64 
instead.

In addition, you can add `if distributed and comm.is_main_process()` in `COCOEvaluator.__init__`.

corysaildrone:
I reorganized the code a little bit, because previously `COCOEvaluator.__init__()` was relying on `convert_to_coco_json` to create the cached filename - this meant it had to be run in each process to update the metadata. Let me know if this seems.

Just adding the `file_lock` would be less invasive, but then the dataset conversion would have been performed multiple times sequentially.

detectron2/data/datasets/coco.py
```diff
@@ -387,36 +387,37 @@ def convert_to_coco_dict(dataset_name):
...
     # TODO: The dataset or the conversion script *may* change,
     # a checksum would be useful for validating the cached data
-    cache_path = os.path.join(output_folder, f"{dataset_name}_coco_format.json")
-    PathManager.mkdirs(output_folder)
-    if os.path.exists(cache_path) and allow_cached:
-        logger.info(f"Reading cached annotations in COCO format from:{cache_path} ...")
+
+    PathManager.mkdirs(os.path.dirname(output_file))
+    if os.path.exists(output_file) and allow_cached:
+        logger.info(f"Cached annotations in COCO format already exist: {output_file}")
```
ppwwyyxx: At this point the file could be a partial file that's being written to. So we could add 
```
with file_lock(...):
    pass
```
to wait for the writing.

Assuming that the `output_file` is only going to be written by one process (with the comment below), this should be sufficient.

detectron2/data/datasets/coco.py
```diff
@@ -387,36 +387,37 @@ def convert_to_coco_dict(dataset_name):
...
     else:
         logger.info(f"Converting dataset annotations in '{dataset_name}' to COCO format ...)")
         coco_dict = convert_to_coco_dict(dataset_name)
 
-        with PathManager.open(cache_path, "w") as json_file:
-            logger.info(f"Caching annotations in COCO format: {cache_path}")
-            json.dump(coco_dict, json_file)
+        with file_lock(output_file):
```
ppwwyyxx: the typical pattern in https://github.com/facebookresearch/fvcore/blob/8694adf300c4e47d575ad1583bfb9d646fe9c12c/fvcore/common/file_io.py#L46-L64 is to check whether the file exists after obtaining the lock.
This will avoid rewriting the file, which makes sense here.

detectron2/data/datasets/coco.py
```diff
@@ -387,36 +387,37 @@ def convert_to_coco_dict(dataset_name):
...
 
     Args:
         dataset_name:
             reference from the config file to the catalogs
-            must be registered in DatastCatalog and in detectron2's standard format
-        output_folder: where json file will be saved and loaded from
+            must be registered in DatasetCatalog and in detectron2's standard format
+        output_file: path of json file that will be saved to
         allow_cached: if json file is already present then skip conversion
     Returns:
```
ppwwyyxx: could you also make it return nothing?

detectron2/data/datasets/coco.py
corysaildrone: Ah I see, that makes more sense. I'll move the check. Should I remove the process checking in `COCOEvaluator` as well then?

PR #643, opened 2020-01-07
Introduced vertical flipping in RandomFlip.

### LINKED ISSUE #507: Implement/Enable vertical flipping of images
## 🚀 Feature
Implement/Enable vertical flipping of images.

## Motivation
Vertical flipping is a useful augmentation for many kinds of images. 

## Pitch
We should implement/enable vertical flipping of images. If I see it correctly, most of it should already be available (e.g. `VFlipTransform` is already implemented in fvcore). At first glance, mainly `RandomFlip` in `detectron2/data/transforms/transform.py` needs to be changed.

Are there other pieces of code that would need adaption? For instance in the `config` module?

If desired, I could try myself on Implementing it and make a PR.

maxfrei750:
Fixes #507.

### DIFF
--- a/detectron2/data/transforms/transform_gen.py
+++ b/detectron2/data/transforms/transform_gen.py
@@ -144,7 +144,6 @@ class RandomFlip(TransformGen):
             return NoOpTransform()
 
 
-
 class Resize(TransformGen):
     """ Resize image to a target size"""
 
--- a/detectron2/data/transforms/transform_gen.py
+++ b/detectron2/data/transforms/transform_gen.py
@@ -14,6 +14,7 @@ from fvcore.transforms.transform import (
     NoOpTransform,
     Transform,
     TransformList,
+    VFlipTransform,
 )
 from PIL import Image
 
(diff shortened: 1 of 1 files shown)


maxfrei750:
With hindsight to https://github.com/facebookresearch/detectron2/issues/507#issuecomment-571532082, it might make sense to
a. Rename the config item `TEST.AUG.FLIP` to `TEST.AUG.HFLIP`
and/or
b. Introduce a config item `TEST.AUG.VFLIP`.

I think, option a would be good, to to be more explicit. The implementation of b on the other hand might not be as important.

detectron2/data/transforms/transform_gen.py
```diff
@@ -113,18 +114,16 @@ def __repr__(self):
...
 class RandomFlip(TransformGen):
     """
-    Flip the image horizontally with the given probability.
-
-    TODO Vertical flip to be implemented.
+    Flip the image horizontally or vertically with the given probability.
     """
 
-    def __init__(self, prob=0.5):
+    def __init__(self, prob=0.5, horiz=True, vert=False):
```
ppwwyyxx: Could you use the signature
```
__init__(self, prob=0.5, *, horizontal=True, vertical=False):
```
They now becomes an API so we expect clearer names. Also it's good to prevent users writing `RandomFlip(0.5, True)` since it's generally confusing.

maxfrei750: Done.

maxfrei750:
What about https://github.com/facebookresearch/detectron2/pull/643#issuecomment-571566468?

detectron2/data/transforms/transform_gen.py
```diff
@@ -113,31 +114,32 @@ def __repr__(self):
...
-        # TODO implement vertical flip when we need it
         super().__init__()
 
-        if horiz and vert:
+        if horizontal and vertical:
             raise ValueError("Cannot do both horiz and vert. Please use two Flip instead.")
-        if not horiz and not vert:
+        if not horizontal and not vertical:
             raise ValueError("At least one of horiz or vert has to be True!")
         self._init(locals())
```
maxfrei750: This line got me. Is there a special reason, that the attribute definition is implicit? As we just saw, it is prone to error.

PR #648, opened 2020-01-08
pin memory in DataLoader

xmyqsh:


ppwwyyxx:
Thanks for bringing this up. I'm wondering what types of perf improvements can you see, and under what environment & job setup?

xmyqsh:
@ppwwyyxx 
After using pin memory:
data_time: 0.08 --> 0.0002
time: no big difference

It could be concluded that:
~~1. In GPU, computation time is large enough to overlap gpu data time (unpinned CPU memory --> GPU) when pin_memory=False.~~   No overlap here.
2. What interesting is that when using pin_memory=True, the data_time decreases from 0.08 to 0.0002, which implies that the torch.utils.data.DataLoader could be optimized when pin_memory=False.

```
def pin_memory_batch_collator(batch):
    """
    A naive pin memory batch collator.
    """
    for sample in batch:
        sample['image'] = sample['image'].pip_memory()
    return batch
```
This way will make use lots of extra GPU memory, do you know why?

PR #650, opened 2020-01-09
add model arguments to densepose apply_net.py

shenberg:
Added the ability to provide additional model arguments to densepose's `apply_net.py`, principally in order to be able to run it on CPU. Small copy-paste from `demo/demo.py`


shenberg:
Signed CLA

PR #672, opened 2020-01-10
Refactor test_boxes, test_model_zoo, test_rpn

kondela:
I've noticed that there is some inconsistency between different tests, i.e. some use python's builtin `assert` while others use `unittest.TestCase` asserts (a clear preference is for the latter). I went through `test_boxes.py`, `test_model_zoo.py`, and `test_rpn.py` and refactored it. Also some asserting did not work so I fixed that as well + did some minor refactoring in general.


ppwwyyxx:
the changes look good. could you fix the linter tests?

PR #677, opened 2020-01-11
image.shape in mapper doc

ArutyunovG:


ppwwyyxx:
the shape is correct.

ArutyunovG:
It was not working for me.
What's the difference with
https://github.com/facebookresearch/detectron2/blob/555200fa414e7cd35ad35a00a3e98f0545f5832e/detectron2/data/dataset_mapper.py#L96

ppwwyyxx:
sorry that I misread it!

PR #717, opened 2020-01-17
COCO AP(@.5) evaluation & group normalization error

cgh1392:
I added AP(@.5) metric per category for COCO format data and saving metrics.
Group normalization(ex: cascade-X152) makes error when input is empty(as I guess), so I made some ugly modification to wrappers.py


ppwwyyxx:
(1) we definitely do not accept copy-pasting a whole file just to change a few lines
(2) the fix is incorrect. Please refer to other code in `wrappers.py` that addresses empty input issue.

Issue #72, opened 2019-10-15
Support for mixed precision training 


sadanand-singh:
## ❓

Is there any plan to support mixed precision training soon? It was already being supported in the maskrcnn-benchmark code.

lbin:
@ppwwyyxx  I would like to send a PR for apex training. https://github.com/lbin/detectron2/blob/apex/projects/ApexTrainer/README.md

blueardour:
@lbin Hi, Thanks for the great contribution.

I recently also focused in the FP16 training. After checking your code, it seemed that only the amp.initialize and amp.scale_loss APIs are employed. With my experience in image classification scenario, the DistributedDataParallel should also be replaced by from apex.parallel import DistributedDataParallel as DDP. However, I noticed you comment out line 75 in apex_trainer.py.  Is there any reason why not using the amp DDP?

BTW, I'm also confused by the data in page https://github.com/lbin/detectron2/blob/apex/projects/ApexTrainer/README.md   
The speed and memory cost are higher for the ApexTrainer case, which I expected to be lower than the FP32 version.

Appreciated if you can give some clarification.   Thanks

lbin:
@blueardour I used batchsize 32 not 16, so the speed and memory are higher in my case.

for ddp:
https://github.com/NVIDIA/apex/tree/master/examples/imagenet#distributed-training

blueardour:
@lbin Thanks for the tips.

meetshah1995:
Given that Pytorch natively supports mixed precision training, is there plan to integrate that?

indigoviolet:
At least for RetinaNet, it was quite straightforward to use mixed precision training. You can see how I did it: https://github.com/indigoviolet/detectron2/commit/d175f4a823cfdcce417e5ea38acbabc5215b9294

lbin:
@indigoviolet I think we should move to pytorch 1.6

ppwwyyxx:
We plan to add support this year. https://github.com/facebookresearch/detectron2/pull/925 is a good example of how to do it with apex.

alexge233:
@ppwwyyxx Hi, sorry to ask again, can I force only `fp16` for models and training, or is this still work in progress?

PR #720, opened 2020-01-20
add apex trainer example project

lbin:
#72 
We Create a ApexTrainer based on SimpleTrainer, so we could use much more batch size, for example, in retinanet_r50, we could use 32 and got 36.7( box AP)


ppwwyyxx:
Thanks for the PR! After a quick look it seems like a pretty nice addition to me. I'll have time to look at it more thoroughly some time next week.

blueardour:
@lbin Hi, Thanks for the great contribution.

I recently also focused in the FP16 training. After checking your code, it seemed that only the amp.initialize and amp.scale_loss APIs are employed. With my experience in image classification scenario, the DistributedDataParallel should also be replaced by from apex.parallel import DistributedDataParallel as DDP. However, I noticed you comment out line 75 in apex_trainer.py. Is there any reason why not using the amp DDP?

soldierofhell:
@blueardour, NVIDIA doesn't recommend to use their DDP anymore, see e.g. https://github.com/NVIDIA/apex/tree/master/examples/imagenet#distributed-training

Has anyone tried to patch [Mask X152](https://github.com/facebookresearch/detectron2/blob/master/configs/Misc/cascade_mask_rcnn_X_152_32x8d_FPN_IN5k_gn_dconv.yaml) model. I did it (by forcing roialign and deconv to 32 bit) but unfortunately see no memory gains in opt_level=O1 (O2 doesn't work at all)?

soldierofhell:
@ppwwyyxx, @lbin as for the PR I think it would be better to integrate apex (as an option) in current trainers classes. Creating separete class for 3 lines of code will lead to unnecessary maintenance for apex trainer class

lbin:
@soldierofhell this is an example project for detectron2, not only for fp16 or apex, your suggestion is my next step.

sharathts:
@lbin I notice that training with AMP matches the fp32 loss curve. But evaluating aafter training gives me bad mAP (~9). Reevaluating on amp saved checkpoint gives me good mAP. 

Did you encounter anything similar?

lbin:
> @lbin I notice that training with AMP matches the fp32 loss curve. But evaluating aafter training gives me bad mAP (~9). Reevaluating on amp saved checkpoint gives me good mAP.
> 
> Did you encounter anything similar?

@sharathts you should try this https://github.com/facebookresearch/detectron2/pull/925

CarryHJR:
@lbin in your commit, I find there isn't any code changes during saving model, in my experience, it seems we need extra codes when saving model

batrlatom:
I tried your apex code with maskrcnn, but it needs more research. I am getting errors. Do you know how to modify maskrcnn yaml file to make it work ( https://github.com/facebookresearch/detectron2/blob/master/configs/Base-RCNN-FPN.yaml ) ?

PR #796, opened 2020-02-03
Install opencv via pip instead of apt-get, to have the most current version.

### LINKED ISSUE #87: Failed to load OpenCL runtime
Although it can train successfully, but does anyone know how can I solve the bug?

## ❓ Questions and Help

General questions about detectron2.

NOTE:

If you met any unexpected issue when using detectron2 and wish to know why,
please use the "Unexpected Problems / Bugs" issue template.

`
Failed to load OpenCL runtime
Command Line Args: Namespace(config_file='configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml', dist_url='tcp://127.0.0.1:49152', eval_only=False, machine_rank=0, num_gpus=8, num_machines=1, opts=
[], resume=False)
Failed to load OpenCL runtime
Failed to load OpenCL runtime
Failed to load OpenCL runtime
Failed to load OpenCL runtime
Failed to load OpenCL runtime
Failed to load OpenCL runtime
Failed to load OpenCL runtime
Failed to load OpenCL runtime
Config 'configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml' has no VERSION. Assuming it to be compatible with latest v2.
Config 'configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml' has no VERSION. Assuming it to be compatible with latest v2.
Config 'configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml' has no VERSION. Assuming it to be compatible with latest v2.
Config 'configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml' has no VERSION. Assuming it to be compatible with latest v2.
Config 'configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml' has no VERSION. Assuming it to be compatible with latest v2.
Config 'configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml' has no VERSION. Assuming it to be compatible with latest v2.
Config 'configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml' has no VERSION. Assuming it to be compatible with latest v2.
Config 'configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml' has no VERSION. Assuming it to be compatible with latest v2.
[10/16 06:32:13 detectron2]: Rank of current process: 0. World size: 8
[10/16 06:32:13 detectron2]: Environment info:
---------------------  --------------------------------------------------
Python                 3.6.8 (default, Oct  7 2019, 12:59:55) [GCC 8.3.0]
Detectron2 Compiler    GCC 7.4
DETECTRON2_ENV_MODULE  <not set>
PyTorch                1.3.0
PyTorch Debug Build    False
CUDA available         True
GPU 0,1,2,3,4,5,6,7    Tesla P40
Pillow                 6.2.0
cv2                    3.2.0
`

maxfrei750:
Fixes #87. 

Although it would be reasonable to remove the apt-get installation of open cv, doing this results in an error when opencv is imported.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -19,7 +19,7 @@ RUN wget https://bootstrap.pypa.io/get-pip.py && \
 
 # install dependencies
 # See https://pytorch.org/ for other options if you use a different version of CUDA
-RUN pip install --user torch torchvision tensorboard cython
+RUN pip install --user torch torchvision tensorboard cython opencv-python
 RUN pip install --user 'git+https://github.com/cocodataset/cocoapi.git#subdirectory=PythonAPI'
 
 RUN pip install --user 'git+https://github.com/facebookresearch/fvcore'


ppwwyyxx:
The following code works originally in docker, but seems to fail after your change.
```python
import numpy as np
import cv2

h, w = 100, 100
of = cv2.VideoWriter(
    "output.mkv",
    fourcc=cv2.VideoWriter_fourcc(*"x264"),
    fps=float(30),
    frameSize=(w, h),
    isColor=True)


for k in range(100):
    fr = np.random.rand(h, w, 3) * 255
    of.write(fr.astype("uint8"))
of.release()
```

Usually I don't recommend `opencv-python` due to weird issues it has.

maxfrei750:
I see. Can you recommend another, more stable, way to obtain a more recent opencv version, e.g. compiling from source?

ppwwyyxx:
Compiling from source would work, but that's too much for the Dockerfile. Some unofficial ubuntu PPAs might work, though they have quite some uncertainty as well.

maxfrei750:
> that's too much for the Dockerfile

Do you mean too much to maintain, too much building time or too many lines? Or simply all of it?

Because the building does not seem to be too complicated:
https://linuxize.com/post/how-to-install-opencv-on-ubuntu-18-04/

However, I didn't test it so far and I agree that it maybe a little overkill just to get rid of a warning.

ppwwyyxx:
All of it :rofl: . That's quite an overkill.

maxfrei750:
😄 Ok, agreed. Then I'd say we close this PR, unless you have another solution for the warning.

ppwwyyxx:
OK, closing and thanks for helping with the issue.

PR #809, opened 2020-02-05
fix bug in coco dataset categories mapping

bendidi:
The function `convert_to_coco_dict` doesn't currently take into account whether there is a `thing_dataset_id_to_contiguous_id` mapping in metadata, which is inconsistent with the behavior of `COCOEvaluator` when generating coco format predictions (uses `thing_dataset_id_to_contiguous_id` for category mapping)

### DIFF
--- a/detectron2/data/datasets/coco.py
+++ b/detectron2/data/datasets/coco.py
@@ -294,14 +294,11 @@ def convert_to_coco_dict(dataset_name):
     # unmap the category ids for COCO
     if hasattr(metadata, "thing_dataset_id_to_contiguous_id"):
         reverse_id_mapping = {v: k for k, v in metadata.thing_dataset_id_to_contiguous_id.items()}
-        for dataset_dict in dataset_dicts:
-            dataset_dict["category_id"] = reverse_id_mapping[dataset_dict["category_id"]]
         categories = [
             {"id": reverse_id_mapping[id], "name": name}
             for id, name in enumerate(metadata.thing_classes)
         ]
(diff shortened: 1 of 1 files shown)


bendidi:
another reason for this MR, is that most custom dataset will want to set up `thing_dataset_id_to_contiguous_id` if they want to use coco evaluator, because for coco category ids are 1-indexed while they are 0-indexed in d2

ppwwyyxx:
> another reason for this MR, is that most custom dataset will want to set up thing_dataset_id_to_contiguous_id

I'm not able to understand this. Could you explain what issues will appear if a custom dataset does not have this attribute but uses COCOEvaluator? I would expect it to work the same.

bendidi:
my bad, I got it confused with this [line] 😅 (https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocotools/cocoeval.py#L39) in cocoEval api, (assumed categories should be 1-indexed not annotations) 

so yes as you've said, if a custom dataset does not have this attribute, it should still work the same.

Thought I still think that `convert_to_coco_dict` should have the same behavior as COCOEvaluator when generating the coco dataset.

ppwwyyxx:
Could you fix the linter warnings shown in https://circleci.com/gh/facebookresearch/detectron2/328?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link ?

For this specific case it seems ok to just use #noqa

bendidi:
fixed

PR #813, opened 2020-02-06
fix filtering of foreground indexes

xmyqsh:


xmyqsh:
Close. Can be lazy process by [select_foreground_proposals](https://github.com/facebookresearch/detectron2/blob/master/detectron2/modeling/roi_heads/roi_heads.py#L666)

PR #828, opened 2020-02-10
Added path for demo input

klintan:
Added ability to pass a path to the demo script. Something I needed to be able to more conveniently run several files.


ppwwyyxx:
You can already use `directory/*.jpg` or something similar according to line 78.

All scripts are examples for using the library and you can follow them to implement your own if you need more complicated logic.

PR #862, opened 2020-02-13
Optimize summing of losses in run_step() & Fix typos

kondela:
I noticed that the `run_step()` method uses `sum(loss for loss in loss_dict.values())` which could be simplified to `sum(loss_dict.values())`. A side benefit is faster execution.
Tested locally with a `loss_dict` consisting of 32 losses:

Proposed: `482 ns ± 53.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)`
Original: `2.02 µs ± 77.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)`

Plus I fixed some typos and an assertion along the way :)

### DIFF
--- a/tools/plain_train_net.py
+++ b/tools/plain_train_net.py
@@ -153,7 +153,7 @@ def do_train(cfg, model, resume=False):
             storage.step()
 
             loss_dict = model(data)
-            losses = sum(loss for loss in loss_dict.values())
+            losses = sum(loss_dict.values())
             assert torch.isfinite(losses).all(), loss_dict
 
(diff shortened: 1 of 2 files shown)


tools/plain_train_net.py
```diff
@@ -153,8 +153,9 @@ def do_train(cfg, model, resume=False):
             storage.step()
 
             loss_dict = model(data)
-            losses = sum(loss for loss in loss_dict.values())
-            assert torch.isfinite(losses).all(), loss_dict
```
ppwwyyxx: it means "print loss_dict if the assertion failed". It's not an assertion on two statements.

kondela: You are right, I reverted the commit.

PR #866, opened 2020-02-13
installation: conda environment with all dependencies

breznak:
simply run
`conda env create (-f environment.yml)`
to setup all dependencies.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,24 @@
+name: detectron2-env
+channels:
+  - pytorch
+  - defaults
+dependencies:
+  - python==3.7
+  - pip
+  - cython
+  - pytorch::pytorch
+  - pytorch::torchvision
+  - pyyaml==5.1
+  - gcc_linux-64 ## gcc, g++ only needed for source installs of Detectron
+  - gxx_linux-64
+  - git
+  - wget
+  - unzip
+  - matplotlib
+  - conda-forge::pycocotools
+  - pip:
+    - opencv-contrib-python
+    ## install Detectron v2 itself:
+    #- git+https://github.com/facebookresearch/detectron2.git ## source build (needed if you use custom compiled pytorch etc) 
+    #or run: pip install detectron2 -f https://dl.fbaipublicfiles.com/detectron2/wheels/cu101/index.html # replace cu101 with cu100,90, or cpu. 
+
--- a/environment.yml
+++ b/environment.yml
@@ -15,9 +15,12 @@ dependencies:
   - wget
   - unzip
   - matplotlib
-  - conda-forge::pycocotools
+  - conda-forge::pycocotools>=2.0
+  - conda-forge::opencv>=4.1
   - pip:
-    - opencv-contrib-python
+    - tensorboard
+    - tensorflow-gpu
+    - PIL #for displaying images over ssh
     ## install Detectron v2 itself:
     #- git+https://github.com/facebookresearch/detectron2.git ## source build (needed if you use custom compiled pytorch etc) 
     #or run: pip install detectron2 -f https://dl.fbaipublicfiles.com/detectron2/wheels/cu101/index.html # replace cu101 with cu100,90, or cpu. 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,6 +6,10 @@ The [Dockerfile](https://github.com/facebookresearch/detectron2/blob/master/dock
 also installs detectron2 with a few simple commands.
 
 ### Requirements
+- You can use convenient [Anaconda Python](https://www.anaconda.com/distribution/#download-section) environment to 
+manage all dependencies (and/or source install): `conda env create`
+
+Or install yourself: 
 - Linux or macOS with Python ≥ 3.6
 - PyTorch ≥ 1.3
 - [torchvision](https://github.com/pytorch/vision/) that matches the PyTorch installation.


environment.yml
```diff
@@ -0,0 +1,24 @@
...
+  - git
+  - wget
+  - unzip
+  - matplotlib
+  - conda-forge::pycocotools
+  - pip:
+    - opencv-contrib-python
+    ## install Detectron v2 itself:
+    #- git+https://github.com/facebookresearch/detectron2.git ## source build (needed if you use custom compiled pytorch etc) 
+    #or run: pip install detectron2 -f https://dl.fbaipublicfiles.com/detectron2/wheels/cu101/index.html # replace cu101 with cu100,90, or cpu.
```
ppwwyyxx: This yaml config does not specify a cuda version for pytorch/torchvision, etc, so my guess is that it will not always pick the cuda version that the user had on the machine, which is going to cause problems.

environment.yml
```diff
@@ -0,0 +1,24 @@
...
+  - pytorch::pytorch
+  - pytorch::torchvision
+  - pyyaml==5.1
+  - gcc_linux-64 ## gcc, g++ only needed for source installs of Detectron
+  - gxx_linux-64
+  - git
+  - wget
+  - unzip
+  - matplotlib
+  - conda-forge::pycocotools
```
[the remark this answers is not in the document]
breznak: is PyPI version any better? Both are at 2.0.0, from 3/2018. 
On the other hand, the code seems to work with such version..(or atleast the demo) 
https://pypi.org/project/pycocotools/

environment.yml
```diff
@@ -0,0 +1,24 @@
...
+  - pip:
+    - opencv-contrib-python
```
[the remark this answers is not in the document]
breznak: opencv from conda pkgs/main seems to have some problems with Gtk2 on Ubuntu, is `conda-forge::opencv>=4.0` acceptable? Works for me

environment.yml
breznak: > This yaml config does not specify a cuda version for pytorch/torchvision, etc, so my guess is that it will not always pick the cuda version that the user had on the machine

actually that's the point, afaik conda will try to resolve and pick the version it found on the machine. I could easily force the `10.1` version in those deps, but I chose to let it open so the user can have whichever they use.

environment.yml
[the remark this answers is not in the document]
breznak: From pip we can install directly from `pip install git+https://github.com/cocodataset/cocoapi.git#subdirectory=PythonAPI` but conda's pip in environment.yml does not seem to handle that.

breznak:
Thank you for a swift review, @ppwwyyxx :wave: 

I see your point with potential issues in outdated or unofficial versions of packages. 
On the other hand, I've been contacted by my colleagues who struggled to install Detectron, so this `conda env` tries to provide an alternative easy install for the source version. 
(if any users run into weird struggles, they can be always suggested to try the docker / manuall source install). 
What do you think?

environment.yml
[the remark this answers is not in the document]
ppwwyyxx: The PyPI version is even older.

The conda version will not work with latest numpy.

environment.yml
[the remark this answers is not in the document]
ppwwyyxx: Most python packages of opencv have problems and the safest option is to use apt install or whatever the system provides. That's why there is no instructions to install opencv in readme.

Maybe `conda-forge::opencv>=4.0` is OK, but I'll need to try it to see.

environment.yml
ppwwyyxx: > afaik conda will try to resolve and pick the version it found on the machine

I am not a conda user but this surprises me. Actually I doubt it's actually true because guessing what cuda version to use (especially if there are multiple cuda installed in the system) is not a reasonable thing for conda to do. Unless we know how conda selects the cuda version to use, we should allow users to specify which version to use.

Forcing a 10.1 version is only going to cause more environment issues for users without 10.1.

breznak: I had just cuda 10.1 installed and this seemed to work, I'm testing with multiple cuda versions (although that might not be a typical config)

ppwwyyxx: I think pytorch defaults to use 10.1 so it's reasonable this works on 10.1.
I'm more interested to know whether it would automatically install the cuda10 version of pytorch (which should be confirmed by `print(torch.__config__.show())`)

breznak: > to know whether it would automatically install the cuda10 version of pytorch (which should be confirmed by print(torch.__config__.show())) 

I was going to test with cuda-10-2, so should I rather try cuda-10-0?

ppwwyyxx: Pytorch doesn't have packages for cuda 10.2: https://anaconda.org/pytorch/pytorch/files

So if you try cuda 10.2 and it still successfully installs - then it must be doing things incorrectly.

breznak: > So if you try cuda 10.2 and it still successfully installs - then it must be doing things incorrectly 

yep, you're right, conda does not check which/if CUDA is installed at all on the "host" system. So the env creates. 

OT: funnily, this still runs with cuda-10-2, even though pytorch and the wheel package are for 10.1 only. 

How can we go about this, then?

ppwwyyxx:
Closing as there is no activity and problems in the PR have not been addressed

PR #869, opened 2020-02-14
fix and enhance COMMON_STRIDE handling

xmyqsh:
fix and enhance MODEL.SEM_SEG_HEAD.COMMON_STRIDE handling


ppwwyyxx:
It's unclear what issue it is "fixing" (and with no test on what is fixed), so closing.

PR #870, opened 2020-02-14
remove redundant upsampling and reduce semantic_seg loss computation

xmyqsh:
remove redundant upsampling, sem_seg_postprocess will do it later
reduce semantic_seg loss computation, and
make it compatible with PointRend


ppwwyyxx:
> remove redundant upsampling, sem_seg_postprocess will do it later

It's not redundant and `sem_seg_postprocess` does not do the same thing.

The code is also different now and the PR no longer applies.

PR #879, opened 2020-02-15
clarify the meaning of uncertainty with comment

xmyqsh:
The thought of the original comment is incorrect, though the example it provided is no problem.


ppwwyyxx:
The original comment is correct, and we also clarified it a bit more recently.

PR #880, opened 2020-02-15
fix the sampling process to align the accurate math meaning

xmyqsh:
The fixed sampling process describes the accurate meaning of `oversample_ratio` and `importance_sample_ratio`.
NOTICE: I'm not sure the original implementation is same as here or not.
               The optimized hyperparameter of `oversample_ratio` and `importance_sample_ratio` may be re-grid-search again.


alexander-kirillov:
Dear @xmyqsh, the current implementation of the sampling process exactly matches one described in the [paper](https://arxiv.org/abs/1912.08193). It is simpler (less lines) and slightly faster. Hence, we will stick to the original implementation for now.

PR #890, opened 2020-02-17
Training visualisation for RetinaNet 

RaymondKirk:
Purely additive PR, added a training visualization option to the RetinaNet class, similar to the rcnn that allows logging of best scoring predictions every cfg.VIS_PERIOD storage iters.

### DIFF
--- a/detectron2/modeling/meta_arch/retinanet.py
+++ b/detectron2/modeling/meta_arch/retinanet.py
@@ -138,7 +138,7 @@ class RetinaNet(nn.Module):
         prop_img = v_pred.get_image()
         vis_img = np.vstack((anno_img, prop_img))
         vis_img = vis_img.transpose(2, 0, 1)
-        vis_name = f"Left: GT bounding boxes; Right: {max_boxes} Highest Scoring Results"
+        vis_name = f"Top: GT bounding boxes; Bottom: {max_boxes} Highest Scoring Results"
         storage.put_image(vis_name, vis_img)
 
     def forward(self, batched_inputs):
--- a/detectron2/modeling/meta_arch/retinanet.py
+++ b/detectron2/modeling/meta_arch/retinanet.py
@@ -8,6 +8,7 @@ from torch import nn
 
 from detectron2.layers import ShapeSpec, batched_nms, cat
 from detectron2.structures import Boxes, ImageList, Instances, pairwise_iou
+from detectron2.utils.events import get_event_storage
 from detectron2.utils.logger import log_first_n
 
 from ..anchor_generator import build_anchor_generator
@@ -18,6 +19,7 @@ from ..postprocessing import detector_postprocess
 from .build import META_ARCH_REGISTRY
 
 __all__ = ["RetinaNet"]
+
 
 
 def permute_to_N_HWA_K(tensor, K):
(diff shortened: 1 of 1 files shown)


detectron2/modeling/meta_arch/retinanet.py
```diff
@@ -92,12 +93,53 @@ def __init__(self, cfg):
...
+        A function used to visualize images and intermidiate network results. It shows ground truth
+        bounding boxes on the original image and up to 20 predicted object
+        bounding boxes on the original image.
+
+        Args:
+            batched_inputs (list): a list that contains input to the model.
+            results (List[Instances]): a list of #images elements.
+        """
+        from detectron2.utils.visualizer import Visualizer
+        import numpy as np
```
ppwwyyxx: `Visualizer` is imported here because it uses opencv, which is not a required dependency for training.
`import numpy` should be moved to beginning of file.

detectron2/modeling/meta_arch/retinanet.py
```diff
@@ -92,12 +93,53 @@ def __init__(self, cfg):
...
+        self.input_format = cfg.INPUT.FORMAT
 
         pixel_mean = torch.Tensor(cfg.MODEL.PIXEL_MEAN).to(self.device).view(3, 1, 1)
         pixel_std = torch.Tensor(cfg.MODEL.PIXEL_STD).to(self.device).view(3, 1, 1)
         self.normalizer = lambda x: (x - pixel_mean) / pixel_std
         self.to(self.device)
 
+    def visualize_training(self, batched_inputs, results):
+        """
+        A function used to visualize images and intermidiate network results. It shows ground truth
```
ppwwyyxx: not "intermediate" -- it visualizes the final predictions

detectron2/modeling/meta_arch/retinanet.py
```diff
@@ -92,12 +93,53 @@ def __init__(self, cfg):
...
+        assert img.shape[0] == 3, "Images should have 3 channels."
+        if self.input_format == "BGR":
+            img = img[::-1, :, :]
+        img = img.transpose(1, 2, 0)
+        v_gt = Visualizer(img, None)
+        v_gt = v_gt.overlay_instances(boxes=batched_inputs[image_index]["instances"].gt_boxes)
+        anno_img = v_gt.get_image()
+        processed_results = detector_postprocess(results[image_index], img.shape[0], img.shape[1])
+        predicted_boxes = processed_results.pred_boxes.tensor.detach().cpu().numpy()
+        box_size = min(len(predicted_boxes), max_vis_prop)
```
ppwwyyxx: no need to use `min()`. `boxes[:N]`  will not lead to overflow error for large N.

detectron2/modeling/meta_arch/retinanet.py
```diff
@@ -92,12 +93,53 @@ def __init__(self, cfg):
...
+            results (List[Instances]): a list of #images elements.
+        """
+        from detectron2.utils.visualizer import Visualizer
+        import numpy as np
+
+        assert len(batched_inputs) == len(
+            results
+        ), "Cannot visualize inputs and results of different sizes"
+        storage = get_event_storage()
+        max_vis_prop = 20
```
ppwwyyxx: It's named "prop" in R-CNN for "proposals". The name does not make sense here anymore. Maybe call it "max_boxes" or "max_results".

detectron2/modeling/meta_arch/retinanet.py
```diff
@@ -92,12 +93,53 @@ def __init__(self, cfg):
             cfg.MODEL.RETINANET.IOU_LABELS,
             allow_low_quality_matches=True,
         )
+        self.vis_period = cfg.VIS_PERIOD
+        self.input_format = cfg.INPUT.FORMAT
```
ppwwyyxx: To have a consistent style, move them above in the `self.xxx = cfg.xxx` block and align the "=" sign.

detectron2/modeling/meta_arch/retinanet.py
RaymondKirk: Fixed in 45d9600

detectron2/modeling/meta_arch/retinanet.py
RaymondKirk: Fixed in 45d9600

detectron2/modeling/meta_arch/retinanet.py
RaymondKirk: Fixed in 45d9600

detectron2/modeling/meta_arch/retinanet.py
RaymondKirk: Fixed in 45d9600

RaymondKirk:
All comments have been addressed, ready for merge?

ppwwyyxx:
The PR looks good. As part of the test plan - could you provide a screenshot how it looks like in tensorboard?

RaymondKirk:
For a project with 16:9 images, it looks like this:
 
image

If images are more likely to be taller than wider I'd suggest changing line https://github.com/facebookresearch/detectron2/pull/890/files/8dd7e60f2ba7b566378274dad400e65484146a6f#diff-ca56fc273f4f8c69ec8a0ae22af3092fR138 to `vis_img = np.hstack((anno_img, prop_img))` instead, however tensor board has a very small cut-off for the max width of each image element.

ppwwyyxx:
Looks like the name should be changed to "Top:   Bottom:  " instead of "left right".

RaymondKirk:
> Looks like the name should be changed to "Top: Bottom: " instead of "left right".

Done

PR #900, opened 2020-02-19
add comment to PointRend sampling process

xmyqsh:


alexander-kirillov:
Dear @xmyqsh, 

Thank you very much for being so rigorous with our codebase!

You are absolutely correct that the importance of the order between the interpolation and uncertainty computation is natural and obvious for anyone who would think about it. At the same time, I know that when I was re-implementing this algorithm, I tried to optimize the code and forgot about this issue. As the result, I made an error there. So, I added this comment in case someone else would not think about this obvious problem to help other people who may do the same thing.

PR #925, opened 2020-02-23
add apex fp16 trainer

lbin:
trained with 8*2080ti
### faster rcnn
lr sched 1x 
| Name | train time(s/iter)|infer time(s/im)|train mem(GB)|box AP|
|:------:|:------:|:------:|:------:|:------:|
|R50-FPN-V100| 0.210|	0.055|3.0|37.9|
|R50-FPN|0.312 |0.056 | 2.893| 37.9|
|R50-FPN-Apex|0.322|0.053|2.793|38.0|

### RetineNet
lr sched 1x 
| Name | train time(s/iter)|infer time(s/im)|train mem(GB)|box AP|
|:------:|:------:|:------:|:------:|:------:|
|R50-V100|0.200|	0.062	|3.9	|36.5|
|R50| 0.285|0.079 |3.934 | 36.5|
|R50-Apex|0.248|0.067|3.275|36.4|

### Mask R-CNN
lr sched 1x 
| Name | train time(s/iter)|infer time(s/im)|train mem(GB)|box AP|mask AP|
|:------:|:------:|:------:|:------:|:------:|:------:|
|R50-V100|0.261|	0.053	|3.4	|38.6|35.2|
|R50-FPN|0.348 |0.063 | 3.44|38.5|35.1|
|R50-FPN-Apex|0.363|0.061|3.26|38.6|35.2|


ppwwyyxx:
Do you know what was the previous issue the cause RetinaNet to perform poor in evaluation?

lbin:
> Do you know what was the previous issue that cause RetinaNet to perform poor in evaluation?

we should use fp32(float_function) for nms or batch_nms
https://github.com/facebookresearch/detectron2/pull/925/files#diff-6b2545951b0ccc03b243cecc5c198d33R10

erickim555:
Hi there, for your benchmarks, does `R50-FPN` use the same batchsize as `R50-FPN-Apex`?

I tried out this patch on my end, and found that using Apex (O1) with a ResNeXt101-FRCNN-FPN model yielded slightly worse train throughput (vs no Apex), as well as slightly higher GPU memory usage. I tested on a machine with 8 Tesla-V100's, specifically the aws ec2 p3.16xlarge instance.

ppwwyyxx:
FYI, since pytorch starts to support AMP since 1.5 (https://pytorch.org/docs/master/amp.html), we would like to use its native support rather than apex. 
At this moment it's unclear whether there will be any issues with pytorch's AMP since it's relatively new, so this PR would be very useful as a reference.

erickim555:
FYI: adding the `@float_function` decorators to these functions ends up breaking torch scripting (as of several months ago). Not sure if there's a simple fix, but given the modest gains (actually, worse) train throughput I got for training X101-FRCNN-FPN models in fp16 (O1), I'm not sure if it's worth pursuing fp16 + Detectron2 training for now. Maybe someone else observed differently?

riskySneep (approved):

PR #975, opened 2020-03-03
Make FPN work when exported to TensorRT

corysaildrone:
I have been working towards exporting RetinaNet to ONNX and then TensorRT. I use the onnx2trt tool to convert from the ONNX model to TensorRT, but the `F.interpolate` layers in the FPN cause the conversion to fail with an error like:
```While parsing node number 127 [Gather -> "487"]:
ERROR: /home/jack/dev3/TensorRT/parsers/onnx/onnx2trt_utils.hpp:412 In function convert_axis:
[8] Assertion failed: axis >= 0 && axis < nbDims
```
This change seems harmless for all other purposes and fixes the error. Let me know what you think as it's much easier to make the fix upstream than in my own code.

### DIFF
--- a/detectron2/modeling/backbone/fpn.py
+++ b/detectron2/modeling/backbone/fpn.py
@@ -1,6 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 import math
 import fvcore.nn.weight_init as weight_init
+import torch
 import torch.nn.functional as F
 from torch import nn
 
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
We already handle this issue at https://github.com/facebookresearch/detectron2/blob/5e2a1ecccd228227c5a605c0a98d58e1b2db3640/detectron2/export/shared.py#L82-L84 so no interpolate will use scale_factor when exporting to onnx. Not sure why you would have the issue.

corysaildrone:
If I build an ONNX model with `torch.ops._caffe2.ResizeNearest` then it can't be converted to a tensorRT engine by NVIDIA's `onnx2trt` utility.

corysaildrone:
I'm happy to follow a similar pattern of using `mock` to replace the `interpolate` layer with what I have here if you don't think this is a good change to have in the main code body - it just seems like a lot of code compared to this change. I think this change could also help future users that want to use RetinaNet models in TensorRT.

ppwwyyxx:
Thanks for the clarification. In general it's hard for us to accommodate different runtimes - especially third-party runtime. Even if we accept the change we cannot guarantee it stays effective and up to date with limitations in different runtimes. Also, imagine that different runtimes could have different limitations - e.g. maybe `onnx2tensorflow` or `onnx2mxnet` would require the use of `scale_factor` rather than `size`.
This is why we use the more hacky `mock` mechanism for the caffe2 runtime and I would recommend you do the same.

LeonNerd:
hi,I want to convert pt to onnx,and then caffe.because my terminal only supports it,do you have any good suggestions?I see that you have completed the conversion to onnx. thanks

PR #996, opened 2020-03-05
Fixes for Rotated RPN (RRPN)

### LINKED ISSUE #21: Thanks for  great work, i have a question, how can i set config file to detect oriention target in faster rcnn or other network?
Thanks for  great work, i have a question, how can i set config file to detect oriention target in faster rcnn or other network?

breznak:
- [x] fixed inheritance and use of proper "Rotated" classes where needed
- [x] adds checks for valid params for RRPN 
- [ ] TODO custom `DataLoader` needed for angles (help wanted) 
- [ ] (optional) visualizations for rotated bounding boxes (code snips somewhere here on GH)  

Fixes #21


detectron2/modeling/anchor_generator.py
```diff
@@ -362,4 +362,8 @@ def build_anchor_generator(cfg, input_shape):
     Built an anchor generator from `cfg.MODEL.ANCHOR_GENERATOR.NAME`.
     """
     anchor_generator = cfg.MODEL.ANCHOR_GENERATOR.NAME
+    if cfg.MODEL.PROPOSAL_GENERATOR.NAME == "RRPN":
+        print("Forcing RotatedAnchorGenerator for RRPN")
+        anchor_generator = "RotatedAnchorGenerator"
```
breznak: "auto forced" correction

detectron2/modeling/proposal_generator/rrpn.py
```diff
@@ -25,7 +26,36 @@ class RRPN(RPN):
...
+        # RRPN enabled we provide the default values in correct dim (5, not 4).
+        # This is so people don't have to modify their configs on unrelated places.
+        weights = cfg.MODEL.RPN.BBOX_REG_WEIGHTS
+        if weights == (
+            1.0,
+            1.0,
+            1.0,
+            1.0,
+        ):  # 4-element, default values -> we can automatically change to correct dim
+            weights = (1, 1, 1, 1, 1)  # 5-elem
```
breznak: is this ok, or too just assert?

breznak:
> I'm not sure if this is the right place to discuss (or if it should be on your PR) but I have been running into a problem with your code. The error occurs when I am trying to train (I'm using some of the above config options as well as my own data loaded): 

@ryangooch thanks for testing out the code! Yea, let's continue the implementation specific details under the PR here. 

>  as well as my own data loader

would you share (ideally commit to this PR) the code for custom DataLoader and an example dataset (10 items, just to validate it runs)? That would be great :+1: 

> I suspect that the RRPN class in rrpn.py is trying to initialize the RPN class, which has in its __init __ block a call to Box2BoxTransform, which expects 4 weights, and receives the 5 weights from RRPN. 

good point! true, I'll see how to handle the call to super class (ideally we'd not send the weights in the constructor, or just some dummy?)

st7ma784:
To solve visualizing rotated bboxes... 

I dont know enough python to do an inline if statement in here
`boxes = [BoxMode.convert(x["bbox"], x["bbox_mode"], BoxMode.XYWHA_ABS) for x in annos]
`
as the default is `BoxMode.XYXY_ABS`... 

I presume its something like 
`boxes = [BoxMode.convert(x["bbox"], x["bbox_mode"], BoxMode.XYXY_ABS) if x["bbox_mode"] != BoxMode.XYWHA for x in annos]
`

but this works: 
```

import random
dataset_dicts = returnTRAINRECORDS()
print(ClassesTotals)
class myVisualizer(Visualizer):
  
    def draw_dataset_dict(self, dic):
        annos = dic.get("annotations", None)
        if annos:
            if "segmentation" in annos[0]:
                masks = [x["segmentation"] for x in annos]
            else:
                masks = None
            if "keypoints" in annos[0]:
                keypts = [x["keypoints"] for x in annos]
                keypts = np.array(keypts).reshape(len(annos), -1, 3)
            else:
                keypts = None

            boxes = [BoxMode.convert(x["bbox"], x["bbox_mode"], BoxMode.XYWHA_ABS) for x in annos]

            labels = [x["category_id"] for x in annos]
            names = self.metadata.get("thing_classes", None)
            if names:
                labels = [names[i] for i in labels]
            labels = [
                "{}".format(i) + ("|crowd" if a.get("iscrowd", 0) else "")
                for i, a in zip(labels, annos)
            ]
            self.overlay_instances(labels=labels, boxes=boxes, masks=masks, keypoints=keypts)

        sem_seg = dic.get("sem_seg", None)
        if sem_seg is None and "sem_seg_file_name" in dic:
            sem_seg = cv2.imread(dic["sem_seg_file_name"], cv2.IMREAD_GRAYSCALE)
        if sem_seg is not None:
            self.draw_sem_seg(sem_seg, area_threshold=0, alpha=0.5)
        return self.output
for d in random.sample(dataset_dicts, 3):
    e=dict(d)
    name=e.get("file_name")
    print(name)
    print(e.items())
    img = cv2.imread(name)
    visualizer = myVisualizer(img[:, :, ::-1], metadata={}, scale=1)
    vis = visualizer.draw_dataset_dict(e, )
    cv2_imshow(vis.get_image()[:, :, ::-1])
```

ryangooch:
@breznak Thanks for responding! I want to make sure I've ironed out all the issues but I'll try to pull together a smaller example with what I've done (using @st7ma784 code too) after that.

ppwwyyxx:
The work appears stuck for >1 year so closing.

PR #998, opened 2020-03-06
Add linear wrapper because of pytorch bug

lsrock1:
Because of [Linear with empty tensor backward error](https://github.com/pytorch/pytorch/issues/34202)
I made a wrapper for linear layer. This bug is occurred version 1.3 and 1.4 (not > 1.4).

### DIFF
--- a/detectron2/layers/wrappers.py
+++ b/detectron2/layers/wrappers.py
@@ -161,9 +161,6 @@ else:
         Because of https://github.com/pytorch/pytorch/issues/34202
         """
 
-        def __init__(self, *args, **kwargs):
-            super().__init__(*args, **kwargs)
-
         def forward(self, x):
             if x.numel() == 0:
                 output_shape = [x.shape[0], self.weight.shape[0]]
--- a/detectron2/layers/wrappers.py
+++ b/detectron2/layers/wrappers.py
@@ -154,7 +154,7 @@ else:
 if TORCH_VERSION > (1, 4):
     Linear = torch.nn.Linear
 else:
-    
+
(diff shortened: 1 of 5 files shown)


detectron2/layers/wrappers.py
```diff
@@ -151,6 +151,36 @@ def forward(self, x):
...
+
+    class Linear(torch.nn.Linear):
+        """
+        A wrapper around :class:`torch.nn.Linear` to support empty inputs and more features.
+        Because of https://github.com/pytorch/pytorch/issues/34202
+        """
+
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+
```
ppwwyyxx: can be removed since it's the same as base class

detectron2/modeling/backbone/resnet.py
```diff
@@ -361,7 +362,7 @@ def __init__(self, stem, stages, num_classes=None, out_features=None):
 
         if num_classes is not None:
             self.avgpool = nn.AdaptiveAvgPool2d((1, 1))
-            self.linear = nn.Linear(curr_channels, num_classes)
+            self.linear = Linear(curr_channels, num_classes)
```
ppwwyyxx: could you undo this since we know it will not happen to backbone.
We're eventually going to change back in the future anyway when moving to higher version of pytorch.

lsrock1: I thought linear should be also replaced because conv2d is replaced, i will undo this!

lsrock1:
@ppwwyyxx updated!

PR #1073, opened 2020-03-19
Enabling detectron2 on AMD's ROCM stack

hgaspar:
Following the changes in pytorch to enable extensions, and in torchvision, to enable it as an extension on AMD's rocm stack, this pull request follows the same pattern of changes to also enable detectron2 on AMD's rocm stack.

### DIFF
--- a/detectron2/layers/csrc/box_iou_rotated/box_iou_rotated_utils.h
+++ b/detectron2/layers/csrc/box_iou_rotated/box_iou_rotated_utils.h
@@ -4,7 +4,7 @@
 #include <cassert>
 #include <cmath>
 
-#ifdef __CUDACC__
+#if defined(__CUDACC__) || __HCC__==1 ||  __HIP__==1
 // Designates functions callable from the host (CPU) and the device (GPU)
 #define HOST_DEVICE __host__ __device__
 #define HOST_DEVICE_INLINE HOST_DEVICE __forceinline__
@@ -192,7 +192,7 @@ HOST_DEVICE_INLINE int convex_hull_graham(
   // (essentially sorting according to angles)
   // If the angles are the same, sort according to their distance to origin
   T dist[24];
-#ifdef __CUDACC__
+#if defined(__CUDACC__) || __HCC__==1 || __HIP__==1
   // compute distance to origin before sort, and sort them together with the
   // points
   for (int i = 0; i < num_in; i++) {
--- a/detectron2/layers/csrc/cuda_version.cu
+++ b/detectron2/layers/csrc/cuda_version.cu
@@ -4,6 +4,23 @@
 
 namespace detectron2 {
 int get_cudart_version() {
+// Not a ROCM platform: Either HIP is not used, or
+// it is used, but platform is not ROCM (i.e. it is CUDA) 
+#if !defined(__HIP_PLATFORM_HCC__) 
   return CUDART_VERSION;
+#else
+  int version = 0;
+  
+#if HIP_VERSION_MAJOR != 0
+// Create a convention similar to that of CUDA, as assumed by other 
+// parts of the code.
+
+  version = HIP_VERSION_MINOR;
+  version += (HIP_VERSION_MAJOR*100);
+#else
+  hipRuntimeGetVersion( &version );  
+#endif
+  return version;
+#endif
 }
 } // namespace detectron2
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,10 @@ def get_extensions():
     is_rocm_pytorch = False
     if torch_ver >= [1, 5]:
         from torch.utils.cpp_extension import ROCM_HOME
-        is_rocm_pytorch = True if ((torch.version.hip is not None) and (ROCM_HOME is not None)) else False
+
+        is_rocm_pytorch = (
+            True if ((torch.version.hip is not None) and (ROCM_HOME is not None)) else False
+        )
 
     if is_rocm_pytorch:
         hipify_python.hipify(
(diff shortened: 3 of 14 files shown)


setup.py
```diff
@@ -43,28 +44,54 @@ def get_extensions():
...
+            show_detailed=True,
+            is_pytorch_extension=True)
+
+        source_cuda = glob.glob(path.join(extensions_dir, "**", 'hip', "*.hip")) + glob.glob(
+            path.join(extensions_dir, 'hip', "*.hip"))
+
+        shutil.copy("detectron2/layers/csrc/box_iou_rotated/box_iou_rotated_utils.h", 
+                    "detectron2/layers/csrc/box_iou_rotated/hip/box_iou_rotated_utils.h")
+        shutil.copy("detectron2/layers/csrc/deformable/deform_conv.h", 
+                    "detectron2/layers/csrc/deformable/hip/deform_conv.h")
```
ppwwyyxx: What is the reason for these hacks & can we avoid it with a better solutions?
I'm afraid that if we include them into the source code while we don't know what they mean or how to test them, they will easily become broken. For example I have no idea whether I should do something like this if I add a new C++ file.

hgaspar: Thank you, we are excited to contribute to this great project!

Let me describe the issue, and a possibly better long term solution:  As you know, for pytorch, and also torchvision, we have a "on the fly" compile time hipification of needed cuda code, meaning translating cuda code (e.g. .cu files), to hip code (which the can run on AMD GPUs.  Usually this is done in a "tidy" manner, meaning that if there is a "cuda" directory, then we create a "hip" directory at the same level.  This is very safe, in the sense that we are preserving the directory/path relationship between the cu files and header files that may be included by those cu files.  In that case, there is no need for any ugliness...

However, there are situation where cuda files just exist in a random, non canonically named directory.  In that case we still create a hip subdirectoy.  In that case however, obviously the relationship between the hipified device code, and the included header files changes (by adding a directory level between the hipified code and the included header files).

This is the reason for the explicit copies of header files, so that the header files are still found, and can be included.

The most elegant solution therefore, would be to have a codebase where all .cu files are in subdirectories named "cuda", and we can have whatever rich directory structure you would like. The emphasis is on the name "cuda".  In that case, the on-the-fly (compile time) hipification will just create a "hip" subdirectoy at the same level, will convert the cuda code for the cuda directory to hip code, and place it in the hip directory, and nothing will break, and no voodoo (copying of files) will be needed.  Arguably this will be the most tidy approach, even in principle, so that different platforms can have their own implementations, mirroring the same directory hierarchy.

Is this a direction you would like to pursue?   We can also start with the "hack" right now, so we can get it working on our gpus asap (a similar hack was accepted in at least two other projects, including torchvision), and we can have a plan for the (near?) future to slightly rearrange the directory structure as described above, to have a completely elegant solution.

ppwwyyxx: I'm not familiar with the HIP stack and thanks for the clarifications!  
If the goal is to make sure the relative positions between sources and headers stay unchanged, wouldn't it easier to just generate the hip code in the original directory, regardless of its name?

From reading https://github.com/pytorch/pytorch/blob/1f87f15ba3cd46fe5474f7783ff80ef06950e157/torch/utils/hipify/hipify_python.py#L487-L515, it seems that as long as our ".cu" files contain "cuda" in the file name, the new hip file will be generated under the same directory, therefore headers should be included correctly. 

Renaming the cuda files would be a smaller change than re-organizing the directory structure. Is that a reasonable solution?

ppwwyyxx: I just realized that we already have "cuda" in the filename of ".cu" files. It turns out that `hipify_python` does not do what the documentation says:
```
    if dirpath == orig_dirpath:
        dirpath = os.path.join(dirpath, 'hip')
```
it will always add "hip" directory as long as the directory name does not contain "cuda". This contradicts with its documentation and should probably be fixed in pytorch. After that it seems we don't need this hack.

detectron2/layers/csrc/box_iou_rotated/box_iou_rotated_utils.h
```diff
@@ -194,7 +194,7 @@ HOST_DEVICE_INLINE int convex_hull_graham(
     dist[i] = dot_2d<T>(q[i], q[i]);
   }
 
-#ifdef __CUDACC__
+#if defined(__CUDACC__) ||  __HIP_DEVICE_COMPILE__==1
```
sunway513, line 197: Hi @hgaspar , should we update here the same way as 6c216765478ad9eea97437b795fc294d1812c2c1?

setup.py
hgaspar: In commit
1679856eb56e8d236fdbd8164bfaf43150262bb6
I made the following change:  Instead of explicitly copying header files into generated "hip" directories, I make the needed directories (that contain those header files) known to the hipification script instead.  Does that sound sufficiently better?

Regarding putting the .hip files side by side to the .cu files, in the same directory, that indeed seems to require a change in pytorch.  However there is a downside to that approach, namely the git status (of git controlled directories) would be polluted with autogenerated files. That is the reason we try to avoid that situation...

ppwwyyxx: The current solution does look better. If `git status` is the concern it should be easy to just add `*.hip` to `.gitignore`. The current method creates a `hip` directory and will show up in `git status` anyway.

hgaspar: Ok sure, sounds good. 

I also implemented the change you suggested (hipify in the same directory) in pytorch, and I have a pull request in (in the pytorch ROCmSoftwarePlatform staging fork). Once this change is available in upstream pytorch, it can also be used in detectron2 (but we would need to test for pytorch minimum version to take advantage of same-directory hipification).  However we would like to have detectron2 work on earlier pytorch versions as well, so I would suggest that we start with the code as is today and once the pytorch change is merged and available I will improve the detectron2 setup.py to take advantage of it (if pytorch version allows the new feature). Does that sound ok?

ppwwyyxx: sounds good! please add a comment that links to the PR or issue in pytorch

setup.py
```diff
@@ -43,28 +44,50 @@ def get_extensions():
 
     main_source = path.join(extensions_dir, "vision.cpp")
     sources = glob.glob(path.join(extensions_dir, "**", "*.cpp"))
-    source_cuda = glob.glob(path.join(extensions_dir, "**", "*.cu")) + glob.glob(
-        path.join(extensions_dir, "*.cu")
-    )
+
+    is_rocm_pytorch = False
+    if torch.__version__ >= '1.5':
```
ppwwyyxx, line 49: not sure if it will happen, but '1.10' < '1.5' in terms of string. You can compare by tuple or list (like line 14)

ppwwyyxx:
Is it easy to test this in CircleCI? From reading the docs it seems that the HIP stack can be used on CUDA machines as well.

setup.py
hgaspar: The pytorch change PR currently is in:
https://github.com/ROCmSoftwarePlatform/pytorch/pull/656

It will undergo QA process before a PR to the upstream, but shouldn't take long.

setup.py
hgaspar: Agreed.

detectron2/layers/csrc/box_iou_rotated/box_iou_rotated_utils.h
hgaspar: Thank you, yes, already done!

setup.py
hgaspar: Update: I have addressed the torch version concern.  I will commit it, and also at the same time consolidate all my commits into one, to keep things self contained and then rebase onto the current tip, so it can be pulled easily.   Sounds ok?

Regarding circlci qst: Absolutely a good idea. However: And, yes, the hip kernel language can be executed on nvidia machines as well. And also we have some hip layers for some math libraries (e.g. blas) that also can be executed on both. However pytorch also uses "MIOpen" library, which is AMD's open source analogue of cudnn, and that will not run on nvidia.  You may be able to compile pytorch without using cudnn, and that may have a non zero chance of running on nvidia machines. But if it doesn't. this is not really a much supported configuration, and arguably not super interesting. 

On the other hand, pytorch circleci has AMD rocm target.  Can we do something similar for the detectron2 circleci?

sunway513:
@hgaspar can you rebase the PR against the tip of master?

detectron2/layers/csrc/cuda_version.cu
```diff
@@ -4,6 +4,10 @@
 
 namespace detectron2 {
 int get_cudart_version() {
+#if __HIP_PLATFORM_NVCC__
```
ppwwyyxx, line 7: When compiled with regular NVCC, does this macro exist?

setup.py
ppwwyyxx: Sounds good! 

PyTorch's rocm CI has a very complicated setup outside of circleci. It doesn't seem easy to achieve that. If testing the compilation on nvidia machines does not provide much value, I'm ok with skipping it.

sunway513: Hi @ppwwyyxx , we have also been working on migrating ROCm CI from Jenkins to CircleCI. Once that is in place we can work with you to add the ROCm CI in Detectron2 repo. Thank you for the help!

detectron2/layers/csrc/cuda_version.cu
hgaspar: Done. Also modified collect_env.py to print if the GPU platform is CUDA or HIP, and what version etc.

setup.py
hgaspar: I rebased my changes onto the current upstream master, as of now (5/26, 11 am EST)

ppwwyyxx:
You can see in history https://github.com/facebookresearch/detectron2/pull/1073/commits that many commits that belong to this PR are not on the top but in the middle, which causes this PR to explode. You can do `push -f` to your fork after fixing the history.

hgaspar:
As suggested, I cleaned up (=rewrote) the history, and also collapsed all the ROCM related commits in one, and cleanly rebased on current upstream tip.

ppwwyyxx:
thanks! could you fix the lint with `./dev/linter.sh`?

PR #1076, opened 2020-03-20
Set min size for proposal generator

### LINKED ISSUE #350: Rotated bounding box iou returns negative value
During training of rotated bounding box detection, matcher [at line 82](https://github.com/facebookresearch/detectron2/blob/f485b4684c5cf58b53263706be386f57e28d1841/detectron2/modeling/matcher.py#L82) sometimes fails.

Match quality is produced by rotated bounding box iou extension in files detectron2/layers/csrc/box_iou_rotated/. The output values of iou should always be in range [0, 1]

Specifically, the function single_box_iou_rotated in box_iou_rotated_utils.h isn't guaranteed to be in range [0, 1].

Workaround is to clamp values in match quality matrix in range [0, 1] so it passes the assertion

Perhaps add checks for valid input bounding box and assertion for valid output.

## Environment

------------------------  --------------------------------------------------
sys.platform              linux
Python                    3.7.5 (default, Oct 25 2019, 15:51:11) [GCC 7.3.0]
Numpy                     1.17.3
Detectron2 Compiler       GCC 7.4
Detectron2 CUDA Compiler  10.1
DETECTRON2_ENV_MODULE     <not set>
PyTorch                   1.3.1
PyTorch Debug Build       False
torchvision               0.4.2
CUDA available            True
GPU 0,1                   GeForce RTX 2080 Ti
CUDA_HOME                 /usr/local/cuda-10.1
NVCC                      Cuda compilation tools, release 10.1, V10.1.243
Pillow                    6.2.1
cv2                       4.1.1
------------------------  --------------------------------------------------
PyTorch built with:
  - GCC 7.3
  - Intel(R) Math Kernel Library Version 2019.0.4 Product Build 20190411 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v0.20.5 (Git Hash 0125f28c61c1f822fd48570b4c1066f96fcb9b2e)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - NNPACK is enabled
  - CUDA Runtime 10.1
  - NVCC architecture flags: -gencode;arch=compute_37,code=sm_37;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_60,code=sm_60;-gencode;arch=compute_61,code=sm_61;-gencode;arch=compute_70,code=sm_70;-gencode;arch=compute_75,code=sm_75;-gencode;arch=compute_37,code=compute_37
  - CuDNN 7.6.3
  - Magma 2.5.1
  - Build settings: BLAS=MKL, BUILD_NAMEDTENSOR=OFF, BUILD_TYPE=Release, CXX_FLAGS= -Wno-deprecated -fvisibility-inlines-hidden -fopenmp -DUSE_FBGEMM -DUSE_QNNPACK -DUSE_PYTORCH_QNNPACK -O2 -fPIC -Wno-narrowing -Wall -Wextra -Wno-missing-field-initializers -Wno-type-limits -Wno-array-bounds -Wno-unknown-pragmas -Wno-sign-compare -Wno-unused-parameter -Wno-unused-variable -Wno-unused-function -Wno-unused-result -Wno-strict-overflow -Wno-strict-aliasing -Wno-error=deprecated-declarations -Wno-stringop-overflow -Wno-error=pedantic -Wno-error=redundant-decls -Wno-error=old-style-cast -fdiagnostics-color=always -faligned-new -Wno-unused-but-set-variable -Wno-maybe-uninitialized -fno-math-errno -fno-trapping-math -Wno-stringop-overflow, DISABLE_NUMA=1, PERF_WITH_AVX=1, PERF_WITH_AVX2=1, PERF_WITH_AVX512=1, USE_CUDA=True, USE_EXCEPTION_PTR=1, USE_GFLAGS=OFF, USE_GLOG=OFF, USE_MKL=ON, USE_MKLDNN=ON, USE_MPI=OFF, USE_NCCL=ON, USE_NNPACK=ON, USE_OPENMP=ON, USE_STATIC_DISPATCH=OFF,

dnandha:
This change sets the minimum size for generated proposals from '0' to '1'. While this may seem like a small change it has a significant effect: It prevents box regression deltas from becoming negative. The way box regression 'width' (and 'height') are calculated is by doing `dw = ww * torch.log(target_widths / src_widths)` (see `box_regression.py`). The log becomes negative when 'target_widths / src_widths' < 1 which is always the case when 'target_widths' (generated proposal widths) is < 1.

To get back to the width of the predicted box in the code we calculate `pred_w = torch.exp(dw) * widths[:, None]`. Because exp of 'dw' now becomes really small we get almost '0' width. The area becomes almost '0', which in the cuda kernel leads to computation of negative IoU values.

While we could alternatively capture or clamp values in the code this is the cleanest way to solve this.

Fixes #350, it was closed without solution and people were still reporting the error

### DIFF
--- a/detectron2/config/defaults.py
+++ b/detectron2/config/defaults.py
@@ -156,7 +156,7 @@ _C.MODEL.PROPOSAL_GENERATOR = CN()
 _C.MODEL.PROPOSAL_GENERATOR.NAME = "RPN"
 # Proposal height and width both need to be greater than MIN_SIZE
 # (a the scale used during training or inference)
-_C.MODEL.PROPOSAL_GENERATOR.MIN_SIZE = 0
+_C.MODEL.PROPOSAL_GENERATOR.MIN_SIZE = 1
 
 
 # ---------------------------------------------------------------------------- #


ppwwyyxx:
> It prevents box regression deltas from becoming negative

The deltas are, by design, allowed to be negative. So there is no reason to prevent it.

> The area becomes almost '0', which in the cuda kernel leads to computation of negative IoU values.

If this is the reason that leads to negative IoU (an example would be good), the correct fix should be to improve numerical stability of the cuda kernel.

Issue #572, opened 2019-12-23
Visualization of rotated bounding boxes


kurapan:
Hi!
I'm working with rotated bounding boxes now and I found out that the visualizer cannot display the bounding box annotation for rotated boxes. The reason it cannot display rotated bounding boxes for a dataset by using `draw_dataset_dict` is simply because the code expects a `BoxMode` to convert the dataset coordinates to the variant which is used for visualization. However, rotated bounding boxes currently don't have any BoxMode. 
I made a small fix that let me skip the coordinate conversion and visualize the rotated bounding boxes. I was considering submitting a pull request, but I think that a cleaner solution would be to add an additional BoxMode value for rotated bounding boxes. What is your opinion?

This is my makeshift solution. I replaced line https://github.com/facebookresearch/detectron2/blob/706a3185d187ea912b324eb0b5eac064ca0a62ce/detectron2/utils/visualizer.py#L489 with this code.
```
if "bbox" in annos[0] and len(annos[0]["bbox"]) == 4:
    boxes = [BoxMode.convert(x["bbox"], x["bbox_mode"], BoxMode.XYXY_ABS) for x in annos]
else:
    boxes = [x["bbox"] for x in annos]
```

If you think this is okay, I will submit a pull request. If you think that an additional BoxMode value would be better, I can try adding one and submit a pull request then. 
I'm looking forward to your reply.

ppwwyyxx:
We have an internal commit that's going to add a box mode for rotated boxes soon.

kurapan:
I see. Thanks for the reply!

ppwwyyxx:
added in b83f35d30a016841041038caa4d04b4ae7feffc0

ppwwyyxx:
Though the rotated box mode is added, `draw_dataset_dict` will now convert rotated boxes to axis-aligned boxes and draw them like axis-align boxes. However it should be able to draw them using `overlay_rotated_instances`. This is still TBD.

ssaru:
Hi! now i trying visualize rotated bbox in detectron2.
but it's not visualize well

i build own get_dataset_dict as here
```
def get_rbox_detection_dicts():
    annotations = read_dataset()
    detection_anno = DetectionAnnotations(annotations)
    classes_list = collect_class_info(detection_anno)

    dataset_dicts = []
    for idx, FILE in enumerate(detection_anno.FILES):
        record = {}

        filename = FILE.FILEPATH
        height = FILE.IMAGE_HEIGHT
[...]

PR #1079, opened 2020-03-20
Support for loading / visualizing rotated box annotations / predictions

### LINKED ISSUE #21: Thanks for  great work, i have a question, how can i set config file to detect oriention target in faster rcnn or other network?

dnandha:
This PR provides the bare minimum changes to allow detectron2 users to load rotated ground truth boxes and display rotated box predictions. Further, an example config file for RRPN/FPN is provided, which definitely fixes #21 and #572.

Problem statement: Currently detectron2 doesn't allow for loading rotated `XYWHA` box ground truth instances. The loaded instances get transformed to 4 dimensional `XYWH` boxes in the annotation instance loader and visualization.

Code changes: The visualizer and `annotations_to_instances` function now handle rotated instance boxes. The `annotations_to_instances_rotated` function has been removed to eliminate redundant code. No further changes are required in the calling code (dataset mapper can be left unchanged).

The changes have been used together with (custom) dataloaders for the Jacquard and Cornell robotic grasping dataset which load rotated boxes annotating robotic grasps.

To validate the changes you can just take the example config provided together with a custom dataloader which loads box annotations in the `XYWHA` format.

I'm aware of #996: while we have similar goals, my PR has fewer changes limited only on fixing the box conversion handling in the loader and visualizer.

### DIFF
--- a/configs/Base-RRCNN-FPN.yaml
+++ b/configs/Base-RRCNN-FPN.yaml
@@ -1,0 +1,54 @@
+MODEL:
+  META_ARCHITECTURE: "GeneralizedRCNN"
+  BACKBONE:
+    NAME: "build_resnet_fpn_backbone"
+  RESNETS:
+    OUT_FEATURES: ["res2", "res3", "res4", "res5"]
+  FPN:
+    IN_FEATURES: ["res2", "res3", "res4", "res5"]
+  ANCHOR_GENERATOR:
+    NAME: "RotatedAnchorGenerator"
+    SIZES: [[32], [64], [128], [256], [512]]  # One size for each in feature map
+    ASPECT_RATIOS: [[0.5, 1.0, 2.0]]
+    ANGLES: [[0, 30, 60]]
+  PROPOSAL_GENERATOR:
+    NAME: "RRPN"
+    MIN_SIZE: 1 # VERY IMPORTANT! otherwise various crashes
+  RPN:
+    HEAD_NAME: "StandardRPNHead"
+    BBOX_REG_WEIGHTS: [1, 1, 1, 1, 1]
+    IN_FEATURES: ["p2", "p3", "p4", "p5", "p6"]
+    PRE_NMS_TOPK_TRAIN: 2000  # Per FPN level
+    PRE_NMS_TOPK_TEST: 1000  # Per FPN level
+    # Detectron1 uses 2000 proposals per-batch,
+    # (See "modeling/rpn/rpn_outputs.py" for details of this legacy issue)
+    # which is approximately 1000 proposals per-image since the default batch size for FPN is 2.
+    POST_NMS_TOPK_TRAIN: 1000
+    POST_NMS_TOPK_TEST: 1000
+    IOU_THRESHOLDS: [0.3, 0.7]
+  ROI_HEADS:
+    NAME: "RROIHeads"
+    IN_FEATURES: ["p2", "p3", "p4", "p5"]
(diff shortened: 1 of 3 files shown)


configs/Base-RRCNN-FPN.yaml
```diff
@@ -0,0 +1,54 @@
...
+  FPN:
+    IN_FEATURES: ["res2", "res3", "res4", "res5"]
+  ANCHOR_GENERATOR:
+    NAME: "RotatedAnchorGenerator"
+    SIZES: [[32], [64], [128], [256], [512]]  # One size for each in feature map
+    ASPECT_RATIOS: [[0.5, 1.0, 2.0]]
+    ANGLES: [[0, 30, 60]]
+  PROPOSAL_GENERATOR:
+    NAME: "RRPN"
+    MIN_SIZE: 1 # VERY IMPORTANT! otherwise various crashes
```
ppwwyyxx: as said in another PR this is incorrect

configs/Base-RRCNN-FPN.yaml
```diff
@@ -0,0 +1,54 @@
+MODEL:
```
ppwwyyxx: Please inherit from `Base-RCNN-FPN.yaml` as much as possible.
And add a comment here that this is just an example of what may need to be changed, and it is NOT a config meant for training directly (since we don't have code for a rotated dataset).

ppwwyyxx:
Closing as there is no more activity and review feedbacks have not been addressed.

detectron2/data/detection_utils.py
```diff
@@ -296,35 +305,6 @@ def annotations_to_instances(annos, image_size, mask_format="polygon"):
     return target
 
 
-def annotations_to_instances_rotated(annos, image_size):
```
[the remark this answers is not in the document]
dnandha: No. Whatever you expect from your users is complicated and unnecessary and the reason why so many people find working with detectron a pain.
1. `transform_instance_annotations` before my change transforms all Boxes, no matter rotated or not, to XYWH Boxes. That's a bug, in a place where you argue you don't want users to make changes.
2. When `transform_instance_annotations` is allowed to perform Box mode conversions then there's no need to write redundant and duplicate code with a separate `_rotated` function. I've never heard that an `if` complicates the code.
Btw. 'detection_utils' is a terrible name for a file.

configs/Base-RRCNN-FPN.yaml
dnandha: that's your opinion

configs/Base-RRCNN-FPN.yaml
dnandha: I agree with the first part.
I don't think a comment is necessary. We never use the `Base-` configs to train directly.

detectron2/utils/visualizer.py
```diff
@@ -487,7 +487,12 @@ def draw_dataset_dict(self, dic):
             else:
                 keypts = None
 
-            boxes = [BoxMode.convert(x["bbox"], x["bbox_mode"], BoxMode.XYXY_ABS) for x in annos]
+            if annos[0]["bbox_mode"] == BoxMode.XYWHA_ABS:
```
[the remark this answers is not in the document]
dnandha: can do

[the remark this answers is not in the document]
dnandha: can do

Issue #1077, opened 2020-03-20
Assigned colors for bounding boxes drawn by Visualizer


sheromon:
## 🚀 Feature
The Visualizer class has convenient high-level methods `draw_dataset_dict` and `draw_instance_predictions`, and the lower-level method `draw_box` has an edge color argument that is not used by these high-level methods. I would like to be able to use per-class color-coding when drawing truth and prediction bounding boxes similar to that already used for segmentation predictions.

It would  also be nice to expose an optional line style input arg to one or both of these methods, but the color-coding is more important to me.

## Motivation & Examples

Color-coding is a simple way to convey information in the ground truth and prediction bounding box plots that is currently not being used by the methods referenced above. I expect it will be convenient for many users to be able to associate each class with a bounding box of a specific color. I've implemented it in my own project, and it makes my visualizations much easier to digest compared to having random colors for each bounding box.

It is potentially straightforward to implement and consistent with the existing use of color-coding for segmentation by taking advantage of `thing_colors` in the metadata, just extending the use from segmentation to bounding boxes as well.

ppwwyyxx:
Bounding box colors already match the segmentation, as shown in https://colab.research.google.com/drive/16jcaJoc6bCFAQ96jDe2HwtXj7BMD_-m5

sheromon:
Thanks for pointing that out! I will take a look at the example and see if I can do the same thing to fill my need.

sheromon:
Oh, I see what's going on. The boxes get the same colors as the segmentation if you're visualizing both segmentation and bounding box results at the same time. If you're visualizing only bounding boxes by themselves, the colors are random.

pvtien96:
@sheromon @ppwwyyxx by the way, is there any function available in detectron2 that allow draw a mask with specific color on the original image? 
Assume that:
```
outputs = predictor(img)
masks = outputs["instances"].pred_masks.cpu().numpy()
for mask in masks:
    #where mask is np.darry of True and False. I want each instance has a color that I define myself.
    color, alpha = #what I define my self, ex: (100, 20, 20), 0.5
    draw_alpha_mask(img, binary_mask, color, alpha)
```
do you know how to write draw_alpha_mask?
Many thanks!

ppwwyyxx:
There are functions in https://detectron2.readthedocs.io/modules/utils.html#detectron2.utils.visualizer.Visualizer

ppwwyyxx:
[...]

PR #1091, opened 2020-03-24
If metadata has asigned colors, use them in draw_dataset_dict.

sheromon:
Trying to address issue #1077.

### DIFF
--- a/detectron2/utils/visualizer.py
+++ b/detectron2/utils/visualizer.py
@@ -490,6 +490,9 @@ class Visualizer:
             boxes = [BoxMode.convert(x["bbox"], x["bbox_mode"], BoxMode.XYXY_ABS) for x in annos]
 
             labels = [x["category_id"] for x in annos]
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
Thanks! Could you follow the logic in `draw_instance_predictions` Line 347-354 (perhaps make these lines a new method) so the two methods are consistent in their drawing style?

sheromon:
Right, so I updated my change to be more similar to what's in draw_instance_predictions. I thought about making a method, but it looked like it could get messy to incorporate the full ColorMode handling in draw_instance_predictions.

I can see the rationale for including the color jitter -- if you have overlapping instances of the same class colored by class, the boundaries may be less clear or unclear. But I'd like to turn it down or off. Any thoughts about that? I could see there being a scalar between 0 and 1 controlling how much noise is added in the jitter step, but on the other hand, it's additional complexity.

sheromon:
Also, thank you for getting back to me quickly!

PR #1166, opened 2020-04-07
Support empty predictions when visualizing results

c1aris:
This PR fixes a potential error when no predictions found in `tools/visualize_json_results.py`.

### DIFF
--- a/tools/visualize_json_results.py
+++ b/tools/visualize_json_results.py
@@ -22,6 +22,9 @@ def create_instances(predictions, image_size):
     score = np.asarray([x["score"] for x in predictions])
     chosen = (score > args.conf_threshold).nonzero()[0]
     score = score[chosen]
+    if score.shape[0] == 0:
+        return None
+
     bbox = np.asarray([predictions[i]["bbox"] for i in chosen])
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
What is the error without this PR?

c1aris:
The `BoxMode.convert` method would raise an IndexError when the shape of `bbox` becomes `(0,)`, which may be caused by empty prediction or all the confidences are below `conf_threshold`.

Besides, maybe checking the shape of `chosen` instead of `score` is more reasonable, I'll create one more commit to fix that.

Sorry but I can not reproduce the bug because I've deleted the json file by mistake.

ppwwyyxx:
Does it work if you add `bbox = bbox.reshape(-1, 4)` to make it a ndarray of shape (0, 4) ?
It's reasonable to expect `BoxMode.convert` to produce a (0, 4) array when the input shape is (0, 4). If that's not working we can get it fixed.

c1aris:
Yes it works after reshaping `bbox`. I've tested it by using another json file and setting `conf_threshold` to 2.0. Thanks for your suggestion.

PR #1170, opened 2020-04-07
rename in_channels variable used in FPN constructor

nanddalal:
This PR renames the `in_channels` variable used in constructing the FPN.

I found it a bit confusing that we were both enumerating over `in_channels` and assigning each element to a variable also called `in_channels` so at the end of the for loop, `in_channels` would actually be the last element of the original `in_channels` list.

### DIFF
--- a/detectron2/modeling/backbone/fpn.py
+++ b/detectron2/modeling/backbone/fpn.py
@@ -50,24 +50,20 @@ class FPN(Backbone):
 
         # Feature map strides and channels from the bottom up network (e.g. ResNet)
         input_shapes = bottom_up.output_shape()
-        in_strides = [input_shapes[f].stride for f in in_features]
-        in_channels = [input_shapes[f].channels for f in in_features]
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
I agree it's a bit confusing. Would you maybe rename it to `in_channels_per_feature`? I'm also confused in what `in_channels_idx` means

rbgirshick:
I agree that `in_channels_per_feature` is a better name.

A slightly larger change that I think might be nicer is:

```python
strides = [input_shapes[f].stride for f in in_features]
channels = [input_shapes[f].channels for f in in_features]
...
for stride, in_channels in zip(all_strides, all_channels):
  ...
```

ppwwyyxx:
@nanddalal  could you sign the CLA?

nanddalal:
hi @ppwwyyxx -- yes I am working on it with my company. Thank you both for the PR review!

PR #1198, opened 2020-04-10
filter_empty_instances with box threshold

alekseynp:
Minor change to enable passing a threshold through filter_empty_instances.
Due to floating point issues I had boxes with width 1e-14 that passed the 0 threshold but blew up later.
Also fixed the type hinting to float. These are floats.

### DIFF
--- a/detectron2/data/detection_utils.py
+++ b/detectron2/data/detection_utils.py
@@ -325,7 +325,7 @@ def annotations_to_instances_rotated(annos, image_size):
     return target
 
 
-def filter_empty_instances(instances, by_box=True, by_mask=True):
+def filter_empty_instances(instances, by_box=True, by_mask=True, box_threshold=1e-5):
     """
     Filter out empty instances in an `Instances` object.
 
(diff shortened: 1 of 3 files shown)


detectron2/data/detection_utils.py
```diff
@@ -325,22 +325,23 @@ def annotations_to_instances_rotated(annos, image_size):
     return target
 
 
-def filter_empty_instances(instances, by_box=True, by_mask=True):
+def filter_empty_instances(instances, by_box=True, by_mask=True, box_threshold=0):
```
ppwwyyxx, line 328: Maybe it makes sense to use a slightly larger default like `1e-5` to avoid the issue you met.

alekseynp: I agree. Would you want that default in Boxes.nonempty and RotatedBoxes.nonempty as well?

ppwwyyxx: The results of `Boxes.nonempty` may be used for many purposes, so it might be more natural to use 0 as default there.

However here the purpose is specifically about filtering bad boxes - and it's safe to say that boxes<1e-5 are by default "bad"

alekseynp: Done

PR #1208, opened 2020-04-14
Make padding shape tracable.

xkszltl:
Previously `max_size` will be hardcoded in traced ONNX model since some dataflow/computation is outside PyTorch.

The entire chunk between `[Sub, Div]`(whitening) and `Pad` shows how this shape inferencing works, with `size_divisibility==32` for this backbone.
`ceil()` is replaced with `[Add, Div, Mul]` in the middle, i.e. `(n + (m - 1)) // m * m` with constant folding for `m-1`.
image

### DIFF
--- a/detectron2/structures/image_list.py
+++ b/detectron2/structures/image_list.py
@@ -70,16 +70,21 @@ class ImageList(object):
             assert isinstance(t, torch.Tensor), type(t)
             assert t.shape[1:-2] == tensors[0].shape[1:-2], t.shape
         # per dimension maximum (H, W) or (C_1, ..., C_K, H, W) where K >= 1 among all tensors
-        max_size = tuple(max(s) for s in zip(*[img.shape for img in tensors]))
+        max_size = (
+            torch.stack(
+                [
+                    # Keep original tensor for tracing.
+                    torch.stack(size) if torch.is_tensor(size[0]) else torch.tensor(size)
+                    for size in [tuple(img.shape) for img in tensors]
+                ]
+            )
+            .max(0)
+            .values
+        )
 
         if size_divisibility > 0:
-            import math
-
             stride = size_divisibility
-            max_size = list(max_size)  # type: ignore
-            max_size[-2] = int(math.ceil(max_size[-2] / stride) * stride)  # type: ignore
-            max_size[-1] = int(math.ceil(max_size[-1] / stride) * stride)  # type: ignore
-            max_size = tuple(max_size)
+            max_size = torch.cat([max_size[:-2], (max_size[-2:] + (stride - 1)) // stride * stride])
 
         image_sizes = [tuple(im.shape[-2:]) for im in tensors]
 
@@ -94,8 +99,10 @@ class ImageList(object):
                 padded = F.pad(tensors[0], padding_size, value=pad_value)
                 batched_imgs = padded.unsqueeze_(0)
         else:
-            batch_shape = (len(tensors),) + max_size
-            batched_imgs = tensors[0].new_full(batch_shape, pad_value)
+            # Keep original tensor for tracing.
+            num_imgs = len(tensors) if torch.is_tensor(len(tensors)) else torch.tensor(len(tensors))
+            batch_shape = torch.cat([num_imgs[None], max_size])
+            batched_imgs = tensors[0].new_full(tuple(batch_shape), pad_value)
             for img, pad_img in zip(tensors, batched_imgs):
                 pad_img[..., : img.shape[-2], : img.shape[-1]].copy_(img)
 
--- a/detectron2/structures/image_list.py
+++ b/detectron2/structures/image_list.py
@@ -74,7 +74,7 @@ class ImageList(object):
             torch.stack(
                 [
                     # Keep original tensor for tracing.
-                    torch.stack(size) if torch.is_tensor(size[0]) else torch.tensor(size)
+                    torch.stack([torch.as_tensor(dim) for dim in size])
                     for size in [tuple(img.shape) for img in tensors]
                 ]
             )
@@ -100,8 +100,7 @@ class ImageList(object):
                 batched_imgs = padded.unsqueeze_(0)
         else:
             # Keep original tensor for tracing.
-            num_imgs = len(tensors) if torch.is_tensor(len(tensors)) else torch.tensor(len(tensors))
-            batch_shape = torch.cat([num_imgs[None], max_size])
+            batch_shape = torch.cat([torch.as_tensor(len(tensors))[None], max_size])
             batched_imgs = tensors[0].new_full(tuple(batch_shape), pad_value)
             for img, pad_img in zip(tensors, batched_imgs):
                 pad_img[..., : img.shape[-2], : img.shape[-1]].copy_(img)
--- a/tests/test_data_loader.py
+++ b/tests/test_data_loader.py
@@ -7,7 +7,9 @@ import pycocotools.mask as mask_util
 
 from detectron2.data import detection_utils
 from detectron2.data import transforms as T
-from detectron2.structures import BitMasks, BoxMode
+from detectron2.structures import BitMasks, BoxMode, ImageList
+
+import torch
 
 
 class TestTransformAnnotations(unittest.TestCase):
@@ -114,3 +116,10 @@ class TestTransformAnnotations(unittest.TestCase):
         instance = {"bbox": [10, 10, 100, 100], "bbox_mode": BoxMode.XYXY_ABS}
         with self.assertRaises(AssertionError):
             detection_utils.gen_crop_transform_with_instance((10, 10), (15, 15), instance)
+
+    def test_imagelist_padding_shape(self):
+        class TensorToImageList(torch.nn.Module):
+            def forward(self, tensor: torch.Tensor):
+                return ImageList.from_tensors([torch.ones((3, 10, 10))], 4).tensor
+        func = torch.jit.trace(TensorToImageList(), torch.ones((3, 10, 10), dtype=torch.float32))
+        func(torch.ones((3, 15, 20), dtype=torch.float32))
--- a/tests/test_data_loader.py
+++ b/tests/test_data_loader.py
@@ -117,18 +117,21 @@ class TestTransformAnnotations(unittest.TestCase):
         with self.assertRaises(AssertionError):
             detection_utils.gen_crop_transform_with_instance((10, 10), (15, 15), instance)
 
-    def test_imagelist_padding_shape(self):
-        class TensorToImageList(torch.nn.Module):
-            def forward(self, tensors: Sequence[torch.Tensor]):
-                return ImageList.from_tensors(tensors, 4).tensor
 
+class TestInputShapeInferencing(unittest.TestCase):
+    class TensorsToImageList(torch.nn.Module):
+        def forward(self, tensors: Sequence[torch.Tensor]):
+            return ImageList.from_tensors(tensors, 4).tensor
+
+    def test_imagelist_single_padding_shape(self):
         func = torch.jit.trace(
-            TensorToImageList(), ([torch.ones((3, 10, 10), dtype=torch.float32)],)
+            self.TensorsToImageList(), ([torch.ones((3, 10, 10), dtype=torch.float32)],)
         )
         func([torch.ones((3, 15, 20), dtype=torch.float32)])
 
+    def test_imagelist_multi_padding_shape(self):
         func = torch.jit.trace(
-            TensorToImageList(),
+            self.TensorsToImageList(),
             (
                 [
                     torch.ones((3, 16, 10), dtype=torch.float32),
--- a/tests/test_data_loader.py
+++ b/tests/test_data_loader.py
@@ -3,6 +3,7 @@
 import copy
 import numpy as np
 import unittest
+from typing import Sequence
 import pycocotools.mask as mask_util
 import torch
 
@@ -118,8 +119,26 @@ class TestTransformAnnotations(unittest.TestCase):
 
     def test_imagelist_padding_shape(self):
         class TensorToImageList(torch.nn.Module):
-            def forward(self, tensor: torch.Tensor):
-                return ImageList.from_tensors([torch.ones((3, 10, 10))], 4).tensor
+            def forward(self, tensors: Sequence[torch.Tensor]):
+                return ImageList.from_tensors(tensors, 4).tensor
 
-        func = torch.jit.trace(TensorToImageList(), torch.ones((3, 10, 10), dtype=torch.float32))
-        func(torch.ones((3, 15, 20), dtype=torch.float32))
+        func = torch.jit.trace(
+            TensorToImageList(), ([torch.ones((3, 10, 10), dtype=torch.float32)],)
+        )
+        func([torch.ones((3, 15, 20), dtype=torch.float32)])
+
+        func = torch.jit.trace(
+            TensorToImageList(),
+            (
+                [
+                    torch.ones((3, 16, 10), dtype=torch.float32),
+                    torch.ones((3, 13, 11), dtype=torch.float32),
+                ],
+            ),
+        )
+        func(
+            [
+                torch.ones((3, 25, 20), dtype=torch.float32),
+                torch.ones((3, 10, 10), dtype=torch.float32),
+            ]
+        )
--- a/tests/test_data_loader.py
+++ b/tests/test_data_loader.py
@@ -4,12 +4,11 @@ import copy
 import numpy as np
 import unittest
 import pycocotools.mask as mask_util
+import torch
 
 from detectron2.data import detection_utils
 from detectron2.data import transforms as T
 from detectron2.structures import BitMasks, BoxMode, ImageList
-
-import torch
 
 
 class TestTransformAnnotations(unittest.TestCase):
@@ -121,5 +120,6 @@ class TestTransformAnnotations(unittest.TestCase):
         class TensorToImageList(torch.nn.Module):
             def forward(self, tensor: torch.Tensor):
                 return ImageList.from_tensors([torch.ones((3, 10, 10))], 4).tensor
+
         func = torch.jit.trace(TensorToImageList(), torch.ones((3, 10, 10), dtype=torch.float32))
         func(torch.ones((3, 15, 20), dtype=torch.float32))


xkszltl:
retest this please

ppwwyyxx:
We typically avoid complicating our code to support onnx.
For this particular case, when we convert models to onnx and then caffe2 we do not trace `ImageList.from_tensors`. Instead the converted model starts from a NCHW tensor directly. This is already implemented in https://detectron2.readthedocs.io/tutorials/deployment.html

xkszltl:
@ppwwyyxx 
Thanks for reply.
I'm aware of that.
In our case we're using onnxruntime instead of caffe2 as backend, so we need to export standard ONNX model instead.

I agree we should not over-complicate code for ONNX, and followed this idea when crafting this PR.
IMHO model-specific parameters, e.g. whitening avg/var and backbone-specific padding for size_divisibility should be keep together with model to make it work "out-of-box".

The challenge is mainly to bridge `torch.Tensor` and `torch.Size`, which becomes tensor instead of int when taking subset/elements during export.
Previously the code is already doing a good job by using `tuple(torch.Size)` to get tracable shape as `torch.Size=>(torch.tensor(C),torch.tensor(H),torch.tensor(W))`, I'm just making sure the we doesn't become `(int(C), int(H), int(W))`.

xkszltl:
Add some background:

We build our own model upon Detectron2 and use detectron2 purely as a library without and code change.
To export ONNX model for us, there're 2 gaps to fix:
    1. `class _ROIAlign(autograd.Function)` is not tracable. We fixed this by `_ROIAlign.symbolic=...` to inject custom rule right before ONNX export.
    2. Keep "must-have" data preprocessing exported automatically. This is what this PR trying to do.

We'd like to upstream our patches so it's sharable with others and avoid dirty hacks around.
Let me know if there's good suggestions or code changes needed to get this in.
Thanks!

xkszltl:
Regarding code complexity, I replaced custom math computation with PyTorch native ops.
Line-of-code is actually decreased before running linter.

xkszltl:
Dark side shows the traced ONNX model before this PR.
As you can see 1024x512 are hard-coded by the tracing image.
This PR add the **red** edge.

Screen Shot 2020-04-14 at 22 57 56

ppwwyyxx:
Thanks for sharing more context.
It would be awesome to support onnxruntime as well. There are two ways to support conversion towards a new runtime:
1. write a wrapper that modifies behavior of an existing model so it successfully converts to the new runtime
2. modify existing model so it converts to the new runtime

The problems with (2) are mainly the following:
* more complicated code. Even if it's simple now it can become complicated when more models need to be supported
* hard to track: the magic hacks of each runtime is scattered around and, even if one day some of them are no longer necessary due to improvement in tracing or in this runtime, it becomes very hard to undo those hacks
* different runtime may require different or potentially conflicting modifications
* it has to be implemented inside detectron2 while (1) can be maintained separately by community. Though we welcome PRs as well as long as they are not too complicated and have instructions to use.

This is why we use (1) for caffe2 and I think we probably will use (2) only for torchscript because it's more "official".

xkszltl:
I agree with your point, but I wonder if you misunderstood the change.

The code added here has nothing to do with new runtime.
Instead, it simply make it tracable by pytorch.
There's no runtime-specific magic here.

Actually if you try to export this preprocesing part to Caffe2, you'll run into the same issue.

xkszltl:
In other words, this is not a "new feature", but a "bugfix".

xkszltl:
If the magic you're talking about is `torch.Size` vs. `torch.Tensor`, that's for pytorch tracer, not onnx.

ppwwyyxx:
OK. I actually consider traceable (with generic shapes) as a feature that we currently don't support in the models. 
But I saw your point. If the change is not targeted specifically at onnxruntime, I think it's reasonable to add support.
Some follow up:
* Are there any other shapes that have the same issues or is this probably the only one? If I didn't misunderstand your comment above this seems to be the only one. 
* A test is important in particular because the original code has problems only in a situation that's currently unused. Is there a way (hopefully not complicated) to test the behavior?

xkszltl:
Interesting...Most of the shape inferencing was already there so I thought it's part of the design.

Answer to your questions:
- This is the only one I ran into.
- Is there any test coverage currently for export? Maybe we can trace data_loader with a small image and then inference with a large one. I expect old code to crash due to negative shape.

ppwwyyxx:
We have internal tests for the export, basically they just run what `caffe2_converter` does. Haven't got a chance to open source them since they require downloading models so can be a bit complicated

xkszltl:
Don't worry.
I've created a simple test which crashed on master and passed with this PR.
Will push it soon.

xkszltl:
**Before**
```
# python3 -m unittest -v tests.test_data_loader.TestTransformAnnotations.test_imagelist_padding_shape
test_imagelist_padding_shape (tests.test_data_loader.TestTransformAnnotations) ... ../aten/src/ATen/native/BinaryOps.cpp:81: UserWarning: Integer division of tensors using div or / is deprecated, and in a future release div will perform true division as in Python 3. Use true_divide or floor_divide (// in Python) instead.
/mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:80: TracerWarning: Converting a tensor to a Python float might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  max_size[-2] = int(math.ceil(max_size[-2] / stride) * stride)  # type: ignore
/mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:81: TracerWarning: Converting a tensor to a Python float might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  max_size[-1] = int(math.ceil(max_size[-1] / stride) * stride)  # type: ignore
/mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:91: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if all(x == 0 for x in padding_size):  # https://github.com/pytorch/pytorch/issues/31734
ERROR

======================================================================
ERROR: test_imagelist_padding_shape (tests.test_data_loader.TestTransformAnnotations)
----------------------------------------------------------------------
  File "/mnt/Projects/PyTorchDetector/detectron2/tests/test_data_loader.py", line 124, in test_imagelist_padding_shape
    func = torch.jit.trace(TensorToImageList(), torch.ones((3, 10, 10), dtype=torch.float32))
  File "/usr/local/lib/python3.6/dist-packages/torch/jit/__init__.py", line 874, in trace
    check_tolerance, _force_outplace, _module_class)
  File "/usr/local/lib/python3.6/dist-packages/torch/jit/__init__.py", line 1036, in trace_module
    check_tolerance, _force_outplace, True, _module_class)
  File "/usr/local/lib/python3.6/dist-packages/torch/autograd/grad_mode.py", line 15, in decorate_context
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/torch/jit/__init__.py", line 668, in _check_trace
    if compare_outputs(traced_outs, fn_outs, 'Python function'):
  File "/usr/local/lib/python3.6/dist-packages/torch/jit/__init__.py", line 656, in compare_outputs
    atol=torch.testing._get_default_tolerance(orig, ref)[1])
  File "/usr/local/lib/python3.6/dist-packages/torch/testing/__init__.py", line 24, in assert_allclose
    expected = expected.expand_as(actual)
RuntimeError: The expanded size of the tensor (8) must match the existing size (12) at non-singleton dimension 3.  Target sizes: [1, 3, 8, 8].  Tensor sizes: [1, 3, 12, 12]

----------------------------------------------------------------------
Ran 1 test in 0.020s

FAILED (errors=1)
```

**After**
```
# python3 -m unittest -v tests.test_data_loader.TestTransformAnnotations.test_imagelist_padding_shape
test_imagelist_padding_shape (tests.test_data_loader.TestTransformAnnotations) ... /mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:96: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if all(x == 0 for x in padding_size):  # https://github.com/pytorch/pytorch/issues/31734
ok

----------------------------------------------------------------------
Ran 1 test in 0.032s

OK
```

xkszltl:
Actually as you can see they're also covered by pytorch's warning.

detectron2/structures/image_list.py
```diff
@@ -70,16 +70,21 @@ def from_tensors(
...
+            torch.stack(
+                [
+                    # Keep original tensor for tracing.
+                    torch.stack(size) if torch.is_tensor(size[0]) else torch.tensor(size)
+                    for size in [tuple(img.shape) for img in tensors]
+                ]
+            )
+            .max(0)
+            .values
+        )
```
ppwwyyxx, line 83: how about 
```
max_size = torch.as_tensor([img.shape for img in tensors]).max(0).values
```
?

tests/test_data_loader.py
```diff
@@ -114,3 +115,11 @@ def test_gen_crop_outside_boxes(self):
...
         with self.assertRaises(AssertionError):
             detection_utils.gen_crop_transform_with_instance((10, 10), (15, 15), instance)
+
+    def test_imagelist_padding_shape(self):
+        class TensorToImageList(torch.nn.Module):
+            def forward(self, tensor: torch.Tensor):
+                return ImageList.from_tensors([torch.ones((3, 10, 10))], 4).tensor
+
+        func = torch.jit.trace(TensorToImageList(), torch.ones((3, 10, 10), dtype=torch.float32))
+        func(torch.ones((3, 15, 20), dtype=torch.float32))
```
ppwwyyxx, line 125: could you also test the changes on the `batch_shape` branch?

detectron2/structures/image_list.py
xkszltl: This will failed my unittest in 2 ways:
1. PyTorch:
- Normally:
  - `.shape` => `torch.Size`
  - `.shape[I]` => `int`
- During tracing:
  - `.shape` => `torch.Size`
  - `.shape[i]` => `torch.Tensor`
2. `as_tensor(List)` recreates a tensor with constants while `stack(List)` add a node in computation graph.

But your suggestion is very helpful.
I can replace the if-else with `torch.stack([torch.as_tensor(dim) for dim in size])`.

tests/test_data_loader.py
xkszltl: Sure

xkszltl:
Added more unit tests.

**Before**
```
# python3 -m unittest -v tests.test_data_loader.TestInputShapeInferencing.test_imagelist_{single,multi}_padding_shape                                                                               
test_imagelist_single_padding_shape (tests.test_data_loader.TestInputShapeInferencing) ... ../aten/src/ATen/native/BinaryOps.cpp:81: UserWarning: Integer division of tensors using div or / is deprecated, and in a future release div will perform true divi
sion as in Python 3. Use true_divide or floor_divide (// in Python) instead.
/mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:80: TracerWarning: Converting a tensor to a Python float might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as 
a constant in the future. This means that the trace might not generalize to other inputs!
  max_size[-2] = int(math.ceil(max_size[-2] / stride) * stride)  # type: ignore
/mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:81: TracerWarning: Converting a tensor to a Python float might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as 
a constant in the future. This means that the trace might not generalize to other inputs!
  max_size[-1] = int(math.ceil(max_size[-1] / stride) * stride)  # type: ignore
/mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:91: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated a
s a constant in the future. This means that the trace might not generalize to other inputs!
  if all(x == 0 for x in padding_size):  # https://github.com/pytorch/pytorch/issues/31734
ERROR
test_imagelist_multi_padding_shape (tests.test_data_loader.TestInputShapeInferencing) ... /mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:73: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be 
incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  max_size = tuple(max(s) for s in zip(*[img.shape for img in tensors]))
/usr/local/lib/python3.6/dist-packages/torch/tensor.py:467: RuntimeWarning: Iterating over a tensor might cause the trace to be incorrect. Passing a tensor of different shape won't change the number of iterations executed (and might lead to errors or sil
ently give incorrect results).
  'incorrect results).', category=RuntimeWarning)
/usr/local/lib/python3.6/dist-packages/torch/tensor.py:468: TracerWarning: Converting a tensor to a Python index might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the fut
ure. This means that the trace might not generalize to other inputs!
  return iter(imap(lambda i: self[i], range(self.size(0))))
/mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:100: TracerWarning: Converting a tensor to a Python index might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated a$
 a constant in the future. This means that the trace might not generalize to other inputs!
  pad_img[..., : img.shape[-2], : img.shape[-1]].copy_(img)
ERROR

======================================================================
ERROR: test_imagelist_single_padding_shape (tests.test_data_loader.TestInputShapeInferencing)
----------------------------------------------------------------------
  File "/mnt/Projects/PyTorchDetector/detectron2/tests/test_data_loader.py", line 127, in test_imagelist_single_padding_shape
    self.TensorsToImageList(), ([torch.ones((3, 10, 10), dtype=torch.float32)],)
  File "/usr/local/lib/python3.6/dist-packages/torch/jit/__init__.py", line 874, in trace
    check_tolerance, _force_outplace, _module_class)
  File "/usr/local/lib/python3.6/dist-packages/torch/jit/__init__.py", line 1036, in trace_module
    check_tolerance, _force_outplace, True, _module_class)
  File "/usr/local/lib/python3.6/dist-packages/torch/autograd/grad_mode.py", line 15, in decorate_context
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/torch/jit/__init__.py", line 668, in _check_trace
    if compare_outputs(traced_outs, fn_outs, 'Python function'):
  File "/usr/local/lib/python3.6/dist-packages/torch/jit/__init__.py", line 656, in compare_outputs
    atol=torch.testing._get_default_tolerance(orig, ref)[1])
  File "/usr/local/lib/python3.6/dist-packages/torch/testing/__init__.py", line 24, in assert_allclose
    expected = expected.expand_as(actual)
RuntimeError: The expanded size of the tensor (8) must match the existing size (12) at non-singleton dimension 3.  Target sizes: [1, 3, 8, 8].  Tensor sizes: [1, 3, 12, 12]

======================================================================
ERROR: test_imagelist_multi_padding_shape (tests.test_data_loader.TestInputShapeInferencing)
----------------------------------------------------------------------
  File "/mnt/Projects/PyTorchDetector/detectron2/tests/test_data_loader.py", line 137, in test_imagelist_multi_padding_shape
    torch.ones((3, 13, 11), dtype=torch.float32),
  File "/usr/local/lib/python3.6/dist-packages/torch/jit/__init__.py", line 874, in trace
    check_tolerance, _force_outplace, _module_class)
  File "/usr/local/lib/python3.6/dist-packages/torch/jit/__init__.py", line 1026, in trace_module
    module._c._create_method_from_trace(method_name, func, example_inputs, var_lookup_fn, _force_outplace)
  File "/usr/local/lib/python3.6/dist-packages/torch/nn/modules/module.py", line 556, in __call__
    result = self._slow_forward(*input, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/torch/nn/modules/module.py", line 542, in _slow_forward
    result = self.forward(*input, **kwargs)
  File "/mnt/Projects/PyTorchDetector/detectron2/tests/test_data_loader.py", line 123, in forward
    return ImageList.from_tensors(tensors, 4).tensor
  File "/mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py", line 100, in from_tensors
    pad_img[..., : img.shape[-2], : img.shape[-1]].copy_(img)
RuntimeError: The size of tensor a (8) must match the size of tensor b (10) at non-singleton dimension 2

----------------------------------------------------------------------
Ran 2 tests in 0.027s

FAILED (errors=2)
```

**After**
```
# python3 -m unittest -v tests.test_data_loader.TestInputShapeInferencing.test_imagelist_{single,multi}_padding_shape
test_imagelist_single_padding_shape (tests.test_data_loader.TestInputShapeInferencing) ... /mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:77: TracerWarning: torch.as_tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  torch.stack([torch.as_tensor(dim) for dim in size])
/mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:96: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if all(x == 0 for x in padding_size):  # https://github.com/pytorch/pytorch/issues/31734
ok
test_imagelist_multi_padding_shape (tests.test_data_loader.TestInputShapeInferencing) ... /mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:103: TracerWarning: torch.as_tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  batch_shape = torch.cat([torch.as_tensor(len(tensors))[None], max_size])
/usr/local/lib/python3.6/dist-packages/torch/tensor.py:467: RuntimeWarning: Iterating over a tensor might cause the trace to be incorrect. Passing a tensor of different shape won't change the number of iterations executed (and might lead to errors or silently give incorrect results).
  'incorrect results).', category=RuntimeWarning)
/usr/local/lib/python3.6/dist-packages/torch/tensor.py:468: TracerWarning: Converting a tensor to a Python index might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  return iter(imap(lambda i: self[i], range(self.size(0))))
/mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:104: TracerWarning: Converting a tensor to a Python index might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  batched_imgs = tensors[0].new_full(tuple(batch_shape), pad_value)
/mnt/Projects/PyTorchDetector/detectron2/detectron2/structures/image_list.py:106: TracerWarning: Converting a tensor to a Python index might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  pad_img[..., : img.shape[-2], : img.shape[-1]].copy_(img)
ok

----------------------------------------------------------------------
Ran 2 tests in 0.178s

OK
```

PR #1217, opened 2020-04-15
Clarify about config wrt "resume" option

theotheo:
Just a little clarification on 'resume' option


detectron2/engine/defaults.py
```diff
@@ -297,7 +297,7 @@ def __init__(self, cfg):
     def resume_or_load(self, resume=True):
         """
         If `resume==True`, and last checkpoint exists, resume from it and load all
-        checkpointables (eg. optimizer and scheduler).
+        checkpointables (eg. optimizer and scheduler), i.e. the config will not be used.
```
ppwwyyxx, line 300:
```suggestion
        checkpointables (eg. optimizer and scheduler), i.e. `cfg.MODEL.WEIGHTS` will not be used.
```

also it makes sense to explain more at Line 224 as well

Issue #1221, opened 2020-04-15
PIL warnings for decompression bomb


xkszltl:
We're seeing some warnings in log during training:
```
/usr/local/lib/python3.6/dist-packages/PIL/Image.py:2800: DecompressionBombWarning: Image size (96487340 pixels) exceeds limit of 89478485 pixels, could be decompression bomb DOS attack.
```
I think PIL is complaining about large normal images, and `Image.MAX_IMAGE_PIXELS = None` should be set since training data is usually "safe enough".

ppwwyyxx:
closing as explained in https://github.com/facebookresearch/detectron2/pull/1223#issuecomment-619808236

PR #1223, opened 2020-04-15
Disable PIL DecompositionBomb threshold for reading large images.

xkszltl:
Fix https://github.com/facebookresearch/detectron2/issues/1221

### DIFF
--- a/detectron2/data/detection_utils.py
+++ b/detectron2/data/detection_utils.py
@@ -46,7 +46,10 @@ def read_image(file_name, format=None):
         image (np.ndarray): an HWC image in the given format.
     """
     with PathManager.open(file_name, "rb") as f:
+        pil_max_px = Image.MAX_IMAGE_PIXELS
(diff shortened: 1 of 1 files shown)


xkszltl:
@ppwwyyxx 
Could you help reviewing this?
Thanks!

ppwwyyxx:
The feature (support images with 1Billion pixels) is used by probably <0.1% users and is not worth the maintenance cost - for example it will make it difficult to switch to a faster image reading library. And a feature we can't maintain should not be added.
You can achieve the same by using a custom dataloader: https://detectron2.readthedocs.io/tutorials/data_loading.html#write-a-custom-dataloader

xkszltl:
Just FYI we do have custom data loader and we can even just put the line at the beginning of code to turn it off globally.
There's nothing to maintain, it simply tells PIL to shut up regardless of image size.

The ad-hoc threshold is quite specific to PIL since it's probably more web-intended.
Other "faster" lib should not need this kind of fix.

PR #1251, opened 2020-04-19
Add `prob` argument for RandomRotations

timofurrer:
The `prob` argument for `RandomRotation` (similar to `RandomFlip`)
allows to control the probability the rotation angle will be applied.

This allows for the original image to be sampled with a certain
probability.

The `prob` argument is appended to the argument list for backwards compatibility.

### DIFF
--- a/detectron2/data/transforms/transform_gen.py
+++ b/detectron2/data/transforms/transform_gen.py
@@ -220,7 +220,9 @@ class RandomRotation(TransformGen):
     number of degrees counter clockwise around the given center.
     """
 
-    def __init__(self, angle, expand=True, center=None, sample_style="range", interp=None):
+    def __init__(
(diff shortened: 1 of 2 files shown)


timofurrer:
I think that the test failures are unrelated to my change, can you confirm?

ppwwyyxx:
It would be better if it is implemented as a separate transform such as https://github.com/pytorch/vision/blob/7b60f4db9707d7afdbb87fd4e8ef6906ca014720/torchvision/transforms/transforms.py#L342-L352 or https://github.com/tensorpack/tensorpack/blob/21a6984c77114da3965e88ad1236983fa39e980a/tensorpack/dataflow/imgaug/meta.py#L18-L30

timofurrer:
Sounds good!

I'll create another PR!

PR #1260, opened 2020-04-20
Implement `crop` option in RotationTransform

timofurrer:
The `crop` option in the `RotationTransform` and `RandomRotation`
transformations can be used to crop the rotated image to the largest
possible axis-aligned rectangle with a maximal area within the rotated
image.

It's disabled by default to ensure backwards compatibility.

The following image depicts the `crop` option:

image


timofurrer:
Is there something I can do to move this PR forward?

PR #1319, opened 2020-04-28
Handle pillow exif orientation flag more correctly

ArtificialNotImbecile:
This pr solves `SizeMismatchError` caused by pillow exif orientation flag when the image is rotated 90/270 degrees(It also handles other orientation which will not raise an error).  See [discussion at stackoverflow](https://stackoverflow.com/questions/4228530/pil-thumbnail-is-rotating-my-image/6218425#6218425)

### DIFF
--- a/detectron2/data/detection_utils.py
+++ b/detectron2/data/detection_utils.py
@@ -42,13 +42,8 @@ def apply_exif_orientation(image):
     if exif is None:
         return image
 
-    exif = {
-        ExifTags.TAGS[k]: v
-        for k, v in exif.items()
-        if k in ExifTags.TAGS
-    }
-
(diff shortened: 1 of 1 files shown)


detectron2/data/detection_utils.py
```diff
@@ -50,7 +89,7 @@ def read_image(file_name, format=None):
 
         # capture and ignore this bug: https://github.com/python-pillow/Pillow/issues/3973
         try:
-            image = ImageOps.exif_transpose(image)
```
ppwwyyxx, line 53: If this function (from PIL) has bugs, please submit the fix to PIL. Thanks!

ppwwyyxx:
Closing as this is a PIL issue.

PR #1323, opened 2020-04-28
Allow DefaultTrainer receive a callable mapper

claverru:
Allowing DefaultTrainer receive a callable mapper on build_train_loader would ease integration with custom data augmentation.  Also, it is backwards compatible as it'is None as default.

Workaround is overriding build_train_loader method, adding a mapper as build_detection_train_loader's argument, which is uglier.


ppwwyyxx:
Whenever you want to call `Trainer.build_train_loader` you can call `build_detection_train_loader` directly. So I don't see in what situation this change would provide a benefit.
If you think otherwise, please provide a concrete example.

Issue #321, opened 2019-11-15
About the cost of GPU during test


lyc6749:
## ❓ Questions and Help

when I use the trained maskrcnn-resnet50-FPN model  for testing， I found it cost about 7~8G GPU memory which is more than the version of maskrcnn-benchmark. How can I do to decrease the 
consumption of GPU?

ppwwyyxx:
```
python tools/train_net.py --config-file configs/COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_1x.yaml --num-gpus 1 --eval-only
```
only takes 2G memory on my machine. If this is the case for you as well, then it means your code is not written correctly.
It's unlikely, but if the same commands takes 8G memory on your environment, then it's an environment issue and please include environment details following the issue template.

lyc6749:
# packages in environment at /home/chenlingying/anaconda3/envs/detectron2:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                        main
absl-py                   0.8.1                    pypi_0    pypi
addict                    2.2.1                    pypi_0    pypi
blas                      1.0                         mkl
ca-certificates           2019.10.16                    0
cachetools                3.1.1                    pypi_0    pypi
certifi                   2019.9.11                py36_0
cffi                      1.13.1           py36h2e261b9_0
chardet                   3.0.4                    pypi_0    pypi
cloudpickle               1.2.2                    pypi_0    pypi
cudatoolkit               10.1.243             h6bb024c_0
cycler                    0.10.0                   pypi_0    pypi
cython                    0.29.14                  pypi_0    pypi
decorator                 4.4.1                    pypi_0    pypi
detectron2                0.1                       dev_0    <develop>
freetype                  2.9.1                h8a8886c_1
fvcore                    0.1                      pypi_0    pypi
google-auth               1.6.3                    pypi_0    pypi
google-auth-oauthlib      0.4.1                    pypi_0    pypi
grpcio                    1.25.0rc1                pypi_0    pypi
idna                      2.8                      pypi_0    pypi
imagecorruptions          1.1.0                    pypi_0    pypi
imageio                   2.6.1                    pypi_0    pypi
imgaug                    0.2.6                    pypi_0    pypi
intel-openmp              2019.4                      243
jpeg                      9b                   h024ee3a_2
kiwisolver                1.1.0                    pypi_0    pypi
libedit                   3.1.20181209         hc058e9b_0
libffi                    3.2.1                hd88cf55_4
[...]

PR #1366, opened 2020-05-05
add flag `do_postprocess` to forward function of GeneralisedRCNN, to disable postprecessing

bendidi:
Hello, 

As described in this old issue [comment](https://github.com/facebookresearch/detectron2/issues/321#issuecomment-568343346), storing a few large masks in the GPU can easily lead to an OOM error, and it happens during the postprocessing phase of the inference/eval, my original idea was to add a `postprocessing_device` kwarg to the `forward` function to define which device to use for the postprocessing but it seems like it can lead to a mess and needed more thinking, 

Instead, I propose this simple change to allow users to disable the default `postprocessing` and apply their own afterward in another device if they chose to.

### DIFF
--- a/detectron2/modeling/meta_arch/rcnn.py
+++ b/detectron2/modeling/meta_arch/rcnn.py
@@ -7,7 +7,6 @@ from torch import nn
 from detectron2.structures import ImageList
 from detectron2.utils.events import get_event_storage
 from detectron2.utils.logger import log_first_n
-from detectron2.utils.memory import retry_if_cuda_oom
 
 from ..backbone import build_backbone
 from ..postprocessing import detector_postprocess
@@ -199,7 +198,7 @@ class GeneralizedRCNN(nn.Module):
         ):
             height = input_per_image.get("height", image_size[0])
             width = input_per_image.get("width", image_size[1])
-            r = retry_if_cuda_oom(detector_postprocess)(results_per_image, height, width)
+            r = detector_postprocess(results_per_image, height, width)
             processed_results.append({"instances": r})
         return processed_results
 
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
I would encourage just using `inference(do_postprocessing=)` if this type of fine-level control is needed. 
There are a few options in `inference()` now, and potentially more in the future if you want to be creative. It would not make sense to duplicate them all in `forward()` especially if they don't make sense for training.

Regarding memory, adding `retry_if_cuda_oom` in postprocessing might also help.

bendidi:
> There are a few options in inference() now, and potentially more in the future if you want to be creative. It would not make sense to duplicate them all in forward() especially if they don't make sense for training.

makes sense and that's what I was doing before, I had a problem when running the periodic evaluation during training, since my model is a `DistributedDataParallel` model, I can't just use the `inference` function directly there while the `__call__` function can take a bunch of kwargs that can propagate down to the `__call__` then `forward` functions of `GeneralisedRCNN`

> Regarding memory, adding retry_if_cuda_oom in postprocessing might also help.
seems like it's the better solution here, but instead of a simple retry if cuda OOM, it will be retry on cpu if cuda OOM

detectron2/modeling/meta_arch/rcnn.py
```diff
@@ -198,7 +199,7 @@ def _postprocess(instances, batched_inputs, image_sizes):
         ):
             height = input_per_image.get("height", image_size[0])
             width = input_per_image.get("width", image_size[1])
-            r = detector_postprocess(results_per_image, height, width)
+            r = retry_if_cuda_oom(detector_postprocess)(results_per_image, height, width)
```
ppwwyyxx, line 202: `results_per_image` is a structure that `retry_if_cuda_oom` probably can't handle.
It's better to use it for (or inside) `paste_masks_in_image`.

bendidi: `results_per_image` is an `Instance` so normally it should support `.to(device)`, but anyway it's as you've said better to use it for `paste_masks_in_image`

From what I've checked it's not a good idea to do it inside since there are different behaviors wether the device is `cpu` or `cuda` (and also the function checks the device of the input to decide)

PR #1385, opened 2020-05-08
config MODEL.RESULT.RADIX

limengfei3675:
I noticed the parameter MODEL.RESULT.RADIX in the config file, but there is nothing about it in the code of detectron2. What is its role? Or how is it used?


ppwwyyxx:
wrong project.

PR #1452, opened 2020-05-19
Better handling empty proposals in FastRCNNOutputs

davidas1:
Fixes issue encountered when training CascadeMaskRCNN - empty proposals are generated for all the images in the batch, causing errors in loss calculation


ppwwyyxx:
Unit tests are failing.
Please also add a unit test that fails before the PR, and can be fixed with the PR.

PR #1462, opened 2020-05-22
Make matcher JIT-capable.

chenbohua3:
One point that needs further discussion is how to handle `torch.nonzero`. As discussed in [here](https://github.com/pytorch/pytorch/issues/38718), currently torchscript does not support `torch.nonzero` when `as_tuple` is explicitly set. I write a `as_tuple=True` version of torch.nonzero to replace that in this pr.

### DIFF
--- a/detectron2/modeling/matcher.py
+++ b/detectron2/modeling/matcher.py
@@ -52,8 +52,8 @@ class Matcher(object):
         thresholds.insert(0, -float("inf"))
         thresholds.append(float("inf"))
         # Currently torchscript does not support all + generator
-        assert all(
-            [low <= high for (low, high) in zip(thresholds[:-1], thresholds[1:])]  # noqa: C407
+        assert all(  # noqa: C407
+            [low <= high for (low, high) in zip(thresholds[:-1], thresholds[1:])]
         )
         assert all([l in [-1, 0, 1] for l in labels])  # noqa: C407
         assert len(labels) == len(thresholds) - 1
--- a/detectron2/modeling/matcher.py
+++ b/detectron2/modeling/matcher.py
@@ -96,6 +96,8 @@ class Matcher(object):
 
         for (l, low, high) in zip(self.labels, self.thresholds[:-1], self.thresholds[1:]):
             low_high = (matched_vals >= low) & (matched_vals < high)
+            # directly assign l to match_labels[low_high] is not supported by torch.jit.
+            # https://github.com/pytorch/pytorch/issues/38962
             match_labels.masked_fill_(low_high, l)
 
         if self.allow_low_quality_matches:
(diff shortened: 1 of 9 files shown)


detectron2/modeling/matcher.py
```diff
@@ -48,8 +51,11 @@ def __init__(
         assert thresholds[0] > 0
         thresholds.insert(0, -float("inf"))
         thresholds.append(float("inf"))
-        assert all(low <= high for (low, high) in zip(thresholds[:-1], thresholds[1:]))
-        assert all(l in [-1, 0, 1] for l in labels)
+        # Currently torchscript does not support all + generator
+        assert all(  # noqa: C407
```
ppwwyyxx, line 55: let's ignore C407 in `.flake8`

detectron2/modeling/matcher.py
```diff
@@ -92,7 +98,7 @@ def __call__(self, match_quality_matrix):
 
         for (l, low, high) in zip(self.labels, self.thresholds[:-1], self.thresholds[1:]):
             low_high = (matched_vals >= low) & (matched_vals < high)
-            match_labels[low_high] = l
+            match_labels.masked_fill_(low_high, l)
```
ppwwyyxx, line 101: can you link to a pytorch issue about the lack of indexing?

detectron2/modeling/matcher.py
```diff
@@ -2,7 +2,10 @@
 from typing import List
 import torch
 
+from detectron2.layers import nonzero_tuple
 
+
+@torch.jit.script
```
ppwwyyxx, line 8: can you leave the class not scripted, but add a test in `modeling/test_matcher` that tests for scriptability? It can be like https://github.com/facebookresearch/detectron2/blob/de098423c675dad38c23110407926ccf2919474d/tests/modeling/test_anchor_generator.py#L70-L75 that asserts the outputs are the same.

ppwwyyxx: unless there is a significant speed improvement to always enable script mode for this module, we'd like to keep things eager by default for better flexibility during development

detectron2/modeling/matcher.py
chenbohua3: done

detectron2/modeling/matcher.py
chenbohua3: done

detectron2/modeling/matcher.py
chenbohua3: done:)

chenbohua3: I come across another problem about jit and link to an pytorch issue in the codes. I'm trying to figure out if there's a more intuitive workaround about it.

PR #1483, opened 2020-05-27
make anchor_generator scriptable

ppwwyyxx:
Differential Revision: D21310754


ppwwyyxx:
FYI to @chenbohua3  @bddppq, torch.jit.unused support is added yesterday

PR #1518, opened 2020-06-02
Make anchor_generator scriptable

chenbohua3:
This is a refined version of #1483  to make anchor_generator scriptable. Thanks for your work @ppwwyyxx :)


ppwwyyxx:
Seems like named_buffer is already available with https://github.com/pytorch/pytorch/pull/37905 landed! Let's clean up the hack

PR #1528, opened 2020-06-03
Adding the input device to the new tensor

bomri:
In case `gt_classes.numel()==0`, an all background tensor will be created by running:
`gt_classes = torch.zeros_like(matched_idxs) + self.num_classes`

This line creates a tensor on the CPU rather than on the GPU. It causes an issue later when trying to concatenate all gt_classes from all proposals. 

I modified it to the following by adding the gt_classes input device:
`gt_classes = torch.zeros_like(matched_idxs, device=gt_classes.device) + self.num_classes
`

### DIFF
--- a/detectron2/modeling/roi_heads/roi_heads.py
+++ b/detectron2/modeling/roi_heads/roi_heads.py
@@ -206,7 +206,7 @@ class ROIHeads(torch.nn.Module):
             # Label ignore proposals (-1 label)
             gt_classes[matched_labels == -1] = -1
         else:
-            gt_classes = torch.zeros_like(matched_idxs) + self.num_classes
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
`matched_idxs` should not be on CPU. To fix the root cause, it would be better to make sure the output of `Matcher` is on the same device as the input of `Matcher`.

bomri:
Thanks @ppwwyyxx you're right, I had `matched_idxs` on CPU.

bomri:
Yes. The problem I had was that `match_quality_matrix` was originally created on the wrong device

yezz123 (approved):

ppwwyyxx:
closing due to no activity. 
The latest status seems to be that it's unclear what was the root cause (i.e. the first tensor whose device is undesired) that should be fixed

PR #1529, opened 2020-06-03
Enabling more flexible feature shapes for FPN

jinyeom:
In certain cases, the current implementation of `FPN` doesn't allow for certain image shapes, and raises errors that look like the following:
```python
RuntimeError: The size of tensor a (17) must match the size of tensor b (18) at non-singleton dimension 2
```
This seems to be coming from the fact that features are interpolated with a scale factor of 2, which in some cases must round down the feature size. This PR addresses this issue by explicitly providing the target top-down feature shape for interpolation.


ppwwyyxx:
The FPN model requires input shape to be a multiple of size_divisibility.
Changing the scale factor can make it run but may lead to worse results, so we'd rather not change it.

Issue #991, opened 2020-03-05
Do we have a quick register way for a custom dataset which is in Pascal VOC format? I tried to use register_pascal_voc but it won't work.


SkeletonOne:
## ❓ How to use Detectron2

Questions like:

1. How to do X with detectron2?
2. How detectron2 does X?

NOTE:

1. If you met any unexpected issue when using detectron2 and wish to know why,
   please use the "Unexpected Problems / Bugs" issue template.

2. We do not answer general machine learning / computer vision questions that are not specific to
	 detectron2, such as how a model works, how to improve your training/make it converge, or what algorithm/methods can be
	 used to achieve X.

ppwwyyxx:
No. The function is made for pascal voc dataset only.

nodechef:
So for custom dataset in Pascal VOC, do we have to write a custom dataloader function ?

ppwwyyxx:
[Use Custom Datasets](https://detectron2.readthedocs.io/tutorials/datasets.html#use-custom-datasets) explains how to use custom datasets.

SkeletonOne:
@nodechef A solution is to use scripts to transfer VOC format to COCO format, then it will be easy to use register_coco_instances to register it.

nodechef:
@SkeletonOne have you developed any script ? I looked into https://github.com/yukkyo/voc2coco
but it's giving some errors. if you haven't than I would probably fix this one.

SkeletonOne:
@nodechef I think the script in https://github.com/yukkyo/voc2coco is correct. Note that the only thing you should pay attention to while training is to remove the 'ignore' annotation, as https://github.com/facebookresearch/detectron2/issues/1089 says.

SkeletonOne:
@nodechef This method is just an easy way to train your dataset and the credibility cannot be guaranteed. I would not recommend using this method for serious purposes.

nodechef:
This ```register_coco_instances``` ? Or converting voc2coco ?

SkeletonOne:
@nodechef Both works well.

nodechef:
I am confused a little bit. Which method were you talking about ? I'm training the model for production purposes. So I need to make sure that everything is implemented in optimized way.

SkeletonOne:
@nodechef 
The function register_coco_instances works perfectly in this situation. 

[...]

PR #1603, opened 2020-06-15
Allow registering pascal voc dataset with custom class names

pakornvs:
Relate: #991

### DIFF
--- a/detectron2/data/datasets/pascal_voc.py
+++ b/detectron2/data/datasets/pascal_voc.py
@@ -21,7 +21,7 @@ CLASS_NAMES = [
 # fmt: on
 
 
-def load_voc_instances(dirname: str, split: str):
+def load_voc_instances(dirname: str, split: str, class_names: list):
     """
     Load Pascal VOC detection annotations to Detectron2 format.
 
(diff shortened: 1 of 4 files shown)


detectron2/data/datasets/pascal_voc.py
```diff
@@ -21,7 +21,7 @@
 # fmt: on
 
 
-def load_voc_instances(dirname: str, split: str):
+def load_voc_instances(dirname: str, split: str, class_names: list):
```
ppwwyyxx, line 24: Should be `List[str]` with `from typing import List`.
Documentation is missing.

detectron2/data/datasets/pascal_voc.py
```diff
@@ -66,15 +66,13 @@ def load_voc_instances(dirname: str, split: str):
...
         dicts.append(r)
     return dicts
 
 
-def register_pascal_voc(name, dirname, split, year):
-    DatasetCatalog.register(name, lambda: load_voc_instances(dirname, split))
-    MetadataCatalog.get(name).set(
-        thing_classes=CLASS_NAMES, dirname=dirname, year=year, split=split
-    )
+def register_pascal_voc(name, dirname, split, class_names=CLASS_NAMES, **kwargs):
```
ppwwyyxx, line 76: Please keep `year` since it's required for evaluation.

Don't use mutable default: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

PR #1608, opened 2020-06-16
fix a bug in voc_eval()

coolermzb3:
# fix a KeyError bug in the function voc_eval()

Based on the definition in line 226, the key of `class_recs` should be `imagename`, instead of `image_id`. 

Actually, `image_id` should be the index of `imagename` in `imagenames` , but it can be seen from line 231~234 that the type of `image_id` is string, and need to be converted into int.


ppwwyyxx:
As far as I know the evaluation works fine. 
Would you give an example when it would fail?

PR #1633, opened 2020-06-19
fix IndexError while using RandomCrop with no annotations

vladkrylov:
**Issue**: when a part of dataset contains no annotations and random cropping is enabled in `config` like the following:
```python
INPUT:
  CROP:
    ENABLED: True
    TYPE: "absolute"
    SIZE: (1080, 1920)
```
 an `IndexError` is raised in line:
https://github.com/facebookresearch/detectron2/blob/cd4d88eacfc15b79c6150769d3f8c0eb72b182d7/detectron2/data/dataset_mapper.py#L91
since the `dataset_dict["annotations"]` is empty.

A small change was made to address this issue.


detectron2/data/dataset_mapper.py
```diff
@@ -84,15 +84,16 @@ def __call__(self, dataset_dict):
         else:
             # Crop around an instance if there are instances in the image.
             # USER: Remove if you don't use cropping
-            if self.crop_gen:
+            crop_around_instance = self.crop_gen and dataset_dict["annotations"]
```
ppwwyyxx, line 87: it's easier to just replace L80 by `if dataset_dict.get("annotations", []):`, to still enable cropping.

vladkrylov: :+1: fixed

Issue #933, opened 2020-02-24
check_image_size    ... bug?


JavierClearImageAI:
I think there is a bug in here: 

https://github.com/facebookresearch/detectron2/blob/316469de26f9ef8914762fd144b2c1b2ce2ffa90/detectron2/data/detection_utils.py#L73

it works wrong with images that are not landscape images... so if I use a vertical video, it shows the next error sometimes:

```
  File "/home/ubuntu/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/detectron2/data/detection_utils.py", line 87, in check_image_size
    expected_wh,
```

```
detectron2.data.detection_utils.SizeMismatchError: Mismatched (W,H) for image /home/ubuntu/new_logs_v5/train/Walnut_Veneer_Logs_0476l.jpg, got (720, 960), expect (960, 720)
```

I compared my json height and weight and the images sizes opening them with CV2, then  im.shape() and they are OK.

ppwwyyxx:
We use `PIL` to read image:
https://github.com/facebookresearch/detectron2/blob/316469de26f9ef8914762fd144b2c1b2ce2ffa90/detectron2/data/detection_utils.py#L49-L69

Please try it and if it reads wrong size, please report bugs to PIL.

PR #1666, opened 2020-06-23
Correctly apply EXIF orientation to images

craymichael:
Workaround for the PIL bug https://github.com/python-pillow/Pillow/issues/3973 - the exif orientation metadata is used to correctly transpose a PIL image when reading an image. The current approach silences any exceptions here, leading to strange behavior, such as issue #933. This is indeed a PIL issue, but the fix here performs the same function that is necessary to correctly orient images. The only differences between the committed function `apply_exif_orientation`  and the PIL `exif_transpose` is that a copy of the image is not generated if no transpose is needed, and that the exif data is not updated in the transposed image. As `read_image` returns a NumPy ndarray, this information is discarded, anyway.

### DIFF
--- a/detectron2/data/detection_utils.py
+++ b/detectron2/data/detection_utils.py
@@ -10,7 +10,7 @@ import numpy as np
 import pycocotools.mask as mask_util
 import torch
 from fvcore.common.file_io import PathManager
-from PIL import Image, ImageOps
+from PIL import Image
 
 from detectron2.structures import (
     BitMasks,
@@ -103,17 +103,17 @@ def convert_image_to_rgb(image, format):
 def apply_exif_orientation(image):
     """
     Applies the exif orientation correctly.
-    
+
     This code exists per the bug:
       https://github.com/python-pillow/Pillow/issues/3973
     with the function `ImageOps.exif_transpose`
-    
+
     Modified from:
       https://github.com/wkentaro/labelme/blob/v4.5.4/labelme/utils/image.py#L59
       https://github.com/python-pillow/Pillow/blob/7.1.2/src/PIL/ImageOps.py#L527
-      
+
     The Pillow source raises errors with various methods, especially `tobytes`
-    
+
     Args:
         image (PIL.Image): a PIL image
 
--- a/detectron2/data/detection_utils.py
+++ b/detectron2/data/detection_utils.py
@@ -120,16 +120,16 @@ def apply_exif_orientation(image):
     Returns:
         (PIL.Image): the PIL image with exif orientation applied, if applicable
     """
-    if not hasattr(image, 'getexif'):
+    if not hasattr(image, "getexif"):
         return image
-    
+
     exif = image.getexif()
-    
+
     if exif is None:
         return image
 
     orientation = exif.get(_EXIF_ORIENT)
-    
+
     method = {
         2: Image.FLIP_LEFT_RIGHT,
         3: Image.ROTATE_180,
@@ -139,7 +139,7 @@ def apply_exif_orientation(image):
         7: Image.TRANSVERSE,
         8: Image.ROTATE_90,
     }.get(orientation)
-    
+
     if method is not None:
         return image.transpose(method)
     return image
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
Thanks for the workaround. From the original issue, it seems the bug has been fixed in Pillow>=7.0 already. In that case it would be better to just require Pillow>=7.0 as a dependency.

craymichael:
I would agree with you - however, that issue certainly is not closed, nor does the current fix address this issue, unfortunately. A quick MWE of the error:

```python
Python 3.7.7 (default, Mar 29 2020, 18:40:05) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from PIL import Image, ImageOps
>>> with open('IMG_5370.jpg', 'rb') as f:
...     img = Image.open(f)
...     img = ImageOps.exif_transpose(img)
... 
/.../venv/lib/python3.7/site-packages/PIL/TiffImagePlugin.py:590: UserWarning: Metadata Warning, tag 282 had too many entries: 2, expected 1
  % (tag, len(values))
/.../venv/lib/python3.7/site-packages/PIL/TiffImagePlugin.py:590: UserWarning: Metadata Warning, tag 283 had too many entries: 2, expected 1
  % (tag, len(values))
  File "<stdin>", line 3, in <module>
  File "/.../venv/lib/python3.7/site-packages/PIL/ImageOps.py", line 549, in exif_transpose
    transposed_image.info["exif"] = exif.tobytes()
  File "/.../venv/lib/python3.7/site-packages/PIL/Image.py", line 3292, in tobytes
    return b"Exif\x00\x00" + head + ifd.tobytes(offset)
  File "/.../venv/lib/python3.7/site-packages/PIL/TiffImagePlugin.py", line 808, in tobytes
    data = self._write_dispatch[typ](self, *values)
TypeError: write_undefined() takes 2 positional arguments but 5 were given
>>> import PIL
>>> PIL.__version__
'7.1.2'
```

This in current state would ignore the exif orientation value of this image (which is `6`), thus the image would be read in by detectron without a rotation, in turn having incorrect dimensions that would cause the image width/height check to fail (and the mask to be incorrect).

detectron2/data/detection_utils.py
```diff
@@ -97,6 +100,51 @@ def convert_image_to_rgb(image, format):
     return image
 
 
+def apply_exif_orientation(image):
```
ppwwyyxx, line 103: prefix it by `_` since we don't expect it to be used outside of this file

detectron2/data/detection_utils.py
```diff
@@ -97,6 +100,51 @@ def convert_image_to_rgb(image, format):
...
+
+    This code exists per the bug:
+      https://github.com/python-pillow/Pillow/issues/3973
+    with the function `ImageOps.exif_transpose`
+
+    Modified from:
+      https://github.com/wkentaro/labelme/blob/v4.5.4/labelme/utils/image.py#L59
+      https://github.com/python-pillow/Pillow/blob/7.1.2/src/PIL/ImageOps.py#L527
+
+    The Pillow source raises errors with various methods, especially `tobytes`
```
ppwwyyxx, line 115: it seems this line stay with L107-L109 that explains the bug

detectron2/data/detection_utils.py
craymichael: Good call

detectron2/data/detection_utils.py
craymichael: Right, I'll clean this part up then and push soon

ppwwyyxx:
Can you also report this upstream as well, so there is a chance it gets fixed in Pillow? We obviously don't want to keep a workaround forever.

craymichael:
I definitely agree with that. I just looked into this further. The Pillow PR https://github.com/python-pillow/Pillow/pull/4637 appears to fix this issue. I just tested it with the image that gave me the error with Pillow 7.1.2 successfully using the dev version of Pillow (7.2.0.dev0). It looks like 7.2.0 won't be released until July 1st, 2020 https://github.com/python-pillow/Pillow/issues/4658, however. The original Pillow issue is still open, and I am unsure if there are other cases here that may result in raised exceptions with the exif_transpose function.

PR #1683, opened 2020-06-26
allow convert_to_coco_dict for custom dataset without annotations

### LINKED ISSUE #1684: (coco-style) Inference fails for custom dataset without annotations
When testing for a custom dataset without annotations, it would fail to convert to coco dict.

wangg12:
Fix #1684 . Allowing inference when annotations are not available for custom datasets.

### DIFF
--- a/detectron2/data/datasets/coco.py
+++ b/detectron2/data/datasets/coco.py
@@ -318,7 +318,7 @@ def convert_to_coco_dict(dataset_name):
         }
         coco_images.append(coco_image)
 
-        anns_per_image = image_dict["annotations"]
+        anns_per_image = image_dict.get("annotations", [])
         for annotation in anns_per_image:
             # create a new dict with only COCO fields
             coco_annotation = {}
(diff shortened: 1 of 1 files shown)


wangg12:
Fix #1684

ppwwyyxx:
Please describe the problem first. Currently, the issue is empty.

wangg12:
I think the issue already describes the problem clearly:)

ppwwyyxx:
An easier solution is to let `convert_to_coco_json` support dataset without annotations (instead of throwing an error). That should be doable with <5 lines.

wangg12:
Maybe using
```
anns_per_image = image_dict.get("annotations", [])
```
in https://github.com/facebookresearch/detectron2/blob/master/detectron2/data/datasets/coco.py#L321.
I'm not sure if this is sufficient.

wangg12:
I tested this, it works for me by just changing one line.

ppwwyyxx:
Not sure whether https://github.com/facebookresearch/detectron2/blob/e45cd680d52569494ea2110fbd2ec3d6b48a5af9/detectron2/data/datasets/coco.py#L398 needs to be changed as well, given this line in evaluator:
https://github.com/facebookresearch/detectron2/blob/e45cd680d52569494ea2110fbd2ec3d6b48a5af9/detectron2/evaluation/coco_evaluation.py#L86

wangg12:
By doing this, the evaluation will still be performed. But got all nan results.

wangg12:
Now it should be fine. If there are no annotations at all, the converted coco json file will not contain `"annotations"`.

PR #1734, opened 2020-07-08
Add numpy promise

Jianghanxiao:
When transforms.transforms is [], then after apply, the bbox will still be a list, then .clip will post errors

### DIFF
--- a/detectron2/data/detection_utils.py
+++ b/detectron2/data/detection_utils.py
@@ -276,7 +276,7 @@ def transform_instance_annotations(
     # bbox is 1d (per-instance bounding box)
     bbox = BoxMode.convert(annotation["bbox"], annotation["bbox_mode"], BoxMode.XYXY_ABS)
     # clip transformed bbox to image size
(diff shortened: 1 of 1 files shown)


detectron2/data/detection_utils.py
```diff
@@ -276,7 +276,7 @@ def transform_instance_annotations(
     # bbox is 1d (per-instance bounding box)
     bbox = BoxMode.convert(annotation["bbox"], annotation["bbox_mode"], BoxMode.XYXY_ABS)
     # clip transformed bbox to image size
-    bbox = transforms.apply_box([bbox])[0].clip(min=0)
+    bbox = np.array(transforms.apply_box([bbox])[0]).clip(min=0)
```
ppwwyyxx, line 279: Thanks for the catch! Semantically, I think it would be more correct to do 
```suggestion
    bbox = transforms.apply_box(np.array([bbox]))[0].clip(min=0)
```
since `apply_box` does require a numpy array as input but we did not follow it

PR #1735, opened 2020-07-08
Make batched_nms scriptable.

chenbohua3:
A subtask of making RPN scriptable because `batched_nms` is often the last calculation step of normal RPN

### DIFF
--- a/tests/layers/test_nms.py
+++ b/tests/layers/test_nms.py
@@ -1,6 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 from __future__ import absolute_import, division, print_function, unicode_literals
-import numpy as np
 import unittest
 import torch
 
--- a/tests/layers/test_nms.py
+++ b/tests/layers/test_nms.py
@@ -1,0 +1,69 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+from __future__ import absolute_import, division, print_function, unicode_literals
+import numpy as np
+import unittest
+import torch
+
+from detectron2.layers import batched_nms
+from detectron2.utils.env import TORCH_VERSION
+
+
+def nms_edit_distance(keep1, keep2):
+    """
+    Compare the "keep" result of two nms call.
+    They are allowed to be different in terms of edit distance
+    due to floating point precision issues, e.g.,
+    if a box happen to have an IoU of 0.5 with another box,
+    one implentation may choose to keep it while another may discard it.
+    """
+    if torch.equal(keep1, keep2):
+        # they should be equal most of the time
+        return 0
+    keep1, keep2 = tuple(keep1.cpu()), tuple(keep2.cpu())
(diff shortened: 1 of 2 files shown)


tests/layers/test_nms_rotated.py
```diff
@@ -87,6 +88,13 @@ def test_batched_nms_rotated_0_degree_cpu(self):
             backup = boxes.clone()
             keep_ref = batched_nms(boxes, scores, idxs, iou)
             assert torch.allclose(boxes, backup), "boxes modified by batched_nms"
+            # test batched_nms's scriptability
+            if TORCH_VERSION >= (1, 7):
```
ppwwyyxx, line 92: does it need 1.7? 
If it's about `unique`, it should be supported in 1.6: https://github.com/pytorch/pytorch/commit/eb3e9872c9ed038ead67a832e7e5295628cbb55a

tests/layers/test_nms_rotated.py
```diff
@@ -87,6 +88,13 @@ def test_batched_nms_rotated_0_degree_cpu(self):
             backup = boxes.clone()
```
ppwwyyxx, line 88: Can you create a separate test method to test scriptability? Also this class and this file can be renamed to just NMS, since it's no longer specific to NMSRotated.

tests/layers/test_nms_rotated.py
chenbohua3: let me test it in 1.6

tests/layers/test_nms_rotated.py
chenbohua3: Got it. I will create `test_nms.py` and move this ut to it.

tests/layers/test_nms.py
```diff
@@ -0,0 +1,69 @@
...
+        idxs = torch.randint(0, num_classes, (N,))
+        scripted_batched_nms = torch.jit.script(batched_nms)
+        err_msg = "NMS is incompatible with jit-scripted NMS for IoU={}"
+
+        for iou in [0.2, 0.5, 0.8]:
+            backup = boxes.clone()
+            keep_ref = batched_nms(boxes, scores, idxs, iou)
+            scripted_keep = scripted_batched_nms(boxes, scores, idxs, iou)
+            assert torch.allclose(boxes, backup), "boxes modified by jit-scripted batched_nms"
+            self.assertEqual(nms_edit_distance(keep_ref, scripted_keep), 0, err_msg.format(iou))
```
ppwwyyxx, line 65: If the distance tolerance is 0 (probably should be, since they are using the same ops), this is equivalent to `torch.equal`

chenbohua3: you mean `self.assertEqual(torch.equal(keep_ref, scripted_keep), True, err_msg.format(iou))`? Or just delete assertEqual: `assert torch.equal(keep_ref, scripted_keep)`

chenbohua3: done

PR #1758, opened 2020-07-10
Bitmask tight bounding boxes

constiDisch:
Just implement bit mask get_bounding_boxes function

### DIFF
--- a/detectron2/structures/masks.py
+++ b/detectron2/structures/masks.py
@@ -212,7 +212,9 @@ class BitMasks:
             x = torch.where(x_any[idx, :])[0]
             y = torch.where(y_any[idx, :])[0]
             if len(x) > 0 and len(y) > 0:
-                boxes[idx, :] = torch.as_tensor([x[0], y[0], x[-1], y[-1]], dtype=torch.float32)
+                boxes[idx, :] = torch.as_tensor(
+                    [x[0], y[0], x[-1] + 1, y[-1] + 1], dtype=torch.float32
+                )
         return Boxes(boxes)
 
     @staticmethod
--- a/detectron2/structures/masks.py
+++ b/detectron2/structures/masks.py
@@ -211,7 +211,6 @@ class BitMasks:
             boxes[idx, :] = torch.as_tensor([x[0], y[0], x[-1], y[-1]], dtype=torch.float32)
         return Boxes(boxes)
 
-
     @staticmethod
     def cat(bitmasks_list: List["BitMasks"]) -> "BitMasks":
         """
--- a/detectron2/structures/masks.py
+++ b/detectron2/structures/masks.py
@@ -208,7 +208,10 @@ class BitMasks:
         for idx in range(self.tensor.shape[0]):
             x = torch.where(torch.any(self.tensor[idx, :], dim=0))[0]
             y = torch.where(torch.any(self.tensor[idx, :], dim=1))[0]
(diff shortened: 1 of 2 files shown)


detectron2/structures/masks.py
```diff
@@ -199,9 +199,17 @@ def crop_and_resize(self, boxes: torch.Tensor, mask_size: int) -> torch.Tensor:
...
-        raise NotImplementedError
+    def get_bounding_boxes(self) -> Boxes:
+        """
+        Returns:
+            Boxes: tight bounding boxes around bit masks.
+        """
+        boxes = torch.zeros(self.tensor.shape[0], 4, dtype=torch.float32)
+        for idx in range(self.tensor.shape[0]):
+            x = torch.where(torch.any(self.tensor[idx, :], dim=0))[0]
+            y = torch.where(torch.any(self.tensor[idx, :], dim=1))[0]
```
ppwwyyxx, line 210: The two `torch.any` operations can be done outside the forloop on the entire tensor

detectron2/structures/masks.py
```diff
@@ -199,9 +199,20 @@ def crop_and_resize(self, boxes: torch.Tensor, mask_size: int) -> torch.Tensor:
         output = output >= 0.5
         return output
 
-    def get_bounding_boxes(self) -> None:
-        # not needed now
-        raise NotImplementedError
+    def get_bounding_boxes(self) -> Boxes:
+        """
+        Returns:
+            Boxes: tight bounding boxes around bit masks.
```
ppwwyyxx, line 205:
```suggestion
            Boxes: tight bounding boxes around bitmasks. If a mask is empty, it's bounding box will be all zero.
```

detectron2/structures/masks.py
```diff
@@ -199,9 +199,20 @@ def crop_and_resize(self, boxes: torch.Tensor, mask_size: int) -> torch.Tensor:
...
+        """
+        boxes = torch.zeros(self.tensor.shape[0], 4, dtype=torch.float32)
+        x_any = torch.any(self.tensor, dim=1)
+        y_any = torch.any(self.tensor, dim=2)
+        for idx in range(self.tensor.shape[0]):
+            x = torch.where(x_any[idx, :])[0]
+            y = torch.where(y_any[idx, :])[0]
+            if len(x) > 0 and len(y) > 0:
+                boxes[idx, :] = torch.as_tensor([x[0], y[0], x[-1], y[-1]], dtype=torch.float32)
```
ppwwyyxx, line 214: in detectron2, bounding box coordinates are defined to be floating point numbers that range from 0 to H or W, not H-1 or W-1. See https://github.com/tensorpack/tensorpack/blob/master/examples/FasterRCNN/NOTES.md#model for an illustration of coordinates.
So here it should be `x[-1] + 1` and `y[-1] + 1`. Consequently a mask with only one pixel will become a box of size 1x1, which matches our definition.

tests/structures/test_masks.py
```diff
@@ -0,0 +1,34 @@
...
+                    [False, True, True, False],
+                    [False, True, True, False],
+                    [False, False, False, False],
+                ],
+                [
+                    [False, False, False, False],
+                    [False, False, False, False],
+                    [False, False, False, False],
+                    [False, False, False, False],
+                    [False, False, False, False],
```
ppwwyyxx, line 23: replacing it by `torch.zeros(4, 4)` should work

PR #1765, opened 2020-07-14
Make RPN scriptable.

chenbohua3:


detectron2/modeling/proposal_generator/rpn.py
```diff
@@ -431,24 +431,25 @@ def forward(
...
 
-        if self.training:
+        if self.training and not torch.jit.is_scripting():
             gt_labels, gt_boxes = self.label_and_sample_anchors(anchors, gt_instances)
             losses = self.losses(
                 anchors, pred_objectness_logits, gt_labels, pred_anchor_deltas, gt_boxes
             )
-            losses = {k: v * self.loss_weight.get(k, 1.0) for k, v in losses.items()}
+            for k, v in losses.items():
+                losses[k] = v * self.loss_weight.get(k, 1.0)
```
ppwwyyxx, line 440: please add a note linked to a pytorch feature request about dict comprehension, and boolean in dictionary.

detectron2/modeling/proposal_generator/rpn.py
```diff
@@ -431,24 +431,25 @@ def forward(
...
         else:
             losses = {}
-
+        pred_objectness_logits = [t.detach() for t in pred_objectness_logits]
+        pred_anchor_deltas = [t.detach() for t in pred_anchor_deltas]
```
ppwwyyxx, line 444: move it into `predict_proposals`? this will make the semantics of `predict_proposals` unchanged

detectron2/modeling/proposal_generator/rpn.py
```diff
@@ -431,24 +431,25 @@ def forward(
...
         proposals = self.predict_proposals(
             anchors, pred_objectness_logits, pred_anchor_deltas, images.image_sizes
         )
         return proposals, losses
 
-    @torch.no_grad()
```
ppwwyyxx, line 448: add a TODO: use torch.no_grad when torchscript supports it

detectron2/structures/boxes.py
```diff
@@ -211,8 +211,7 @@ def nonempty(self, threshold: float = 0.0) -> torch.Tensor:
         keep = (widths > threshold) & (heights > threshold)
         return keep
 
-    @torch.jit.unused
-    def __getitem__(self, item: Union[int, slice, torch.BoolTensor]):
+    def __getitem__(self, item):
```
ppwwyyxx, line 214: mention the type as comment

detectron2/utils/patch_instance.py
```diff
@@ -0,0 +1,93 @@
...
+        # getter
+        lines.append(indent(1, "def get_{}(self) -> {}:".format(name, type_)))
+        lines.append(indent(2, "val = self.{}".format(name)))
+        lines.append(indent(2, "assert val is not None"))
+        lines.append(indent(2, "return val"))
+        lines.append("")
+        # setter
+        lines.append(indent(1, "def set_{}(self, val: {}):".format(name, type_)))
+        lines.append(indent(2, "self.{} = val".format(name)))
+        lines.append("")
```
ppwwyyxx, line 73: getter/setter doesn't seem useful now? pytorch will soon support `@property`, we can add property later then.

detectron2/utils/patch_instance.py
```diff
@@ -0,0 +1,93 @@
...
+        module = _import(f.name)
+        new_instance = getattr(module, cls_name)
+        _ = torch.jit.script(new_instance)
+
+        Instances.__torch_script_class__ = True
+        Instances._jit_override_qualname = torch._jit_internal._qualified_name(new_instance)
+        return torch.jit.script(model), new_instance
+
+
+def _typing_imports_str():
```
ppwwyyxx, line 28: return a multi-line string directly to be more readable

detectron2/utils/patch_instance.py
```diff
@@ -0,0 +1,93 @@
...
+        return " " * 4 * level + s
+
+    lines = []
+
+    cls_name = "Instance{}".format(_counter)
+    lines.append("class {}:".format(cls_name))
+    lines.append("")
+
+    lines.append(indent(1, "def __init__(self, image_size: Tuple[int, int]):"))
+    lines.append(indent(2, "self.image_size = image_size"))
```
ppwwyyxx, line 56: similarly it can use multi-line string and f-string to improve readability

detectron2/utils/patch_instance.py
```diff
@@ -0,0 +1,93 @@
...
+_counter = 0  # TODO replace by a hash?
+
+
+def _gen_class(fields):
+    def indent(level, s):
+        return " " * 4 * level + s
+
+    lines = []
+
+    cls_name = "Instance{}".format(_counter)
```
ppwwyyxx, line 51: `_counter += 1` after the line

tests/structures/test_imagelist.py
```diff
@@ -44,7 +44,10 @@ def test_imagelist_scriptability(self):
         image_tensor = torch.randn((image_nums, 10, 20), dtype=torch.float32)
         image_shape = [(10, 20)] * image_nums
 
-        ImageList_script = torch.jit.script(ImageList)
+        qualified_name = torch._jit_internal._qualified_name(ImageList)
+        ImageList_script = torch.jit._state._get_script_class(qualified_name)
+        if not ImageList_script:
+            ImageList_script = torch.jit.script(ImageList)
```
ppwwyyxx, line 50: what's the reason for this change?

tests/modeling/test_rpn.py
```diff
@@ -77,6 +79,68 @@ def test_rpn(self):
             )
             self.assertTrue(torch.allclose(proposal.objectness_logits, expected_objectness_logit))
 
+    @unittest.skipIf(TORCH_VERSION < (1, 6), "Insufficient pytorch version")
+    def test_rpn_scriptability(self):
+        torch.manual_seed(121)
```
ppwwyyxx, line 84: The expected output was tested in the method above already. In this test we can just use random inputs and only test that the two RPN computes identical results (i.e. no need to compare with `expected_*` again).

detectron2/modeling/proposal_generator/rpn.py
chenbohua3: done

detectron2/modeling/proposal_generator/rpn.py
chenbohua3: done

detectron2/structures/boxes.py
chenbohua3: it seems that the type is already mentioned in comment, do you mean make it more detailed?

tests/structures/test_imagelist.py
chenbohua3: It seems that exporting RPN to torchscript will register `ImageList` to JIT. Without these changes we will get:
`RuntimeError: class '__torch__.detectron2.structures.image_list.ImageList' already defined.`

detectron2/structures/boxes.py
ppwwyyxx: didn't notice that! this is fine

detectron2/export/patch_instance.py
```diff
@@ -0,0 +1,88 @@
...
+
+def _typing_imports_str():
+
+    import_str = """
+import torch
+from torch import Tensor
+import typing
+from typing import *
+
+from detectron2.structures import Boxes
```
ppwwyyxx, line 38: I'm wondering how to enable import of other classes that might appear in `Instances`. A naive way is to let users pass them in, but maybe something can be done using the qualname of the field type? Maybe left a TODO here for future work

detectron2/export/patch_instance.py
```diff
@@ -0,0 +1,88 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import importlib.util
+import os
+import sys
+import tempfile
+import torch
+
+from detectron2.structures import Boxes, Instances  # noqa F401
```
ppwwyyxx, line 9: add link to https://github.com/pytorch/pytorch/issues/38964 if related

detectron2/export/patch_instance.py
```diff
@@ -0,0 +1,88 @@
...
+
+_counter = 0  # TODO replace by a hash?
```
ppwwyyxx, line 11: maybe the comment isn't needed. I feel that a counter might be good enough and not sure whether a hash will be better in any way

detectron2/export/patch_instance.py
```diff
@@ -0,0 +1,88 @@
...
+        f.flush()
+        f.close()
+
+        module = _import(f.name)
+        new_instance = getattr(module, cls_name)
+        _ = torch.jit.script(new_instance)
+
+        Instances.__torch_script_class__ = True
+        Instances._jit_override_qualname = torch._jit_internal._qualified_name(new_instance)
+        return torch.jit.script(model), new_instance
```
ppwwyyxx, line 27: what's the use case to return `new_instance`?

detectron2/utils/patch_instance.py
chenbohua3: done

detectron2/utils/patch_instance.py
chenbohua3: done

detectron2/utils/patch_instance.py
chenbohua3: done

tests/structures/test_imagelist.py
chenbohua3: I just run the test through the ways described [here](https://github.com/facebookresearch/detectron2/blob/master/tests/README.md) and get this error.
How about add `torch.jit.script(ImageList)` after `export_torchscript_with_patch_instance` in `test_rpn.py` and run this test file alone, This will cause the same error in my development environment.

tests/modeling/test_rpn.py
chenbohua3: done, now i just compare the outputs of the two RPN.

detectron2/export/patch_instance.py
chenbohua3: Used in the UT of RPN. Before executing the forward function of scripted_RPN class, it seems that JIT will 'inspect' all the attributes of scripted_Instance (e.g. the new Instance) like `scripted_Instance.proposal_boxes`. If we pass origin Instance to scripted_RPN, `__getattr__` of origin Instance class will be executed. But in the meantime, `proposal_boxes` is not exist in the `_fields`, AttributeError will be raised.

detectron2/modeling/proposal_generator/proposal_utils.py
```diff
@@ -100,7 +97,7 @@ def find_top_rpn_proposals(
         boxes.clip(image_size)
 
         # filter empty boxes
-        keep = boxes.nonempty(threshold=min_box_size)
+        keep = boxes.nonempty(threshold=float(min_box_size))
```
ppwwyyxx, line 100: This change can be removed if we fix the signature of `find_top_rpn_proposals` to use `float`, and add `float(min_box_size)` in `RPN.__init__`

detectron2/export/patch_instance.py
ppwwyyxx: I see. I think there are two issues to addressed:
1. We should only test for inference mode in the unittest for now. Most of the training codepath is unsupported anyway.
2. Even in inference mode, a unittest might still need to pass `Instances` as input (e.g. in `ROIHeads`). However I think this use case is almost only used by unittest but not by users. So we should provide to users a simpler API.

With the above suggestion on contextmanager, we can use 
```
with patch_instance(...) as new_instance:
    s = torch.jit.script(model)
```
in unittests, but package them together to users in a function without the extra return value that they don't use.

detectron2/modeling/proposal_generator/rpn.py
```diff
@@ -431,24 +433,27 @@ def forward(
             for x in pred_anchor_deltas
         ]
 
-        if self.training:
+        if self.training and not torch.jit.is_scripting():
```
ppwwyyxx, line 436: I found that we can remove some hacks by doing this (requires https://github.com/pytorch/pytorch/pull/41496):
```diff
--- a/detectron2/modeling/proposal_generator/rpn.py
+++ b/detectron2/modeling/proposal_generator/rpn.py
@@ -266,8 +266,10 @@
         label.scatter_(0, neg_idx, 0)
         return label
 
+    @torch.jit.unused
     @torch.no_grad()
-    def label_and_sample_anchors(self, anchors: List[Boxes], gt_instances: List[Instances]):
+    def label_and_sample_anchors(self, anchors: List[Boxes], gt_instances:
+            List[Instances]) -> Tuple[List[torch.Tensor], List[torch.Tensor]]:
         """
         Args:
             anchors (list[Boxes]): anchors for each feature map.
@@ -323,14 +325,15 @@
             matched_gt_boxes.append(matched_gt_boxes_i)
         return gt_labels, matched_gt_boxes
 
+    @torch.jit.unused
     def losses(
         self,
-        anchors,
+        anchors: List[Boxes],
         pred_objectness_logits: List[torch.Tensor],
         gt_labels: List[torch.Tensor],
         pred_anchor_deltas: List[torch.Tensor],
         gt_boxes: List[torch.Tensor],
-    ):
+    ) -> Dict[str, torch.Tensor]:
         """
         Return the losses from a set of RPN predictions and their associated ground-truth.
 
@@ -390,10 +393,13 @@
             reduction="sum",
         )
         normalizer = self.batch_size_per_image * num_images
-        return {
+        ret = {
             "loss_rpn_cls": objectness_loss / normalizer,
             "loss_rpn_loc": localization_loss / normalizer,
         }
+        losses = {k: v * self.loss_weight.get(k, 1.0) for k, v in
+                ret.items()}
+        return losses
 
     def forward(
         self,
@@ -433,15 +439,12 @@
             for x in pred_anchor_deltas
         ]
 
-        if self.training and not torch.jit.is_scripting():
+        if self.training:
+            assert gt_instances is not None, "RPN requires gt_instances in training!"
             gt_labels, gt_boxes = self.label_and_sample_anchors(anchors, gt_instances)
-            losses = self.losses(
+            return self.losses(
                 anchors, pred_objectness_logits, gt_labels, pred_anchor_deltas, gt_boxes
             )
-            # Currently dict comprehension is not supported by JIT.
-            # https://github.com/pytorch/pytorch/issues/41448
-            for k, v in losses.items():
-                losses[k] = v * self.loss_weight.get(k, 1.0)
         else:
             losses = {}
         proposals = self.predict_proposals(
···

File: tests/modeling/test_rpn.py (hunk 10 above)
Comment (ppwwyyxx) [reply tests/modeling/test_rpn.py:84]: no longer need to fixed the seed, since we don't have a fixed expected output any more

File: tests/structures/test_imagelist.py (hunk 9 above)
Comment (ppwwyyxx) [reply tests/structures/test_imagelist.py:50]: Turns out that a non-module class object cannot be scripted twice. This will be a better way to write unittest:
```diff
-        ImageList_script = torch.jit.script(ImageList)
+        def f(image_tensor, image_shape: List[Tuple[int, int]]):
+            return ImageList(image_tensor, image_shape)
 
-        ret = ImageList(image_tensor, image_shape)
-        ret_script = ImageList_script(image_tensor, image_shape)
+        ret = f(image_tensor, image_shape)
+        ret_script = torch.jit.script(f)(image_tensor, image_shape)
+        print(ret_script)
```

chenbohua3: done

detectron2/modeling/proposal_generator/rpn.py
chenbohua3: looks better! Modification is much cleaner by moving codes to a context that is decorated by `torch.jit.unused`

detectron2/export/patch_instance.py
chenbohua3: done

detectron2/export/patch_instance.py
```diff
@@ -0,0 +1,88 @@
...
+import sys
+import tempfile
+import torch
+
+from detectron2.structures import Boxes, Instances  # noqa F401
+
+_counter = 0  # TODO replace by a hash?
+
+
+def export_torchscript_with_patch_instance(model, fields):
```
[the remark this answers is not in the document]
chenbohua3: done

detectron2/export/patch_instance.py
chenbohua3: done, add a todo

detectron2/modeling/proposal_generator/rpn.py
```diff
@@ -431,24 +433,27 @@ def forward(
...
             gt_labels, gt_boxes = self.label_and_sample_anchors(anchors, gt_instances)
             losses = self.losses(
                 anchors, pred_objectness_logits, gt_labels, pred_anchor_deltas, gt_boxes
             )
-            losses = {k: v * self.loss_weight.get(k, 1.0) for k, v in losses.items()}
+            # Currently dict comprehension is not supported by JIT.
```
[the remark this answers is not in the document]
chenbohua3: done, use `torchscript` uniformly

detectron2/export/patch_instance.py
```diff
@@ -0,0 +1,88 @@
...
+
+    for name, type_ in fields.items():
+        lines.append(
+            indent(2, "self.{} = torch.jit.annotate(Optional[{}], None)".format(name, type_))
+        )
+
+    return cls_name, os.linesep.join(lines)
+
+
+def gen_class(fields):
```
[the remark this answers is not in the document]
chenbohua3: done

detectron2/export/patch_instance.py
```diff
@@ -0,0 +1,88 @@
...
+        module = _import(f.name)
+        new_instance = getattr(module, cls_name)
+        _ = torch.jit.script(new_instance)
+
+        Instances.__torch_script_class__ = True
+        Instances._jit_override_qualname = torch._jit_internal._qualified_name(new_instance)
+        return torch.jit.script(model), new_instance
+
+
+def _typing_imports_str():
```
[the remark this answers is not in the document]
chenbohua3: done

detectron2/export/patch_instance.py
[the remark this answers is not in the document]
chenbohua3: done

detectron2/export/patch_instance.py
chenbohua3: done, I feel the use of `counter` is good enough currently.

detectron2/export/patch_instance.py
chenbohua3: done.  `export_torchscript_with_patch_instance` only returns torchscript now.

detectron2/modeling/proposal_generator/proposal_utils.py
chenbohua3: done. I also change `_C.MODEL.PROPOSAL_GENERATOR.MIN_SIZE` in `detectron2/config/defaults.py` from 0 to 0.0 which make it more clear for users.

detectron2/export/patch_instance.py
```diff
@@ -0,0 +1,88 @@
...
+        f.write(s)
+        f.flush()
+        f.close()
+
+        module = _import(f.name)
+        new_instance = getattr(module, cls_name)
+        _ = torch.jit.script(new_instance)
+
+        Instances.__torch_script_class__ = True
+        Instances._jit_override_qualname = torch._jit_internal._qualified_name(new_instance)
```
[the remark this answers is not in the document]
chenbohua3: After step 1, the scripted class will be registered in JIT.  When execute step 2, the same class can be exported successfully whether the hack is reverted or not. So may be the test case will like :
```
 fields = {"proposal_boxes": "Boxes", "objectness_logits": "Tensor"}

        def f(instance: Instances):
            return instance

        def g(instance: Instances):
            return instance
        
        def h(instance: Instances):
            return instance

        with patch_instance(fields) as new_instances:
            torch.jit.script(f)  

        torch.jit.script(g)  # expect failure

        fields = {"mask": "Tensor"}
        with patch_instance(fields) as new_instances:
            torch.jit.script(h)  
```

tests/modeling/test_rpn.py
```diff
@@ -77,6 +79,49 @@ def test_rpn(self):
             )
             self.assertTrue(torch.allclose(proposal.objectness_logits, expected_objectness_logit))
 
+    @unittest.skipIf(TORCH_VERSION < (1, 6), "Insufficient pytorch version")
```
ppwwyyxx, line 82: now it will need 1.7

detectron2/config/defaults.py
```diff
@@ -158,7 +158,7 @@
 _C.MODEL.PROPOSAL_GENERATOR.NAME = "RPN"
 # Proposal height and width both need to be greater than MIN_SIZE
 # (a the scale used during training or inference)
-_C.MODEL.PROPOSAL_GENERATOR.MIN_SIZE = 0
+_C.MODEL.PROPOSAL_GENERATOR.MIN_SIZE = 0.0
```
ppwwyyxx, line 161: I think this may break an old config, when trying to load an old config file that has "0". `yacs` checks types of values and may complain about integer vs. float.

tests/modeling/test_rpn.py
```diff
@@ -77,6 +79,49 @@ def test_rpn(self):
...
+    def test_rpn_scriptability(self):
+        torch.manual_seed(121)
```
ppwwyyxx, line 84: no need to fix the seed

tests/modeling/test_rpn.py
```diff
@@ -77,6 +79,49 @@ def test_rpn(self):
...
+
+        box_data = [[1, 1, 3, 3], [2, 2, 6, 6]]
+        all_gt_instances = []
+        for t in box_data:
+            gt_boxes = torch.tensor([t], dtype=torch.float32)
+            gt_instance = new_instance(image_shape)
+            gt_instance.gt_boxes = Boxes(gt_boxes)
+            all_gt_instances.append(gt_instance)
+
+        with EventStorage():  # capture events in a new storage to discard them
```
ppwwyyxx, line 112: we should run test in eval mode only, and in that case this is not needed

tests/modeling/test_rpn.py
chenbohua3: done

tests/modeling/test_rpn.py
chenbohua3: done

detectron2/config/defaults.py
chenbohua3: done

tests/modeling/test_rpn.py
chenbohua3: done

detectron2/export/patch_instance.py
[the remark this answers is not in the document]
chenbohua3: Test case for new_fields has been added. Is there any needs to add a TODO for **script the same class**?

PR #1782, opened 2020-07-16
Add PyTorch 1.6 to CI, fix errors related to new release

seemethere:
Results from the integration tests that were ran can be found here: https://app.circleci.com/pipelines/github/pytorch/pytorch-integration-testing/28/workflows/80c9c991-949b-4928-a89b-a26c5f023f8f/jobs/149

Ran into an error message when running tests that use this function for
the PyTorch 1.6 release:

> Integer division of tensors using div or / is no longer supported

Replaced the regular divide with a floor divide

### DIFF
--- a/docker/Dockerfile-circleci
+++ b/docker/Dockerfile-circleci
@@ -14,4 +14,4 @@ RUN wget -q https://bootstrap.pypa.io/get-pip.py && \
 RUN pip install tensorboard opencv-python-headless
 ARG PYTORCH_VERSION
 ARG TORCHVISION_VERSION
-RUN pip install torch==${PYTORCH_VERSION} torchvision==${TORCHVISION_VERSION} -f https://download.pytorch.org/whl/cu101/torch_stable.html
+RUN pip install torch==${PYTORCH_VERSION} torchvision==${TORCHVISION_VERSION} -f https://download.pytorch.org/whl/cu101/torch_stable.html -f https://download.pytorch.org/whl/test/cu101/torch_test.html
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,10 +157,10 @@ workflows:
     jobs:
       - cpu_tests
       - gpu_tests:
-          name: gpu_tests_pytorch1.4
-          pytorch_version: '1.4'
-          torchvision_version: '0.5'
-      - gpu_tests:
           name: gpu_tests_pytorch1.5
           pytorch_version: '1.5'
           torchvision_version: '0.6'
+      - gpu_tests:
+          name: gpu_tests_pytorch1.6
+          pytorch_version: '1.6'
+          torchvision_version: '0.7'
--- a/detectron2/modeling/postprocessing.py
+++ b/detectron2/modeling/postprocessing.py
@@ -41,8 +41,8 @@ def detector_postprocess(results, output_height, output_width, mask_threshold=0.
         output_height_tmp = output_height
 
     scale_x, scale_y = (
-        output_width_tmp / results.image_size[1],
-        output_height_tmp / results.image_size[0],
+        output_width_tmp.true_divide(results.image_size[1]),
+        output_height_tmp.true_divide(results.image_size[0]),
     )
     results = Instances((output_height, output_width), **results.get_fields())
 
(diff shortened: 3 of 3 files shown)


ppwwyyxx:
Thanks @seemethere! In fact we expect to do true division there. When the deprecation happened, Mike already helped us migrate this part of code in D21997444. Not sure why it still needs to be changed?

docker/Dockerfile-circleci
```diff
@@ -14,4 +14,4 @@ RUN wget -q https://bootstrap.pypa.io/get-pip.py && \
 RUN pip install tensorboard opencv-python-headless
 ARG PYTORCH_VERSION
 ARG TORCHVISION_VERSION
-RUN pip install torch==${PYTORCH_VERSION} torchvision==${TORCHVISION_VERSION} -f https://download.pytorch.org/whl/cu101/torch_stable.html
+RUN pip install torch==${PYTORCH_VERSION} torchvision==${TORCHVISION_VERSION} -f https://download.pytorch.org/whl/cu101/torch_stable.html -f https://download.pytorch.org/whl/test/cu101/torch_test.html
```
ppwwyyxx, line 17: after stable is released, will this command install 1.6 stable or still 1.6 testing?

seemethere: I believe this should prioritize installs in order of when the indices are noted, for example it should pick out the 1.6 version (if available) in `torch_stable.html` before it picks out the version in `torch_test.html`

seemethere:
> Thanks @seemethere! In fact we expect to do true division there. When the deprecation happened, Mike already helped us migrate this part of code in [D21997444](https://our.intern.facebook.com/intern/diff/D21997444/). Not sure why it still needs to be changed?

I believe the operator alone is not sufficient, and you might need to be more specific given the error message I observed in my testing.

```
>       scale_x, scale_y = (output_width / results.image_size[1], output_height / results.image_size[0])
E       RuntimeError: Integer division of tensors using div or / is no longer supported, and in a future release div will perform true division as in Python 3. Use true_divide or floor_divide (// in Python) instead.
```

cc @mruberry

mruberry:
> > Thanks @seemethere! In fact we expect to do true division there. When the deprecation happened, Mike already helped us migrate this part of code in [[[[[[[[[[[[[[D21997444](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/)](https://our.intern.facebook.com/intern/diff/D21997444/). Not sure why it still needs to be changed?
> 
> I believe the operator alone is not sufficient, and you might need to be more specific given the error message I observed in my testing.
> 
> ```
> >       scale_x, scale_y = (output_width / results.image_size[1], output_height / results.image_size[0])
> E       RuntimeError: Integer division of tensors using div or / is no longer supported, and in a future release div will perform true division as in Python 3. Use true_divide or floor_divide (// in Python) instead.
> ```
> 
> cc @mruberry
> 
> edit: I've updated my PR to use the `true_divide` function

This might happen if output_width and/or output_height are Python integers, which is not a case we encountered previously. New conditionals can be added to test for this and convert the scalars, if needed, or `torch.true_divide` can replace the division, instead.

ppwwyyxx:
Just noticed that when I run the test myself locally with 
```
python3 -m pytest -n 4 -v ./tests/test_model_analysis.py 
```
using pytorch from `pip install torch==1.6.0  -f https://download.pytorch.org/whl/test/cu101/torch_test.html  ` installed just now, the tests are actually passing and I did not observe the error in https://app.circleci.com/pipelines/github/pytorch/pytorch-integration-testing/28/workflows/80c9c991-949b-4928-a89b-a26c5f023f8f/jobs/149. Not sure what is causing the discrepancy.

ppwwyyxx:
We've added 1.6 to CI and tests are passing.

Issue #1763, opened 2020-07-13
Image Augmentation


manav-agarwal1:
So, I am trying to train an object detection model following the detectron2 tutorial and after getting a benchmark I have decided to do image data augmentation and I am stuck badly. As I have tried bunch of different things but I don't know how to get annotations out of them as annotations are available form previous data. I have made a dictionary of it as explained in tutorial and followed it. But now I am stuck at how to should I make changes. Whether I need to write the image in my training folder or how can I get annotations for them. then if I am doing random cropping then how can I obtain changed locations of boxes.
Please help I am just a beginner.

rahulg963:
@manav-agarwal1 can you please help me understand your query further?
Are you stuck, because you are not able to get annotations for your augmented images to be used for training/validation? or You are stuck in adding runtime augmentation while training your model using detectron2?

For the first problem, I you can use a library named [imgaug](https://github.com/aleju/imgaug). You will have to provide your initial images and annotations, this library will do the necessary augmentations and will give you corresponding annotations.

For the second you can try something like this. This will do pick images and do transformations and then use those transformations in training your model using detectron2.

```
def mapper(dataset_dict):
    # Implement a mapper, similar to the default DatasetMapper, but with your own customizations
    dataset_dict = copy.deepcopy(dataset_dict)  # it will be modified by code below
    image = utils.read_image(dataset_dict["file_name"], format="BGR")

    image, transforms = T.apply_transform_gens([
        T.RandomFlip(prob=0.20, horizontal=True, vertical=False),
        T.RandomFlip(prob=0.20, horizontal=False, vertical=True),
        T.RandomApply(transform=T.RandomBrightness(intensity_min=0.75, intensity_max=1.25),
                      prob=0.20),
        T.RandomApply(transform=T.RandomContrast(intensity_min=0.76, intensity_max=1.25),
                      prob=0.20),
        T.RandomApply(transform=T.RandomCrop(crop_type="relative_range", crop_size=(0.8, 0.8)), 
                      prob=0.20),
        T.RandomApply(transform=T.RandomSaturation(intensity_min=0.75, intensity_max=1.25), 
                      prob=0.20),
        T.RandomApply(transform=T.Resize(shape=(800, 800)), 
                      prob=0.20),
        T.RandomApply(transform=T.RandomRotation(angle=[-30,30], expand=True, center=None, sample_style="range", interp=None), 
                      prob=0.20)

    ], image)
    
[...]

Issue #1513, opened 2020-06-01
Make random flipping configurable


nikste:
## 🚀 Feature
it seems random flipping is hardcoded as a preprocessing step.
https://github.com/facebookresearch/detectron2/blob/9714aa4ab835dd64d45610e175464b93fb18fb43/detectron2/data/detection_utils.py#L516

This is not always a good idea (think traffic signs go left or go right which are symmetric but different).
Would be nice to have an easy way to disable this.

## Motivation & Examples
Symmetric but different objects (Traffic signs left, right arrow)

A config handle saying `RANDOM_FLIP` either `True` or `False`
(or even a list of preprocessing steps)

hiyyg:
Maybe `RANDOM_FLIP_TYPE` with options: `horizontal`, `vertical`, `none` is more suitable, since there users may also want to use vertical flip in some cases.

jorgeuliana1:
Is there any prediction for the solution of this issue? I believe it is pretty important.

ppwwyyxx:
closed by https://github.com/facebookresearch/detectron2/pull/1976

PR #1792, opened 2020-07-18
Make augmentation configurable

gen-ko:
Default behavior of the framework is not changed (i.e. only apply RandomFlip with prob 0.5)
Related issues:
[https://github.com/facebookresearch/detectron2/issues/1763](url)
[https://github.com/facebookresearch/detectron2/issues/1513](url)

How to use:
expand the list in config:
cfg.INPUT.AUGMENTATION = [
    {
        "name": "RandomFlip",
        "prob": 0.5,
        "horizontal": True,
        "vertical": False,
    }
]

and if user want to add new type of augmentations, use registry AUGMENTATION_REGISTRY (defined in detectron2/data/transforms/augmentation_impl.py) to register their customized augmentation.

initialization parameters are passed as kwargs except 'name'.

### DIFF
--- a/detectron2/config/defaults.py
+++ b/detectron2/config/defaults.py
@@ -71,7 +71,14 @@ _C.INPUT.CROP.TYPE = "relative_range"
 # Size of crop in range (0, 1] if CROP.TYPE is "relative" or "relative_range" and in number of
 # pixels if CROP.TYPE is "absolute"
 _C.INPUT.CROP.SIZE = [0.9, 0.9]
-
+_C.INPUT.AUGMENTATION = [
+    {
+        "name": "RandomFlip",
+        "prob": 0.5,
+        "horizontal": True,
+        "vertical": False,
+    }
+]
 
 # Whether the model needs RGB, YUV, HSV etc.
 # Should be one of the modes defined here, as we use PIL to read the image:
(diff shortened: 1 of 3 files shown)


gen-ko:
@ppwwyyxx any comments about this pr?

ppwwyyxx:
Augmentations of the default dataloader can be customized by just passing them as arguments, e.g. in https://github.com/facebookresearch/detectron2/blob/8cf31b794435c0065665462376e5a5bb487c4755/projects/DeepLab/train_net.py#L89

As https://detectron2.readthedocs.io/tutorials/configs.html#configs says, we do not expect everything to be available in config files, since config file is a limited language and cannot support any advanced logic. For this particular case of augmentation, writing the "augmentation config string" in this proposal has very similar complexity to the equivalent code, but more limited functionality than using code (e.g., it can only support kwargs; it can only support basic argument types like float, string), therefore I think this additional mechanism does not provide a lot of benefit over writing code directly.

Related to this, we're in the process of trying to make code more like configs, which should make more sense than making yaml config files more like code.

gen-ko:
> Augmentations of the default dataloader can be customized by just passing them as arguments, e.g. in
> 
> https://github.com/facebookresearch/detectron2/blob/8cf31b794435c0065665462376e5a5bb487c4755/projects/DeepLab/train_net.py#L89
> 
> As https://detectron2.readthedocs.io/tutorials/configs.html#configs says, we do not expect everything to be available in config files, since config file is a limited language and cannot support any advanced logic. For this particular case of augmentation, writing the "augmentation config string" in this proposal has very similar complexity to the equivalent code, but more limited functionality than using code (e.g., it can only support kwargs; it can only support basic argument types like float, string), therefore I think this additional mechanism does not provide a lot of benefit over writing code directly.
> 
> Related to this, we're in the process of trying to make code more like configs, which should make more sense than making yaml config files more like code.

Makes sense, thanks for your advice. PR closed.

PR #1804, opened 2020-07-21
fix for single classes (like balloons)

ianormy:
Thanks for your contribution!

If you're sending a large PR (e.g., >50 lines),

See https://detectron2.readthedocs.io/notes/contributing.html#pull-requests
about how we handle PRs.

### DIFF
--- a/detectron2/utils/visualizer.py
+++ b/detectron2/utils/visualizer.py
@@ -205,7 +205,7 @@ def _create_text_labels(classes, scores, class_names):
         list[str] or None
     """
     labels = None
-    if classes is not None and class_names is not None and len(class_names) > 1:
+    if classes is not None and class_names is not None and len(class_names) > 0:
         labels = [class_names[i] for i in classes]
     if scores is not None:
         if labels is None:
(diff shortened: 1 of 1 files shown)


ianormy:
When _create_text_labels is called for a single class, the **class** parameter will have something like this: **['balloon']**. With the existing code this will end up with a label such as: **100%**. With my proposed amendment this label will be correctly produced and say something like: **balloon 100%**. If you look at your example Detectron2 collab notebook you'll see the error with the inference visualization when it uses the balloon (single class) dataset. You don't see it with the coco dataset inference visualization at the beginning, because it has 80 classes.
IncorrectBalloon2

ppwwyyxx:
Thanks for sending PR. The current behavior is actually __intended__ and described in https://detectron2.readthedocs.io/modules/utils.html#detectron2.utils.visualizer.Visualizer.

However it has caused more than one confusions in the past and I agree that it is confusing. We should probably change it, in which case please update the documentation mentioned above as well.

ianormy:
@ppwwyyxx I have made the change to the Docstring for the Visualizer class. I am assuming that you have something that runs as part of your merge/build process that will call Sphinx to create the documentation you mentioned in your comment.

PR #1835, opened 2020-07-29
Make 'ROIAlign' & 'ROIAlignV2' version of ROIPooler scriptable.

chenbohua3:


detectron2/modeling/poolers.py
```diff
@@ -47,7 +60,14 @@ def assign_boxes_to_levels(
     return level_assignments.to(torch.int64) - min_level
 
 
-def convert_boxes_to_pooler_format(box_lists):
+def _fmt_box_list(box_tensor, batch_index: int):
```
chenbohua3, line 63: do we need move codes in this function to `convert_boxes_to_pooler_format`?

Issue #1712, opened 2020-07-04
Pre-Built Detectron2 For Windows


shlpu:
Are you willing to release Pre-Built Detectron2 For Windows? Thanks a lot.

ppwwyyxx:
We don't have such plans, but would appreciate it if anyone is willing to help us set up the CI on windows.

sarthakforwet:
Hey @shlpu nice thought! @ppwwyyxx I wanna work on this issue can you please tell me what I need to know related to the issue before start working on it. Thanks.

ppwwyyxx:
We use CircleCI at https://github.com/facebookresearch/detectron2/blob/master/.circleci/config.yml. It provides windows support so it can be configured with automated scripts to install dependencies, build detectron2 and run tests for every commit, on windows. https://github.com/pytorch/vision/blob/master/.circleci/config.yml can be used as a reference.

sarthakforwet:
@ppwwyyxx I am successful in installing dependencies with `windows executor` but quite unable to find out how should I install `detectron2` on windows.

Here's my sample config file (only for experiments not the complete ready code) - 
```
version: 2.1

windows-cpu: &windows_cpu
    machine:
      resource_class: windows.medium
      image: windows-server-2019-vs2019:stable
      shell: bash.exe

windows-gpu: &windows_gpu
    machine:
      resource_class: windows.gpu.nvidia.medium
      image: windows-server-2019-nvidia:stable
      shell: bash.exe

cpu: &cpu
  docker:
    - image: circleci/python:3.6.8-stretch
  resource_class: medium

gpu: &gpu
  machine:
    image: ubuntu-1604:201903-01
    docker_layer_caching: true
  resource_class: gpu.small

setup_venv: &setup_venv
  - run:
      name: Setup Virtual Env
      command: |
        python -m venv ~/venv
        echo ". ~/venv/bin/activate" >> $BASH_ENV
        . ~/venv/bin/activate
        python --version
        which python
        which pip
        pip install --upgrade pip

install_dep: &install_dep
  - run:
      name: Install Dependencies
      command: |
        python -m pip install --upgrade pip
        python --version
        python -m pip install 'git+https://github.com/facebookresearch/fvcore'
        python -m pip install opencv-python pytest-xdist
        pip install torch==1.6.0+cpu torchvision==0.7.0+cpu -f https://download.pytorch.org/whl/torch_stable.html

install_detectron2: &install_detectron2
  - run:
[...]

PR #1906, opened 2020-08-15
Added Windows build for detectron2

sarthakforwet:
PR from the outcome of the issue #1712 .

Updated:
- Added `<time.h>` header file in `cocoeval.cpp`.
- Changed `localtime_r` function to `localtime_s`.
- Updated config.yml file to add job for windows (cpu) build of detectron2.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,17 +200,16 @@ jobs:
       # Cache the env directory that contains dependencies
       - restore_cache:
           keys:
-            - cache-key-{{ .Branch }}-ID-20200730
+            - cache-key-{{ .Branch }}-ID-20200735
 
       - <<: *install_dep_win
 
       - save_cache:
           paths:
             - env
-          key: cache-key-{{ .Branch }}-ID-20200730
+          key: cache-key-{{ .Branch }}-ID-20200735
 
       - <<: *install_detectron2_win
-      - <<: *run_unittests
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,6 +72,7 @@ install_dep_win: &install_dep_win
   - run:
       name: Install Dependencies on Windows
       command: |
+        pip install numpy
         pip install cython
         pip install pycocotools==2.0.2a1
         pip install git+https://github.com/facebookresearch/fvcore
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -46,6 +68,16 @@ install_dep: &install_dep
         pip install --progress-bar off opencv-python pytest-xdist
         pip install --progress-bar off torch torchvision
 
+install_dep_win: &install_dep_win
+  - run:
+      name: Install Dependencies on Windows
+      command: |
+        pip install cython
```
ppwwyyxx, line 75: From the current failures in https://circleci.com/gh/facebookresearch/detectron2/1690?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link it looks like `numpy` has to be installed here as well. `fetch_build_eggs` for some reason does not work on windows

sarthakforwet:
image

image

This is the error I am getting for `FasterRCNNTest.test_flop` and `RetinaNetTest.test_flop`. 
@ppwwyyxx can you please tell me how can this error be solved? Why value of `res` is becoming zero?

ppwwyyxx:
If test failure is difficult to resolve now, it's fine to first add the CI with builds only but not unittests. At least it will make sure the code compiles on windows (and functional for the most parts)

sarthakforwet:
Due to my college semester, I was unable to lookup on the unittests issue but I think you are right about merging without unittests. However I would like to create an issue (on your call) about the failure of test cases. I would also keep working on that issue.

sarthakforwet:
Further Work: 

- [ ] Add Windows build for GPU
- [ ] Add Unittests to Windows Builds

ppwwyyxx:
After taking a closer look it turns out that the build actually did not succeed in CI. In "installing dependencies" it failed for pycocotools with the message:
> Installing collected packages: cycler, pillow, kiwisolver, python-dateutil, certifi, pyparsing, matplotlib, pycocotools
  Found existing installation: certifi 2019.3.9
Cannot uninstall 'certifi'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.

As a result it causes this in the end:
> detectron2._C          failed to import

which says detectron2 is not compiled

sarthakforwet:
I think its successfully built now @ppwwyyxx. I've added a force installation of the system to newest version of `certifi`.

ppwwyyxx:
nice fix! thanks!

sarthakforwet:
@ppwwyyxx Congratulations on 2.0.2 version of pycocotools. 💥💥

PR #1907, opened 2020-08-15
Update Readme.md

adithyaakrishna:
There was a extra / in line no. 33, removed that :)

Thanks for your contribution!


ppwwyyxx:
it's actually intended and means that "projects/" is a directory

Issue #1909, opened 2020-08-16
Add ignore regions to reduce effect from synonymous objects


Trouble404:
some datasets like CrowdHuman contains **'ignore regions'** bounding boxes annotation, there are 100k ignore bounding boxes in CrowdHuman, when we train a detector based on this dataset, we may want to reduce the effect from human doll, human billboard or dense crowd human. these part of regions should not consider as positive and negative supervise, bug I notice Detectron2 don't have a feature to process 'ignore' annotation during dataloader and matcher, this feature cound be used for lots of projects which need ignore labels
![table.png](https://vip1.loli.net/2020/08/16/UhjYCQ517DEqrm9.png)
![image.jpg](https://vip1.loli.net/2020/08/16/pAMiD9leV8td2xq.jpg)

ppwwyyxx:
It's unlikely we'll add to existing models large features that are specifically tailored to properties of certain datasets. Such features are expected to be implemented separately as new modules in users' code, or potentially in `detectron2/projects` if they appear more popular and widely useful.

If there are any good utilities (e.g., IOU vs IOA) or more general improvements that can simplify the above process or are part of the abovementioned feature, they can be discussed separately as individual feature requests.

ppwwyyxx:
I just noticed https://github.com/facebookresearch/detectron2/pull/1910 and it gave a better idea of what's needed to implement this. For reasons above (and more in [docs](https://detectron2.readthedocs.io/notes/contributing.html#pull-requests)) we may not be able to accept it in its current form, but some independent pieces of it can be polished and included. For example, if we for some reason need to implement this model, we'll take the following approach:
1. start by implementing a mechanism to pump the extra data into the model
    * allow `Intances` to store custom image-level data, if it's easier to use than storing "ignore" as a per-instance annotation.
    * propagate the "ignore" field from dataset to `Instances`

2. (in parallel with 1) implement necessary utility functions such as IoA (or "iof")
3. implement the ignore strategy for one model (e.g. RPN), and potentially more afterwards

Whether (3) can go into the existing models, or should be implemented separately outside detectron2 will depend on its complexity, which also relies on work done in (1) and (2). 
If you like to add them, they can be done in separate PRs (or issues first, if they involve significant changes). Implementing many orthogonal changes for a non-project directory in a single large commit is less effective and something we usually don't do.

Trouble404:
Thanks for your replying and suggestions, I will try to split those features and re-implement them.

[...]

PR #1910, opened 2020-08-16
add function to set certain area to ignore label for matcher from ann…

Trouble404:
#1909

For Add ignore regions to reduce effect from synonymous objects.
First we need pass 'ignore' label in dataloader and annotations_to_instances, like gt_classes, then after augmentation, we split gt_boxes to valid_boxes(origin gt boxes) and ignore_boxes, pass them to Matcher, at the end of matcher, calculate the IOU/IGNORE_BOX_AREA and set overlap > ingore_thresh matched label to -1, so we can ignore those ignore area in loss calculation.

For convenient, I change matcher's api, pass anchors, gt_boxes, ignore_boxes instead of iou match_quality_matrix, and calculate IOU inside matcher, we cound choose IOU calculate method by arg 'rotated' for  pairwise_iou_rotated. 

For calculating IOU/IGNORE_BOX_AREA, add a arg 'mode' is pairwise_iou, 'iof' will return IOU/IGNORE_BOX_AREA.

The ignore_boxes can obtain by get_process_match_data() in Instances Class

Corresponding change in one_stage_detector, two_stage_detector and rpn for using new matcher api , also update the test file of matcher

Add visualize funtion to show 'ignore' label in dataloader or annotation and their test file.

Add a config IGNORE_THRESH to control this feature, -1 means close this function, when thresh > 0 and has 'ignore' boxes, IOU/IGNORE_AREA > thresh's match label will set to -1

When use fastRCNN, the cls loss did't filter ignore obj in pred and target, also fix this issue
---

For test this funtion, I use 'iscrowd' label in coco as 'ignore' label, train model on retinanet_R_50_FPN_1x.yaml and faster_rcnn_R_50_FPN_1x.yaml

For using original detectron2 code, I got 37.28 AP in retina_1x, 37.32 on IGNORE_THRESH==0.7 in retina
* RETINA_BY_DEFAULT_MASTER_DETECTRON2

* RETINA_BY_USING_IGNORE_THRESH==0.7

For using original detectron2 code, I got 37.97 AP in faster_FPN_1X, 37.93 on IGNORE_THRESH==0.7 in RPN AND RPI_HEAD
* FPN_BY_DEFAULT_MASTER_DETECTRON2

* FPN_BY_USING_IGNORE_THRESH==0.7

* unit test passed


Trouble404:
origin test case got iou_matrix: 
([[1.0000, 0.5000, 0.5000, 0.2500, 0.2500, 0.1429, 0.2025],
  [0.2500, 0.5000, 0.5000, 1.0000, 0.0000, 0.0000, 0.1603]]
when apply iou_matrix.max(dim=0)
torch <= 1.5 got [0, 1, 1, 1, 0, 0]
torch == 1.6 and cpu got [0, 0, 0, 1, 0, 0, 0]
which means the boundary of bigger than some value is diff in different torch vision , but when assign label in matcher, this doesn't matter assign different label on those anchors, so I change the test case for pass unittest.

When using torch script, matcher won't use in inference, so temp remove torchscipt test on mather

Lint I use isort to auto sort the import order, but still got faied.

ppwwyyxx:
Sorry for the delay. Some feedbacks about this PR and #1920 after some discussions with the team: 

* We think that the extension of `Instances` is not the right way to pass extra data around. `Instances` is currently designed to store attributes for every instance, and storing arbitrary extra data inside will break this abstraction. Sorry that I may have misled you to this direction, but we think it's still better to pass non-instance data around using an extra argument (of a new model), rather than trying to "shoehorn" this into the existing `Instances` class.
* We generally try not to break APIs, at least not without careful consideration. So breaking changes to matcher and removing unittests should ideally be avoided. 
* Related to the points above, new features can always be added in two ways
  * adding/changing/(possibly breaking) arguments of existing functions/classes
  * write new functions/classes but with some code duplications or (later try to) share code with existing ones. 

  Given the research nature we usually prefer the later approach, especially for features like this which are useful at some situation, but not essential to everyone. The later approach also makes it possible to implement such features under `projects/` so we can manage the relevant pieces together.
* We're still happy to include such a feature under `projects/` (maybe some part of it under `detectron2/`). If you're still interested, let me know and I can start discussing what approaches to take before jumping into code.

Trouble404:
Thanks for your replay, I am willing to cooperate with you to finish this work.

right now all data will get augmentation and filter in annotations_to_instances(), extra data inside Instance, 
should we split them into 2 Instances and controlled by an argument? we could discuss the approaches.

PR #1921, opened 2020-08-18
implement pairwise_ioa by pairwise_iou and add corresponding test case

Trouble404:
ref to issues in #1909 and pr in #1910 

The matcher ignore area strategy split into 3 parts:

1. pump the extra data into the model
2. **utility functions IoA**
3. ignore strategy for RPN

here is the second implementation, 

I extract intersection calculation from pairwise_iou function,  
and implement pairwise_ioa based on this, also add relevant testcase for IoA.

### DIFF
--- a/detectron2/structures/boxes.py
+++ b/detectron2/structures/boxes.py
@@ -311,22 +311,18 @@ class Boxes:
         yield from self.tensor
 
 
-def intersection_cacl(
-    boxes1: Boxes, boxes2: Boxes
-) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+def pairwise_intersection(boxes1: Boxes, boxes2: Boxes) -> torch.Tensor:
     """
     Given two lists of boxes of size N and M,
-    compute their independent boxes areas and intersection areas.
+    compute the intersection area between __all__ N x M pairs of boxes.
+    The box order must be (xmin, ymin, xmax, ymax)
+
     Args:
         boxes1,boxes2 (Boxes): two `Boxes`. Contains N & M boxes, respectively.
 
     Returns:
         Tensor: intersection, sized [N,M].
-        Tensor: area1, area2, sized N and M
     """
-    area1 = boxes1.area()  # [N]
-    area2 = boxes2.area()  # [M]
-
     boxes1, boxes2 = boxes1.tensor, boxes2.tensor
     width_height = torch.min(boxes1[:, None, 2:], boxes2[:, 2:]) - torch.max(
         boxes1[:, None, :2], boxes2[:, :2]
@@ -334,8 +330,7 @@ def intersection_cacl(
 
     width_height.clamp_(min=0)  # [N,M,2]
     intersection = width_height.prod(dim=2)  # [N,M]
-    del width_height
-    return intersection, area1, area2
+    return intersection
 
 
 # implementation from https://github.com/kuangliu/torchcv/blob/master/torchcv/utils/box.py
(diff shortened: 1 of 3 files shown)


detectron2/structures/boxes.py
```diff
@@ -311,33 +311,49 @@ def __iter__(self):
         yield from self.tensor
 
 
-# implementation from https://github.com/kuangliu/torchcv/blob/master/torchcv/utils/box.py
-# with slight modifications
-def pairwise_iou(boxes1: Boxes, boxes2: Boxes) -> torch.Tensor:
+def intersection_cacl(
```
ppwwyyxx, line 314: name it "pairwise_intersection" to be consistent with the other two

detectron2/structures/boxes.py
```diff
@@ -311,33 +311,49 @@ def __iter__(self):
...
-    The box order must be (xmin, ymin, xmax, ymax).
-
+    compute their independent boxes areas and intersection areas.
     Args:
         boxes1,boxes2 (Boxes): two `Boxes`. Contains N & M boxes, respectively.
 
     Returns:
-        Tensor: IoU, sized [N,M].
+        Tensor: intersection, sized [N,M].
+        Tensor: area1, area2, sized N and M
```
ppwwyyxx, line 325: Computing both of them is unnecessary for `pairwise_ioa`.
areas can be computed in `pairwise_iou/ioa` when needed and this function will compute intersection only. This also simplifies the interface of this function.

detectron2/structures/boxes.py
```diff
@@ -311,33 +311,49 @@ def __iter__(self):
...
     boxes1, boxes2 = boxes1.tensor, boxes2.tensor
-
     width_height = torch.min(boxes1[:, None, 2:], boxes2[:, 2:]) - torch.max(
         boxes1[:, None, :2], boxes2[:, :2]
     )  # [N,M,2]
 
     width_height.clamp_(min=0)  # [N,M,2]
-    inter = width_height.prod(dim=2)  # [N,M]
+    intersection = width_height.prod(dim=2)  # [N,M]
     del width_height
```
ppwwyyxx, line 337: this line can be removed because the function ends here

detectron2/structures/boxes.py
```diff
@@ -311,33 +311,49 @@ def __iter__(self):
...
+def intersection_cacl(
+    boxes1: Boxes, boxes2: Boxes
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
     """
     Given two lists of boxes of size N and M,
-    compute the IoU (intersection over union)
-    between __all__ N x M pairs of boxes.
-    The box order must be (xmin, ymin, xmax, ymax).
-
+    compute their independent boxes areas and intersection areas.
```
ppwwyyxx, line 319:
```
compute the intersection area between __all__ N x M pairs of boxes.	
The box order must be (xmin, ymin, xmax, ymax)
```

Trouble404:
I update the pairwise intersection, ioa and iou as your comment .
I only modify the boxes.py but the lint check of other two files reformatted this time.

detectron2/structures/boxes.py
```diff
@@ -311,6 +311,29 @@ def __iter__(self):
         yield from self.tensor
 
 
+def pairwise_intersection(
+    boxes1: Boxes, boxes2: Boxes
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
```
ppwwyyxx, line 316: return type is incorrect

detectron2/structures/boxes.py
```diff
@@ -311,6 +311,29 @@ def __iter__(self):
...
+    """
+    Given two lists of boxes of size N and M,
+    compute the intersection area between __all__ N x M pairs of boxes.
+    The box order must be (xmin, ymin, xmax, ymax)
+    Args:
```
ppwwyyxx, line 321: We follow the [google style](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) and a new line is required before `Args:` here so it can be parsed properly

PR #1976, opened 2020-08-28
Adds config options for randomly flipping images  

kondela:
Adds options to config in order to enable/disable random horizontal/vertical flipping for image augmentation.

Feature labeled as _enhancement_ and requested by [1513](https://github.com/facebookresearch/detectron2/issues/1513)

### DIFF
--- a/detectron2/config/defaults.py
+++ b/detectron2/config/defaults.py
@@ -58,9 +58,7 @@ _C.INPUT.MIN_SIZE_TEST = 800
 # Maximum size of the side of the image during testing
 _C.INPUT.MAX_SIZE_TEST = 1333
 # `True` if flipping is used for data augmentation during training
-_C.INPUT.RANDOM_FLIP = CN({"ENABLED": True})
-_C.INPUT.RANDOM_FLIP.HORIZONTAL = True
-_C.INPUT.RANDOM_FLIP.VERTICAL = False
+_C.INPUT.RANDOM_FLIP = "horizontal/vertical/none"
 
 # `True` if cropping is used for data augmentation during training
 _C.INPUT.CROP = CN({"ENABLED": False})
--- a/detectron2/config/defaults.py
+++ b/detectron2/config/defaults.py
@@ -57,8 +57,9 @@ _C.INPUT.MAX_SIZE_TRAIN = 1333
 _C.INPUT.MIN_SIZE_TEST = 800
 # Maximum size of the side of the image during testing
 _C.INPUT.MAX_SIZE_TEST = 1333
-# `True` if flipping is used for data augmentation during training
-_C.INPUT.RANDOM_FLIP = "horizontal/vertical/none"
+# Mode for flipping images used in data augmentation during training
+# choose one of ["horizontal, "vertical", "none"]
+_C.INPUT.RANDOM_FLIP = "horizontal"
 
 # `True` if cropping is used for data augmentation during training
 _C.INPUT.CROP = CN({"ENABLED": False})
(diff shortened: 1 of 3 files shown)


kondela:
Pipeline failed on `test_nms_rotated_0_degree_cuda` located in `tests/layers/test_nms_rotated.py`. However, when I ran it locally it passed all the tests. After a closer inspection it appears that the `_create_tensors` method is non deterministic as it contains `torch.rand` which once in a while produces values that make the test fail. Confirmed after running the test 200 times it failed once. 

I guess this would deserve a new PR that would make test behavior deterministic. Meanwhile, I propose to simply rerun the failed pipeline, what do you think @ppwwyyxx?

detectron2/config/defaults.py
```diff
@@ -57,6 +57,10 @@
 _C.INPUT.MIN_SIZE_TEST = 800
 # Maximum size of the side of the image during testing
 _C.INPUT.MAX_SIZE_TEST = 1333
+# `True` if flipping is used for data augmentation during training
+_C.INPUT.RANDOM_FLIP = CN({"ENABLED": True})
+_C.INPUT.RANDOM_FLIP.HORIZONTAL = True
+_C.INPUT.RANDOM_FLIP.VERTICAL = False
```
ppwwyyxx, line 63: could you use `INPUT.RANDOM_FLIP="horizontal/vertical/none"`?

ppwwyyxx (commented):
Yeah this test occasionally fails. Instead of making it deterministic we would actually like to understand the root cause of that and whether there is a small bug behind. For now we can just ignore that in this PR

detectron2/config/defaults.py
```diff
@@ -58,9 +58,7 @@
 # Maximum size of the side of the image during testing
 _C.INPUT.MAX_SIZE_TEST = 1333
 # `True` if flipping is used for data augmentation during training
-_C.INPUT.RANDOM_FLIP = CN({"ENABLED": True})
-_C.INPUT.RANDOM_FLIP.HORIZONTAL = True
-_C.INPUT.RANDOM_FLIP.VERTICAL = False
+_C.INPUT.RANDOM_FLIP = "horizontal/vertical/none"
```
ppwwyyxx, line 61: default should be "horizontal"; comment should explain possible values

detectron2/data/detection_utils.py
```diff
@@ -579,8 +579,13 @@ def build_augmentation(cfg, is_train):
         max_size = cfg.INPUT.MAX_SIZE_TEST
         sample_style = "choice"
     augmentation = [T.ResizeShortestEdge(min_size, max_size, sample_style)]
-    if is_train:
-        augmentation.append(T.RandomFlip())
+    if is_train and "none" not in cfg.INPUT.RANDOM_FLIP:
```
ppwwyyxx, line 582:
```suggestion
    if is_train and cfg.INPUT.RANDOM_FLIP != "none":
```

detectron2/data/detection_utils.py
```diff
@@ -579,8 +579,13 @@ def build_augmentation(cfg, is_train):
...
+        augmentation.append(
+            T.RandomFlip(
+                horizontal=True if "horizontal" in cfg.INPUT.RANDOM_FLIP else False,
+                vertical=True if "vertical" in cfg.INPUT.RANDOM_FLIP else False,
```
ppwwyyxx, line 586:
```
horizontal=cfg.INPUT.RANDOM_FLIP == "horizontal",
vertical=...
```

detectron2/data/detection_utils.py
kondela: Oops, that was left by accident from the previous commit. Should be correct now. :)

detectron2/data/detection_utils.py
kondela: Same as https://github.com/facebookresearch/detectron2/pull/1976#discussion_r483426629 - accidentally left from the previous commit, should be correct now.

PR #2003, opened 2020-09-04
Make mask_head scriptable.

chenbohua3:

### DIFF
--- a/detectron2/modeling/roi_heads/mask_head.py
+++ b/detectron2/modeling/roi_heads/mask_head.py
@@ -1,5 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
-from typing import List
+from typing import Any, List
 import fvcore.nn.weight_init as weight_init
 import torch
 from torch import nn
@@ -27,8 +27,15 @@ per-region features.
 The registered object will be called with `obj(cfg, input_shape)`.
 """
 
+# To get torchscript support, we make mask head class a subclass of `nn.Sequential`.
+# If you want to add new layers in these head classes, please make sure they are added to the class
+# in the order of that in the network.
+# todo: revert to nn.Module when torchscript supports List[nn.Module]
+# https://github.com/pytorch/pytorch/pull/38059
 
-def mask_rcnn_loss(pred_mask_logits, instances, vis_period=0):
+
+@torch.jit.unused
+def mask_rcnn_loss(pred_mask_logits: torch.Tensor, instances: List[Instances], vis_period: int = 0):
     """
     Compute the mask prediction loss defined in the Mask R-CNN paper.
 
@@ -110,7 +117,7 @@ def mask_rcnn_loss(pred_mask_logits, instances, vis_period=0):
     return mask_loss
 
 
-def mask_rcnn_inference(pred_mask_logits, pred_instances):
+def mask_rcnn_inference(pred_mask_logits: torch.Tensor, pred_instances: List[Instances]):
     """
     Convert pred_mask_logits to estimated foreground probability masks while also
     extracting only the masks for the predicted classes in pred_instances. For each
@@ -151,7 +158,7 @@ def mask_rcnn_inference(pred_mask_logits, pred_instances):
         instances.pred_masks = prob  # (1, Hmask, Wmask)
 
 
-class BaseMaskRCNNHead(nn.Module):
+class BaseMaskRCNNHead(nn.Sequential):
     """
     Implement the basic Mask R-CNN losses and inference logic described in :paper:`Mask R-CNN`
     """
@@ -171,7 +178,7 @@ class BaseMaskRCNNHead(nn.Module):
     def from_config(cls, cfg, input_shape):
         return {"vis_period": cfg.VIS_PERIOD}
 
-    def forward(self, x, instances: List[Instances]):
+    def forward(self, x, instances: List[Instances]) -> Any:
         """
         Args:
             x: input region feature(s) provided by :class:`ROIHeads`.
@@ -234,7 +241,7 @@ class MaskRCNNConvUpsampleHead(BaseMaskRCNNHead):
                 padding=1,
                 bias=not conv_norm,
                 norm=get_norm(conv_norm, conv_dim),
-                activation=F.relu,
+                activation=nn.ReLU(),
             )
             self.add_module("mask_fcn{}".format(k + 1), conv)
             self.conv_norm_relus.append(conv)
@@ -243,6 +250,7 @@ class MaskRCNNConvUpsampleHead(BaseMaskRCNNHead):
         self.deconv = ConvTranspose2d(
             cur_channels, conv_dims[-1], kernel_size=2, stride=2, padding=0
         )
+        self.add_module("deconv_relu", nn.ReLU())
         cur_channels = conv_dims[-1]
 
         self.predictor = Conv2d(cur_channels, num_classes, kernel_size=1, stride=1, padding=0)
@@ -271,10 +279,9 @@ class MaskRCNNConvUpsampleHead(BaseMaskRCNNHead):
         return ret
 
     def layers(self, x):
-        for layer in self.conv_norm_relus:
+        for layer in self:
             x = layer(x)
-        x = F.relu(self.deconv(x))
-        return self.predictor(x)
+        return x
 
 
 def build_mask_head(cfg, input_shape):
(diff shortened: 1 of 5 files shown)


tests/modeling/test_roi_heads.py
```diff
@@ -131,6 +133,41 @@ def test_rroi_heads(self):
...
+        pred_instance1.pred_classes = pred_classes1
+
+        mask_head = build_mask_head(cfg, input_shape).eval()
+        origin_outputs = mask_head(mask_features, [pred_instance0, pred_instance1])
+
+        fields = {"pred_masks": "Tensor", "pred_classes": "Tensor"}
+        with patch_instances(fields) as new_instance:
+            sciript_mask_head = torch.jit.script(mask_head)
+            pred_instance0 = new_instance(image_shapes[0])
+            pred_instance0.pred_classes = pred_classes0
```
ppwwyyxx, line 161: I think adding a conversion staticmethod `NewInstances.from_instances(instances: Instances) -> NewInstances` would be useful. This method is `@jit.unused` and is only meant for users to construct inputs for modules that need `NewInstances` as input.

detectron2/modeling/roi_heads/mask_head.py
```diff
@@ -234,7 +241,7 @@ def __init__(self, input_shape: ShapeSpec, *, num_classes, conv_dims, conv_norm=
                 padding=1,
                 bias=not conv_norm,
                 norm=get_norm(conv_norm, conv_dim),
-                activation=F.relu,
+                activation=nn.ReLU(),
```
ppwwyyxx, line 244: this change is not necessary

detectron2/modeling/roi_heads/mask_head.py
```diff
@@ -151,7 +158,7 @@ def mask_rcnn_inference(pred_mask_logits, pred_instances):
         instances.pred_masks = prob  # (1, Hmask, Wmask)
 
 
-class BaseMaskRCNNHead(nn.Module):
+class BaseMaskRCNNHead(nn.Sequential):
```
ppwwyyxx, line 161: This might have unexpected side effect for other subclasses. It's safer to limit this change to `MaskRCNNConvUpsampleHead` for now.

detectron2/modeling/roi_heads/mask_head.py
chenbohua3: if this change is not conducted, Runtime Error will be raised when call `torch.jit.script(model)`:
```
RuntimeError: 
function cannot be used as a value:
  File "detectron2/layers/wrappers.py", line 107
        if self.norm is not None:
            x = self.norm(x)
        if self.activation is not None:
           ~~~~~~~~~~~~~~~ <--- HERE
            x = self.activation(x)
        return x
```
if there better solution to handle this?

detectron2/modeling/roi_heads/mask_head.py
chenbohua3: The `forward` function of `MaskRCNNConvUpsampleHead` is defined in `BaseMaskRCNNHead `, you mean make a copy of the `forward` function of `BaseMaskRCNNHead ` and add it to `MaskRCNNConvUpsampleHead`?

tests/modeling/test_roi_heads.py
```diff
@@ -131,6 +133,41 @@ def test_rroi_heads(self):
             ),
         )
 
+    @unittest.skipIf(TORCH_VERSION < (1, 7), "Insufficient pytorch version")
+    def test_mask_head_scriptability(self):
+        cfg = get_cfg()
+        cfg.MODEL.ROI_MASK_HEAD.NUM_CONV = 1
+        cfg.MODEL.MASK_ON = True
+
+        input_shape = ShapeSpec(channels=1024)
```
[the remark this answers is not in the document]
chenbohua3: Your solution is better, i will make it more straightforward:)

detectron2/modeling/roi_heads/mask_head.py
ppwwyyxx: ok, let's keep it like this

detectron2/modeling/roi_heads/mask_head.py
ppwwyyxx: ···
class MaskRCNNConvUpsampleHead(BaseMaskRCNNHead, nn.Sequential):
```
seems to work

File: tests/modeling/test_roi_heads.py (hunk 1 above)
Comment (chenbohua3) [reply tests/modeling/test_roi_heads.py]: done

File: detectron2/modeling/roi_heads/mask_head.py (hunk 3 above)
Comment (chenbohua3) [reply detectron2/modeling/roi_heads/mask_head.py]: done:)

File: tests/modeling/test_roi_heads.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (chenbohua3) [reply tests/modeling/test_roi_heads.py:146]: done

File: tests/modeling/test_roi_heads.py [hunk 5]
```diff
@@ -131,6 +133,41 @@ def test_rroi_heads(self):
...
+        pred_instance0.pred_classes = pred_classes0
+        pred_instance1 = Instances(image_shapes[1])
+        pred_classes1 = torch.tensor([4], dtype=torch.int64)
+        pred_instance1.pred_classes = pred_classes1
+
+        mask_head = build_mask_head(cfg, input_shape).eval()
+        origin_outputs = mask_head(mask_features, [pred_instance0, pred_instance1])
+
+        fields = {"pred_masks": "Tensor", "pred_classes": "Tensor"}
+        with patch_instances(fields) as new_instance:
```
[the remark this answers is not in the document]
chenbohua3: done

tests/modeling/test_roi_heads.py
```diff
@@ -131,6 +137,41 @@ def test_rroi_heads(self):
             ),
         )
 
+    @unittest.skipIf(TORCH_VERSION < (1, 7), "Insufficient pytorch version")
+    def test_mask_head_scriptability(self):
+        cfg = get_cfg()
+        cfg.MODEL.ROI_MASK_HEAD.NUM_CONV = 1
+        cfg.MODEL.MASK_ON = True
```
ppwwyyxx, line 144: cfg can be removed

detectron2/modeling/roi_heads/mask_head.py
```diff
@@ -27,8 +27,15 @@
 The registered object will be called with `obj(cfg, input_shape)`.
 """
 
+# To get torchscript support, we make mask head class `MaskRCNNConvUpsampleHead` a subclass
+# of `nn.Sequential`. If you want to add new layers in this head class, please make sure
+# they are added to the class in the order of that in the network.
+# todo: revert to nn.Module when torchscript supports List[nn.Module]
+# https://github.com/pytorch/pytorch/pull/38059
```
ppwwyyxx, line 34: Let's link to the issue instead. It's not certain yet whether the issue will be addressed by this PR or others

ppwwyyxx: In fact, after a second thought the changes in this class are still quite straightforward and don't have to be reverted. So I would suggest just removing these two lines of comments.

tests/modeling/test_roi_heads.py
chenbohua3: removed

detectron2/modeling/roi_heads/mask_head.py
chenbohua3: for `mask_head`, `nn.Sequential` is indeed more straightforward since the network structure is 'sequential'. I deleted these two lines of comments.

tests/modeling/test_roi_heads.py
```diff
@@ -131,6 +137,36 @@ def test_rroi_heads(self):
...
+        pred_classes0 = torch.tensor([1, 2, 3], dtype=torch.int64)
+        pred_instance0.pred_classes = pred_classes0
+        pred_instance1 = Instances(image_shapes[1])
+        pred_classes1 = torch.tensor([4], dtype=torch.int64)
+        pred_instance1.pred_classes = pred_classes1
+
+        mask_head = MaskRCNNConvUpsampleHead(
+            input_shape, num_classes=80, conv_dims=[256, 256]
+        ).eval()
+        origin_outputs = mask_head(mask_features, [pred_instance0, pred_instance1])
```
chenbohua3, line 156: Do we need to remove `pred_masks`? It is in-place added in `MaskRCNNConvUpsampleHead`. Or we can use `deepcopy` to avoid it.

ppwwyyxx: good point! deepcopy of the input would be good here

chenbohua3: done, I also use `deepcopy` in `from_instances` to avoid unexpected changes for origin instance caused by new instance.

PR #2036, opened 2020-09-16
Make keypoint head scriptable.

chenbohua3:


detectron2/modeling/roi_heads/keypoint_head.py
```diff
@@ -120,7 +120,11 @@ def keypoint_rcnn_inference(pred_keypoint_logits, pred_instances):
     pred_keypoint_logits = pred_keypoint_logits.detach()
     keypoint_results = heatmaps_to_keypoints(pred_keypoint_logits, bboxes_flat.detach())
     num_instances_per_image = [len(i) for i in pred_instances]
-    keypoint_results = keypoint_results[:, :, [0, 1, 3]].split(num_instances_per_image, dim=0)
+    # slicing multiple dimensions with sequences is not supported by torchscript
+    # https://github.com/pytorch/pytorch/issues/43943
+    keypoint_results = keypoint_results.index_select(dim=2, index=torch.tensor([0, 1, 3])).split(
+        num_instances_per_image, dim=0
+    )
```
ppwwyyxx, line 127:
```
import torch

def func(x: torch.Tensor):
    idx = torch.tensor([0, 1, 3])
    return x[:, :, idx]

f = torch.jit.script(func)
```
the above works in my local test. Could you try this? It's a smaller change from the original

detectron2/structures/keypoints.py
```diff
@@ -159,6 +158,12 @@ def heatmaps_to_keypoints(maps: torch.Tensor, rois: torch.Tensor) -> torch.Tenso
     we maintain consistency with :meth:`Keypoints.to_heatmap` by using the conversion from
     Heckbert 1990: c = d + 0.5, where d is a discrete coordinate and c is a continuous coordinate.
     """
+    # functions decorated with @torch.no_grad() can not be exported to torchscript when
+    # other functions are called inside them.
```
ppwwyyxx, line 162: Nothing needed to be done for this PR, but just curious:
The decorator use was not supported according to this PR: https://github.com/pytorch/pytorch/pull/41371
Did you mean that decorator it's actually scriptable if the function does not call other functions inside? If that's true I wonder whether `no_grad` is effective or is simply not parsed by jit -- the latter case should be considered a bug.

detectron2/modeling/roi_heads/keypoint_head.py
chenbohua3: This works well:)

detectron2/structures/keypoints.py
chenbohua3: Yes, if no other functions are called inside, the target function can be exported successfully. (Not sure if this applies to all functions in that situation, I just test several simple functions).

For now I have no ideas about this phenomenon, I can elaborate this in https://github.com/pytorch/pytorch/issues/44768

chenbohua3: This seems to be a bug of JIT, I will report it in that issue.

PR #2060, opened 2020-09-22
Improve ONNX tracing support for Faster-RCNN

crisp-snakey:
Small fixes to better preserve the origin of shape information during ONNX export. This resolves an issue where an ONNX export would generate fixed sized tensors for tensors that are actually of a dynamic shape.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -58,9 +58,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          brew install libomp
           python -m pip install -U pip setuptools==49
-          python -m pip install ninja opencv-python-headless onnx pytest-xdist onnxruntime
+          python -m pip install ninja opencv-python-headless onnx pytest-xdist
           python -m pip install torch==${{matrix.torch}} torchvision==${{matrix.torchvision}} -f https://download.pytorch.org/whl/torch_stable.html
           python -m pip install -U 'git+https://github.com/facebookresearch/fvcore'
 
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install flake8==3.8.1 flake8-bugbear flake8-comprehensions isort==4.3.21 onnxruntime
+          python -m pip install flake8==3.8.1 flake8-bugbear flake8-comprehensions isort==4.3.21
           python -m pip install 'black @ git+https://github.com/psf/black@673327449f86fce558adde153bb6cbe54bfebad2'
           flake8 --version
       - name: Lint
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install flake8==3.8.1 flake8-bugbear flake8-comprehensions isort==4.3.21
+          python -m pip install flake8==3.8.1 flake8-bugbear flake8-comprehensions isort==4.3.21 onnxruntime
           python -m pip install 'black @ git+https://github.com/psf/black@673327449f86fce558adde153bb6cbe54bfebad2'
           flake8 --version
       - name: Lint
@@ -59,7 +59,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install -U pip setuptools==49
-          python -m pip install ninja opencv-python-headless onnx pytest-xdist
+          python -m pip install ninja opencv-python-headless onnx pytest-xdist onnxruntime
           python -m pip install torch==${{matrix.torch}} torchvision==${{matrix.torchvision}} -f https://download.pytorch.org/whl/torch_stable.html
           python -m pip install -U 'git+https://github.com/facebookresearch/fvcore'
 
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -58,6 +58,7 @@ jobs:
 
       - name: Install dependencies
         run: |
+          brew install libomp
           python -m pip install -U pip setuptools==49
           python -m pip install ninja opencv-python-headless onnx pytest-xdist onnxruntime
           python -m pip install torch==${{matrix.torch}} torchvision==${{matrix.torchvision}} -f https://download.pytorch.org/whl/torch_stable.html
--- a/detectron2/modeling/box_regression.py
+++ b/detectron2/modeling/box_regression.py
@@ -102,12 +102,16 @@ class Box2BoxTransform(object):
         pred_w = torch.exp(dw) * widths[:, None]
         pred_h = torch.exp(dh) * heights[:, None]
 
-        pred_boxes = torch.zeros_like(deltas)
-        pred_boxes[:, 0::4] = pred_ctr_x - 0.5 * pred_w  # x1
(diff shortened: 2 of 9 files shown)


ppwwyyxx:
Could you add a unittest that would have failed before? Otherwise it's not clear to me what this PR is trying to fix.

crisp-snakey:
So, I've been trying to add a unittest that shows the problem (at least for the ROIPooler). It requires an extra dependency (onnxruntime) which seems to be broken on MacOS. The pipeline fails during an import inside of onnxruntime (it for some reason can't import a function to retrieve the available backends).

detectron2/modeling/poolers.py
```diff
@@ -59,8 +59,12 @@ def assign_boxes_to_levels(
...
 
 def _fmt_box_list(box_tensor, batch_index: int):
-    repeated_index = torch.full(
-        (len(box_tensor), 1), batch_index, dtype=box_tensor.dtype, device=box_tensor.device
+    repeated_index = torch.full_like(
+        box_tensor[:, :1],
+        batch_index,
+        dtype=box_tensor.dtype,
+        device=box_tensor.device,
+        layout=torch.strided,
```
ppwwyyxx, line 67: layout= is probably not needed

.github/workflows/workflow.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install flake8==3.8.1 flake8-bugbear flake8-comprehensions isort==4.3.21
+          python -m pip install flake8==3.8.1 flake8-bugbear flake8-comprehensions isort==4.3.21 onnxruntime
```
ppwwyyxx, line 20: this test is only about linter and does not require this dependency

detectron2/modeling/poolers.py
```diff
@@ -232,7 +236,7 @@ def forward(self, x: List[torch.Tensor], box_lists: List[Boxes]):
             box_lists, self.min_level, self.max_level, self.canonical_box_size, self.canonical_level
         )
 
-        num_boxes = len(pooler_fmt_boxes)
+        num_boxes = pooler_fmt_boxes.size(0)
```
ppwwyyxx, line 239: When I run it, the tests can pass without this change.

detectron2/modeling/roi_heads/fast_rcnn.py
```diff
@@ -547,7 +547,7 @@ def predict_boxes(self, predictions, proposals):
         if not len(proposals):
             return []
         _, proposal_deltas = predictions
-        num_prop_per_image = [len(p) for p in proposals]
+        num_prop_per_image = [p.proposal_boxes.tensor.size(0) for p in proposals]
```
ppwwyyxx, line 550: Changes in this file are not tested. Are they still needed?

tests/modeling/test_roi_pooler.py
```diff
@@ -134,6 +137,84 @@ def test_no_images(self):
...
+                dynamic_axes={"boxes": {0: "detections"}, "pooled_features": {0: "detections"}},
+                opset_version=11,
+            )
+
+            f.seek(0)
+
+            sess = rt.InferenceSession(f.read(), None)
+
+            sess.run([], {"features": feature.numpy(), "boxes": rois.numpy()})
+            sess.run([], {"features": feature.numpy(), "boxes": rois[:5].numpy()})
```
ppwwyyxx, line 216: Thanks for the thorough tests. Could you assert that the return value has the expected dynamic shape?

detectron2/modeling/poolers.py
crisp-snakey: Indeed it probably isn't. I don't know the probability that a sparse tensor of bounding boxes can be passed in. If this is zero the `layout=torch.strided` is not necessary. I don't think having it there will hurt either.

ppwwyyxx: `box_tensor[:, :1]` will fail on sparse tensor, if it ever appears

crisp-snakey: You're right. I'll remove it then.

.github/workflows/workflow.yml
crisp-snakey: It seems to move the import around depending on whether or not the package is installed. If the package is not installed it's moved below the detectron2 imports. If it is installed then it's moved above the detectron2 imports.

detectron2/modeling/roi_heads/fast_rcnn.py
crisp-snakey: I've now added tests for them.

.github/workflows/workflow.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install -U pip setuptools==49
-          python -m pip install ninja opencv-python-headless onnx pytest-xdist
+          python -m pip install ninja opencv-python-headless onnx pytest-xdist onnxruntime
```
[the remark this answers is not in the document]
crisp-snakey: Thank you that indeed solved the problem.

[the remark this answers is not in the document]
crisp-snakey: Despite that there still seems to be a problem on MacOS. I can't test it myself and CI output is not that helpful.

tests/modeling/test_roi_pooler.py
```diff
@@ -134,6 +138,89 @@ def test_no_images(self):
...
+
+            sess = rt.InferenceSession(f.read(), None)
+
+            assert (
+                sess.run([], {"features": feature.numpy(), "boxes": rois.numpy()})[0].shape[0] == 10
+            )
+            assert (
+                sess.run([], {"features": feature.numpy(), "boxes": rois[:5].numpy()})[0].shape[0]
+                == 5
+            )
```
ppwwyyxx, line 222:
```suggestion
        func = torch.jit.trace(model, (feature, rois))
        o = func(feature, rois)
        self.assertEqual(o.shape, (10, 4, 14, 14))
        o = func(feature, rois[:5])
        self.assertEqual(o.shape, (5, 4, 14, 14))

```

I think this change will test just the same feature you're trying to support. 
Please don't rely on onnxruntime in this PR then. It isn't very friendly to maintainers given it now requires libomp. (also we in general can't provide much support for third-party runtime)

crisp-snakey: Thank you for the suggestion. That works. Though the scripting does seem to run into [this bug](https://github.com/pytorch/pytorch/pull/44335) while applying the box deltas in the box regressor.

detectron2/modeling/box_regression.py
```diff
@@ -102,12 +102,15 @@ def apply_deltas(self, deltas, boxes):
...
-        return pred_boxes
+        return torch.cat(
+            [
+                (pred_ctr_x - 0.5 * pred_w).unsqueeze(dim=-1),  # x1
+                (pred_ctr_y - 0.5 * pred_h).unsqueeze(dim=-1),  # y1
+                (pred_ctr_x + 0.5 * pred_w).unsqueeze(dim=-1),  # x2
+                (pred_ctr_y + 0.5 * pred_h).unsqueeze(dim=-1),  # y2
+            ],
+            dim=-1,
+        ).flatten(1)
```
ppwwyyxx, line 113: fyi, I'm only waiting to first get the fix https://github.com/pytorch/pytorch/pull/45828 into pytorch so we don't have to do this hack

ppwwyyxx:
FYI we also hide the changes to `len()` under a patch function. In general we avoid hacks on modeling code if the same effect can be achieved by more hacks outside the modeling code.

PR #2095, opened 2020-10-03
UpdateReadme.md

gujralsanyam22:
see this errors i think that it can be fixed it because it looks not so good that's why i fix it.


ppwwyyxx:
The PR contains no meaningful changes.

PR #2097, opened 2020-10-03
Add supports for 'cat' of `Boxes` to torchscript.

chenbohua3:

### DIFF
--- a/tests/structures/test_boxes.py
+++ b/tests/structures/test_boxes.py
@@ -189,7 +189,8 @@ class TestBoxes(unittest.TestCase):
         self.assertTrue(x.tensor.shape, (0, 4))
 
     # require https://github.com/pytorch/pytorch/pull/39336
-    @unittest.skipIf(TORCH_VERSION < (1, 7), "Insufficient pytorch version")
+    # require https://github.com/pytorch/pytorch/pull/39821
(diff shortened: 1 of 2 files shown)


tests/structures/test_boxes.py
```diff
@@ -189,7 +189,7 @@ def test_empty_cat(self):
         self.assertTrue(x.tensor.shape, (0, 4))
 
     # require https://github.com/pytorch/pytorch/pull/39336
-    @unittest.skipIf(TORCH_VERSION < (1, 6), "Insufficient pytorch version")
+    @unittest.skipIf(TORCH_VERSION < (1, 7), "Insufficient pytorch version")
```
chenbohua3, line 192: I can't find the pr for torchscript to support "Boxes" annotation in class `Boxes`. Does it belong to 1.6.0 or 1.7.0?:confused:

ppwwyyxx: do you mean https://github.com/pytorch/pytorch/pull/39821 ?

chenbohua3: Yes, that is exactly what I mean, thanks a lot! And this commit belongs to PyTorch 1.6.0, I am going to modify this code now.

Issue #2099, opened 2020-10-04
Extracting stem feature map


ashnair1:
In the ResNet configs, depending on its type, you can extract feature maps from res2 to res5 or res5 alone. I would like to know what would be the simplest way to get the stem feature map (along with res2 to res5) as well.

It's clear from looking at the forward function [here](https://github.com/facebookresearch/detectron2/blob/0dd2b6579a4ba8d6bdddd3f5d010e9c1b620870d/detectron2/modeling/backbone/resnet.py#L433) that this is supported but I'm unsure how to go about this as setting `MODEL.RESNETS.OUT_FEATURES` to ["stem", "res2", "res3", "res4", "res5"] fails at this [line](https://github.com/facebookresearch/detectron2/blob/0dd2b6579a4ba8d6bdddd3f5d010e9c1b620870d/detectron2/modeling/backbone/resnet.py#L613). 

So what is the recommended way to go about this?

ashnair1:
I was able to do this via adding an if check before the line of failure where the stem field is removed before the res blocks are created and then added back before creating the ResNet. 

Regarding the implementation, I had to create a new backbone just to add the if check. Since there is support for extracting the stem feature map, wouldn't it make more sense to add it in `detectron2.modeling.resnet.py`?

ashnair1:
cf3b8583707e507b6515fbdef93ceee3ec8ff467 closes this.

PR #2100, opened 2020-10-04
Allow stem feature map extraction

ashnair1:
From #2099, a fix like this could allow users to extract the 'stem' feature map for downstream tasks by simply specifying `stem` in the `MODELS.RESNETS.OUT_FEATURES`. As mentioned in reference issue, detectron2 seems to support this. It's just that it hits an error while making the res block/stages ([here](https://github.com/facebookresearch/detectron2/blob/0dd2b6579a4ba8d6bdddd3f5d010e9c1b620870d/detectron2/modeling/backbone/resnet.py#L613)).

Example usage
```
BACKBONE:
    NAME: "build_resnet_backbone"
  WEIGHTS: "detectron2://ImageNetPretrained/MSRA/R-50.pkl"
  RESNETS:
    DEPTH: 50
    OUT_FEATURES: ["stem", "res2", "res3", "res4", "res5"]
```

The alternative would be to create a new backbone and implement it there. Personally, I feel its a minor addition that detectron2 could support. However, if it's against its design philosophy, feel free to reject this.

Thanks

### DIFF
--- a/detectron2/modeling/backbone/resnet.py
+++ b/detectron2/modeling/backbone/resnet.py
@@ -608,6 +608,11 @@ def build_resnet_backbone(cfg, input_shape):
 
     stages = []
 
+    rem_feat = []
+    if "stem" in out_features:
+        out_features.remove("stem")
+        rem_feat.append("stem")
+
     # Avoid creating variables without gradients
     # It consumes extra memory and may cause allreduce to fail
     out_stage_idx = [{"res2": 2, "res3": 3, "res4": 4, "res5": 5}[f] for f in out_features]
@@ -641,4 +646,6 @@ def build_resnet_backbone(cfg, input_shape):
         out_channels *= 2
         bottleneck_channels *= 2
         stages.append(blocks)
+    if rem_feat:
+        out_features = out_features + rem_feat
     return ResNet(stem, stages, out_features=out_features).freeze(freeze_at)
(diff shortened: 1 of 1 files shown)


detectron2/modeling/backbone/resnet.py
```diff
@@ -608,6 +608,11 @@ def build_resnet_backbone(cfg, input_shape):
...
     stages = []
 
+    rem_feat = []
+    if "stem" in out_features:
+        out_features.remove("stem")
+        rem_feat.append("stem")
+
     # Avoid creating variables without gradients
     # It consumes extra memory and may cause allreduce to fail
     out_stage_idx = [{"res2": 2, "res3": 3, "res4": 4, "res5": 5}[f] for f in out_features]
```
ppwwyyxx, line 618: this minimal change 
```suggestion
    out_stage_idx = [{"res2": 2, "res3": 3, "res4": 4, "res5": 5}[f] for f in out_features if f != "stem"]
```
is probably enough

ashnair1: Agreed. On a related note, should such a check be added for `linear` as well?

ashnair1: I guess there's no need for it since you won't have that layer when you're using resnet as a backbone.

Issue #2104, opened 2020-10-05
Adding support for Automatic Mixed Precision (AMP)


kondela:
## 🚀 Feature
Since PyTorch version 1.6.0 Automatic Mixed Precision (AMP) is supported natively, implementing it can significantly speed up training and reduce memory footprint without affecting model's performance.

## Motivation & Examples

AFAIK adding AMP is in a roadmap and planned according to this @ppwwyyxx's [comment](https://github.com/facebookresearch/detectron2/issues/72#issuecomment-679642890). If it is not already in internal development phase I was thinking of giving it a try or at least opening up a discussion on how to do it properly.

ppwwyyxx:
This is a duplicate feature request to #72 
An implementation like https://github.com/facebookresearch/detectron2/pull/925, but with native AMP would suffice. The hard work mainly lies in training and benchmarking.

cognitiveRobot:
@ppwwyyxx when I try #925 I get [this issue](https://github.com/facebookresearch/detectron2/pull/925#issuecomment-695898915). Can you please help? Thanks.

PR #776, opened 2020-01-31
add apex fp16 trainer

lbin:
first row: copy from model zoo
second row: our results, 8x2080Ti
faster rcnn
lr sched 3x 
| Name | train time(s/iter)|infer time(s/im)|train mem(GB)|box AP|
|:------:|:------:|:------:|:------:|:------:|
|R50-FPN|0.209|0.047|3.0|40.2|
|R50-FPN-Apex|0.326|0.051|2.8|40.1|

RetineNet
lr sched 3x 
| Name | train time(s/iter)|infer time(s/im)|train mem(GB)|box AP|
|:------:|:------:|:------:|:------:|:------:|
|R50|0.201|0.063|3.9|37.9|
|R50-Apex|0.258|0.077|3.3|38.3|

Mask R-CNN
lr sched 3x 
| Name | train time(s/iter)|infer time(s/im)|train mem(GB)|box AP|mask AP|
|:------:|:------:|:------:|:------:|:------:|:------:|
|R50-FPN|0.261|0.055|3.4|41.0|37.2|
|R50-FPN-Apex|0.388|0.057|3.3|40.9|37.2|

PR #2144, opened 2020-10-18
Adds native PyTorch's automatic mixed precision (AMP) support

kondela:
This is my shot at adding PyTorch's native AMP support. I followed the official examples as close as possible and feedback gathered from https://github.com/facebookresearch/detectron2/issues/2104 and https://github.com/facebookresearch/detectron2/pull/925

Few notes though:
* This probably breaks compatibility with PyTorch 1.5.x
* We should double check modules in `detectron2/layers` for their fp16 and fp32 computational safety. I primarily followed NVIDIA's Apex [whitelist/blacklist of ops](https://github.com/NVIDIA/apex/blob/master/apex/amp/lists/functional_overrides.py) and ones mentioned in #776 but it's possible that I missed some.
* Interaction between JIT tracking/scripting and AMP's autocasting is still work in progress. AFAIK this is only reflected in `test_nms_scriptability` test.
* Most importantly, someone needs to run the benchmarks (it would take ages on my single RTX Titan)

### DIFF
--- a/detectron2/config/defaults.py
+++ b/detectron2/config/defaults.py
@@ -26,6 +26,7 @@ _C.MODEL.LOAD_PROPOSALS = False
 _C.MODEL.MASK_ON = False
 _C.MODEL.KEYPOINT_ON = False
 _C.MODEL.DEVICE = "cuda"
+_C.MODEL.AMP = False
 _C.MODEL.META_ARCHITECTURE = "GeneralizedRCNN"
 
 # Path (a file path, or URL like detectron2://.., https://..) to a checkpoint file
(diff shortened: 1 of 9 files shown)


ppwwyyxx:
Thanks for your PR! We've now added support for native AMP

bsun0802:
May I ask where to find the usage on how to activate the AMP training?

snarb:
@ppwwyyxx  how to use native AMP in detectoron2?

edugp:
here's the commit @snarb @bsun0802 : https://github.com/facebookresearch/detectron2/commit/6be4c339f30238a401d198db96ecd50ac169f137
Adding this to your config should work:
````
SOLVER:
    AMP:
        ENABLED: true
```

PR #2246, opened 2020-11-10
fix(layers): fix logic of shape check in deformalbe

FateScript:
Code [here](https://github.com/facebookresearch/detectron2/blob/master/detectron2/layers/csrc/deformable/deform_conv_cuda.cu#L237) in deformable check the input shape without padding, which is incorrect,  especially for a model using P7 level feature(stride=128). While using 2017_coco_test dataset, such a model with deformable conv leads to crashing.

Or, You might try this code to implement the same crashing.
```Python3
import torch
from torch import nn

from detectron2.layers import DeformConv

x = torch.randn(2, 3, 2, 2).cuda()  # height and width are both 2, which is smaller than kernel_size
offset_conv = nn.Conv2d(3, 18, kernel_size=3, padding=1).cuda()

offset = offset_conv(x)
deform = DeformConv(3, 10, kernel_size=3, padding=1).cuda()

y = deform(x, offset)
``` 

This PR fixes logic of shape checking in dcn. Please kindly review, thanks.

### DIFF
--- a/detectron2/layers/csrc/deformable/deform_conv_cuda.cu
+++ b/detectron2/layers/csrc/deformable/deform_conv_cuda.cu
@@ -234,7 +234,7 @@ void shape_check(
       input.size(1));
 
   TORCH_CHECK(
-      (inputHeight >= kH && inputWidth >= kW),
+      (inputHeight + 2 * padH >= kH && inputWidth + 2 * padW >= kW),
       "input image is smaller than kernel");
 
   TORCH_CHECK(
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
thanks! could you add a unittest?

FateScript:
> thanks! could you add a unittest?

Ok, just unittest  or also with benchmark?

ppwwyyxx:
Just a unittest in `tests/layers` is sufficient

FateScript:
> Just a unittest in `tests/layers` is sufficient

Hello, I add shape check logic for dcnv2 and unittest, but CI failed with `test_box2box_transform.py`, that's wired for me. Could you have a look? Thanks.

ppwwyyxx (approved):

PR #2258, opened 2020-11-13
[layers, deform_conv] Add support for cpu inference of deformable convs

lopeLH:
Since CPU `deform_conv2d` is now available in `torchvision`, supporting this feature is easy. I will be using this in a project, and thought it could be useful for the rest of the world.

About the backwards pass, not sure if it would be easy to implement. I can put some work into it if you can point me in the right direction.


ppwwyyxx:
Thanks. Could you enable or add CPU test in `tests/layers/test_deformable.py`?

detectron2/layers/deform_conv.py
```diff
@@ -46,7 +47,9 @@ def forward(
         ctx.bufs_ = [input.new_empty(0), input.new_empty(0)]  # columns, ones
 
         if not input.is_cuda:
-            raise NotImplementedError("Deformable Conv is not supported on CPUs!")
+            return deform_conv2d(
+                input, offset, weight, stride=stride, padding=padding, dilation=dilation
+            )
```
ppwwyyxx, line 52: Input arguments `groups`, `deformable_groups` are not used.
Either pass them to torchvision, or assert their values if not supported by torchvision.

lopeLH: The `groups` arguments is supported because torchvision infers it from weights' shape. Added the assertion for `deformable_groups`.

lopeLH:
@ppwwyyxx Added a couple of tests. The first one tests the output of the layer on cpu with simple input data (similar to the existing test for gpu). The second test checks that the output is the same whether you run on cpu or gpu, using `groups = 1, 2`.

Issue #2139, opened 2020-10-17
repeat_factors_from_category_frequency does not allow empty annotations


MrParosk:
## 🚀 Feature
At the moment RepeatFactorTrainingSampler.repeat_factors_from_category_frequency does not allow empty annotations (it will throw an error). For some applications, we would like to have a portion of the training samples to have no annotations.

I would like to change that such that if the annotation list is empty, simply set the rep_factor to 1.

## Motivation & Examples

Below is an example of issue:

```python
from detectron2.data.samplers import RepeatFactorTrainingSampler

repeat_thresh = 0.6

# Works since all annotations are non-empty
dataset_dicts_1 = [
    {"annotations": [{"category_id": 0}, {"category_id": 1}]},
    {"annotations": [{"category_id": 0}]},
]

res_1 = RepeatFactorTrainingSampler.repeat_factors_from_category_frequency(
    dataset_dicts_1, repeat_thresh)

print(res_1)

# Does not work due to the empty annotation, will throw a value error.
dataset_dicts_2 = [
    {"annotations": [{"category_id": 0}, {"category_id": 1}]},
    {"annotations": [{"category_id": 0}]},
    {"annotations": []}
]

res_2 = RepeatFactorTrainingSampler.repeat_factors_from_category_frequency(
    dataset_dicts_2, repeat_thresh)

print(res_2)
```

The fix would be easy, simply check if len(cat_ids) == 0, if so set the factor to 1 else do as previously done.

https://github.com/facebookresearch/detectron2/blob/7f713b7a224fbb6d7ee10fe7e7dcd7ed1edfb885/detectron2/data/samplers/distributed_sampler.py#L126.

## Note

I can contribute this change myself, just wanted to see if this change is welcomed before fixing it.

ppwwyyxx:
thanks, the proposed change makes sense to me!

PR #2262, opened 2020-11-14
Allow empty annotations for repeat_factors_from_category_frequency

MrParosk:
This PR enables the use of empty annotations for repeat_factors_from_category_frequency.

Related to issue: https://github.com/facebookresearch/detectron2/issues/2139

### DIFF
--- a/detectron2/data/samplers/distributed_sampler.py
+++ b/detectron2/data/samplers/distributed_sampler.py
@@ -124,7 +124,11 @@ class RepeatFactorTrainingSampler(Sampler):
         rep_factors = []
         for dataset_dict in dataset_dicts:
             cat_ids = {ann["category_id"] for ann in dataset_dict["annotations"]}
-            rep_factor = max({category_rep[cat_id] for cat_id in cat_ids})
+
+            if len(cat_ids) > 0:
+                rep_factor = max({category_rep[cat_id] for cat_id in cat_ids})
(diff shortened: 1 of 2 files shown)


detectron2/data/samplers/distributed_sampler.py
```diff
@@ -124,7 +124,11 @@ def repeat_factors_from_category_frequency(dataset_dicts, repeat_thresh):
         rep_factors = []
         for dataset_dict in dataset_dicts:
             cat_ids = {ann["category_id"] for ann in dataset_dict["annotations"]}
-            rep_factor = max({category_rep[cat_id] for cat_id in cat_ids})
+
+            if len(cat_ids) > 0:
+                rep_factor = max({category_rep[cat_id] for cat_id in cat_ids})
```
ppwwyyxx, line 129: this is sufficient
```suggestion
                rep_factor = max({category_rep[cat_id] for cat_id in cat_ids}, default=1.0)
```

ppwwyyxx (changes requested):

detectron2/data/samplers/distributed_sampler.py
MrParosk: Yes that is a lot cleaner, will fix it, thanks

ppwwyyxx (approved):

PR #2323, opened 2020-11-28
fixed issue with _eval_predictions in rotated-coco

MrParosk:
At the moment RotatedCOCOEvaluator is not working.
The issue is that it is derived from COCOEvaluator, which call the method _eval_predictions with parameter img_ids  https://github.com/facebookresearch/detectron2/blob/cd60faf1427f95d357ba30365a415d0309a8618f/detectron2/evaluation/coco_evaluation.py#L173.

However, RotatedCOCOEvaluator's _eval_predictions doesn't have that parameter https://github.com/facebookresearch/detectron2/blob/cd60faf1427f95d357ba30365a415d0309a8618f/detectron2/evaluation/rotated_coco_evaluation.py#L148.

One could add the img_ids=None (which will be unused) aswell.

### DIFF
--- a/detectron2/evaluation/rotated_coco_evaluation.py
+++ b/detectron2/evaluation/rotated_coco_evaluation.py
@@ -145,7 +145,7 @@ class RotatedCOCOEvaluator(COCOEvaluator):
             results.append(result)
         return results
 
-    def _eval_predictions(self, predictions):
+    def _eval_predictions(self, predictions, **kwargs):
         """
         Evaluate predictions on the given tasks.
         Fill self._results with the metrics of the tasks.


MrParosk:
I notice another problem with RotatedCOCOEvaluator, it seems _task has been renamed to _tasks in COCO-evalutator:

https://github.com/facebookresearch/detectron2/blob/cd60faf1427f95d357ba30365a415d0309a8618f/detectron2/evaluation/coco_evaluation.py#L95

Which causes problems in:

https://github.com/facebookresearch/detectron2/blob/cd60faf1427f95d357ba30365a415d0309a8618f/detectron2/evaluation/rotated_coco_evaluation.py#L177

This should probably be added in this PR, therefore this PR should not be merged yet.

ppwwyyxx:
Thanks for the bug report! Adding an unused `img_ids` to `_eval_predictions` sounds OK.

Issue #2525, opened 2021-01-21
rotated_nms is not scriptable


MrParosk:
## 🚀 Feature

At the moment rotated_nms is not scriptable; see example and discussion below.

## Motivation & Examples

Running the following code yields:

```python
import torch
from detectron2.layers.nms import nms_rotated

class MyModule(torch.nn.Module):
    def forward(self, boxes, scores, threshold):
        return nms_rotated(boxes, scores, threshold)

m = torch.jit.script(MyModule())
torch.jit.save(m, 'scriptmodule.pt')
```

```python
Python builtin <built-in method nms_rotated of PyCapsule object at XXXX> is currently not supported in Torchscript:
```

## Note

I think the reason why the function is not scriptable is that the function is defined with pybind:
https://github.com/facebookresearch/detectron2/blob/42ff5347f177e4f9883efea94da7b51d862c6151/detectron2/layers/csrc/vision.cpp#L70

However in the pytorch-docs they register custom ops in a different way (in order to make it scriptable)(https://pytorch.org/tutorials/advanced/torch_script_custom_ops.html).

Is this assumption correct? I.e. that we only need to register the ops differently and make some (small ?) changes?

Also, if these changes would be implemented, is it better to have them in detectron2 or torchvision? I saw a comment about it (see below):

https://github.com/facebookresearch/detectron2/blob/42ff5347f177e4f9883efea94da7b51d862c6151/detectron2/layers/nms.py#L33

If implemented in detectron2, some extra work would need to be made s.t. we load these ops similar to how they do it in torchvision:

https://github.com/pytorch/vision/blob/072d8b2280569a2d13b91d3ed51546d201a57366/torchvision/extension.py#L47

ppwwyyxx:
> Is this assumption correct? I.e. that we only need to register the ops differently and make some (small ?) changes?

That's correct. I don't know if this will force all C++ code to be migrated to using `TORCH_LIBRRAY` - which might become a large change.

For now I'm not aware of any plans to add rotated NMS to torchvision.

PR #2550, opened 2021-01-26
Add support of torchscript for rotated_nms

MrParosk:
Related to issue: https://github.com/facebookresearch/detectron2/issues/2525

I had to change the types from float to double in the files for it to compile.


MrParosk:
Forgot to test for pytorch 1.5 / 1.6, which probably have less support for torch-script ops compared to 1.7. Is it best to create a macro which checks the pytorch version, and if <= 1.6, include it in the pybind, else in TORCH_LIBRARY?

ppwwyyxx:
There is an open issue in pytorch https://github.com/pytorch/pytorch/issues/44365. Before that happens, maybe you can workaround it by 
```
#ifdef TORCH_LIBRARY  
```
or if this does not work, define macros in `setup.py`.

PR #2584, opened 2021-02-03
docker-compose gpu access updated, since 1.28.0 version

JaledMC:
Updating gpu access from docker-compose, pointed by this comment:

https://github.com/facebookresearch/detectron2/blob/45a8bfb64053d71d9d7f136fb25a6abe841dc91f/docker/docker-compose.yml#L9

The solution comes from this [pull request](https://github.com/docker/compose/issues/6691), and working since 1.28.0 [release](https://github.com/docker/compose/releases). It's the [official](https://github.com/docker/compose/pull/7929) replace of `runtime: nvidia`

In this way, we don't need to install nvidia-docker (less prerequisites :tada:), but nvidia-container-toolkit seems to be needed yet.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -14,9 +14,9 @@ docker run --gpus all -it \
 xhost +local:`docker inspect --format='{{ .Config.Hostname }}' detectron2`
 ```
 
-## Use the container (with docker < 19.03)
+## Use the container (with docker-compose ≥ 1.28.0)
 
-Install docker-compose and nvidia-docker2, then run:
+Install docker-compose and nvidia-docker-toolkit, then run:
 ```
 cd docker && USER_ID=$UID docker-compose run detectron2
 ```
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -21,3 +21,6 @@ services:
     environment:
       - DISPLAY=$DISPLAY
       - NVIDIA_VISIBLE_DEVICES=all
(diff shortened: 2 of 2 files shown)


ppwwyyxx:
Thanks, could you update docker/README.md as well?
We probably don't need `with docker < 19.03` section but need one specifically for docker-compose

JaledMC:
Yes. Good point about documentation.

PD: during testing, I wanted to use webcam inside the docker, and I had to go to previous model, [maskrcnn-benchmark, to enable camera device](https://github.com/facebookresearch/maskrcnn-benchmark/tree/master/demo#with-docker). Do you think this paremeter deserves to appear in the documentation?

ppwwyyxx:
I think you can add https://medium.com/@zwinny/docker-using-webcam-9fafb26cf1e6 do docker-compose config file

JaledMC:
Done. One option could be add to the Readme:

For docker-compose versions lower than 1.28.0, install nvidia-docker2, and replace the code:
```
  deploy:
    resources:
      reservations:
        devices:
          - capabilities:
            - gpu
```
For this line:
```
runtime: nvidia
```

But IMHO, in the future there is no need for this information, and it requires another commit to remove it.

Commented lines for docker-compose webcam added:
```
# Uncomment with proper source to access webcam from docker
# devices: 
#   - /dev/video0:/dev/video0
```


Feel free to point me any required changes, and kudos for this repo!

PR #2636, opened 2021-02-19
Update defaults.py

PradyumnaGupta:
Changes made to Default Trainer to allow for custom model , optimizer and data_loader . This will allow to take all the benefits of Default Trainer for training custom models without having to go to Simple Trainer or writing your own training loops.


ppwwyyxx:
>  . This will allow to take all the benefits of Default Trainer for training custom models without having to go to Simple Trainer or writing your own training loops

You can already custom models by overriding `self.build_model`. In general we try to not mix cfg with arguments in one signature.

PR #2660, opened 2021-02-25
Update write-models.md

edugp:
Update the docs with instructions on how to fully train a custom backbone

Thanks for your contribution!

### DIFF
--- a/docs/tutorials/write-models.md
+++ b/docs/tutorials/write-models.md
@@ -42,6 +42,10 @@ cfg = ...   # read a config
 cfg.MODEL.BACKBONE.NAME = 'ToyBackbone'   # or set it in the config file
 model = build_model(cfg)  # it will find `ToyBackbone` defined above
 ```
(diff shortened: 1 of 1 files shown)


docs/tutorials/write-models.md
```diff
@@ -42,6 +42,10 @@ cfg = ...   # read a config
 cfg.MODEL.BACKBONE.NAME = 'ToyBackbone'   # or set it in the config file
 model = build_model(cfg)  # it will find `ToyBackbone` defined above
 ```
+Note that, if we would like to train all the parameters of our new backbone, we would need to set the `FREEZE_AT` parameter to 0:
```
ppwwyyxx, line 45: this is not true. Because `ToyBackbone.__init__` does not freeze any layers, this setting does not affect it

edugp: oh true! I was trying to make the default `FREEZE_AT = 2` behavior explicit somewhere in the tutorials, but I can't find any good place, so I'll close this. Let me know if you can think of any and I can reopen it. Thank you!

edugp:
The example I added is incorrect

PR #2716, opened 2021-03-09
Replace in-place clamp_ with out-place one

apivovarov:
This PR continues the initiative to replace in-place operators (such as clamp_ , copy_, etc) with out-place version of them. The modification of the graph structure is needed to make it easier to use the model with 3rd party inference frameworks (e.g. TVM, ONNX, etc).

Related discussions:

#2678
@ppwwyyxx

### DIFF
--- a/detectron2/structures/boxes.py
+++ b/detectron2/structures/boxes.py
@@ -202,10 +202,11 @@ class Boxes:
         """
         assert torch.isfinite(self.tensor).all(), "Box tensor contains infinite or NaN!"
         h, w = box_size
-        self.tensor[:, 0].clamp_(min=0, max=w)
-        self.tensor[:, 1].clamp_(min=0, max=h)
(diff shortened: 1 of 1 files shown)


apivovarov:
Yuxin, good point about `relu_` and `clamp_`.
They both are supported in TVM in generic case.
But the combination of slice and in_place operator is not supported in TVM, e.g.
```
# Supported
def forward(self, a):
  a.clamp_(min=0, max=5)
  return a

# Not supported
def forward(self, a):
  a[:,0].clamp_(min=0, max=5)
  return a
```

Let me minimize the changes in my PR and fix `clamp_` where it is actually necessary - in [Boxes.clip()](https://github.com/facebookresearch/detectron2/blob/master/detectron2/structures/boxes.py#L205)

ppwwyyxx (approved):

PR #2742, opened 2021-03-13
Scriptability support for rrpn

MrParosk:
This PR adds torchscript scriptability support for RRPN, similar to RPN.


MrParosk:
Just out of curiosity, will this PR be merged or are there technical / other reason it will not?

PR #2824, opened 2021-03-26
Update loss_weight in FastRCNNOutputLayers

huhuman:
As only 'loss_box_reg' can be assigned by config, I just complete 'loss_weight' which is the same with documentation of its args:
<pre><code>loss_weight (float|dict): weights to use for losses. Can be single float for weighting
                all losses, or a dict of individual weightings. Valid dict keys are:
                    * "loss_cls": applied to classification loss
                    * "loss_box_reg": applied to box regression loss
</code></pre>


ppwwyyxx:
As mentioned in https://github.com/facebookresearch/detectron2/blob/bbdee4cf6bc100fedad7cfad3225bb1da12fa175/detectron2/config/defaults.py#L4-L6 , we have stopped adding new features to CfgNode. 
Loss weight can be customized by the argument of `FastRCNNOutputLayers` already, so the new config system already supports it.

PR #2867, opened 2021-04-03
Update INSTALL.md

j450h1:
Enter full command and clarify that you need to pull from git for macOS as well

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -17,10 +17,10 @@ python -m pip install 'git+https://github.com/facebookresearch/detectron2.git'
 
 # Or, to install it from a local clone:
 git clone https://github.com/facebookresearch/detectron2.git
-python -m pip install -e detectron2
 
+python -m pip install -e detectron2
 # Or if you are on macOS
-CC=clang CXX=clang++ ARCHFLAGS="-arch x86_64" python -m pip install ......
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
>  clarify that you need to pull from git for macOS as well

I don't think that is needed. Both `python -m pip install 'git+https://github.com/facebookresearch/detectron2.git'` and 
`python -m pip install -e detectron2` are supposed to work.

j450h1:
> > clarify that you need to pull from git for macOS as well
> 
> I don't think that is needed. Both `python -m pip install 'git+https://github.com/facebookresearch/detectron2.git'` and
> `python -m pip install -e detectron2` are supposed to work.

The part that was confusing for me when installing were two things:

1. `CC=clang CXX=clang++ ARCHFLAGS="-arch x86_64" python -m pip install ......`

That ..... I just copied and pasted and it didn't work. Then I assumed it meant just copy from above and got it to work.

2. 2nd part is where you put the comment of "# Or if you are on macOS". It made it look like git clone is not required and you could simply copy and paste that 1 line for macOS. Maybe for beginners it could be confusing and they won't be able to deduce that you need to do git clone as I eventually figured out.

I just found it confusing, but maybe its just me and then feel free to close the PR.

Issue #282, opened 2019-11-09
Do you support batch inference?


zhaoxuyan:
## 🚀 Feature
How to easily take advantage of batch processing during inference?

## Motivation

https://github.com/facebookresearch/Detectron/issues/84

ppwwyyxx:
You just need to call model with a batch of inputs (https://detectron2.readthedocs.io/tutorials/models.html#model-input-format).

zhaoxuyan:
> You just need to call model with a batch of inputs (https://detectron2.readthedocs.io/tutorials/models.html#model-input-format).

Thank you very much for your reply~
Can you tell me how to set up a batch of pictures in `/demo/demo.py`
```python
img = read_image(path, format="BGR")
demo.run_on_image(img)
```
It can only infer one picture at a time, how to infer a batch of pictures?

ppwwyyxx:
demo.py does not support batch inference.

zhaoxuyan:
> demo.py does not support batch inference.

So how should I implement it please?

ppwwyyxx:
https://detectron2.readthedocs.io/tutorials/models.html#use-models explains how to use a model.

keko950:
> You just need to call model with a batch of inputs (https://detectron2.readthedocs.io/tutorials/models.html#model-input-format).

That link doesn't refers in any way to batch inference, would be nice a more detailed answer since the api doesn't go deeper.

ppwwyyxx:
The link explains that the input to the model is a `list[{"image": ...}, ...]`. As long as the length of the list > 1, you're doing batch inference.

keko950:
I was thinking in DefaultPredictor, not model class itself, my bad!
Working like a charm, thank you!

zhaoxuyan:
> The link explains that the input to the model is a `list[{"image": ...}, ...]`. As long as the length of the list > 1, you're doing batch inference.

Thank you for your reply. Another question, how to specify multi gpu in DefaultPredictor?

ppwwyyxx:
`DefaultPredictor` uses 1 GPU.

darkAlert:
I create a predictor as follows:
```
	cfg = get_cfg()
	cfg.merge_from_file(path_to_cfg)
	cfg.MODEL.ROI_HEADS.SCORE_THRESH_TEST = 0.5 
	cfg.MODEL.WEIGHTS = path_to_weights
	predictor = DefaultPredictor(cfg)
        
[...]

PR #2888, opened 2021-04-08
Add batch predictor 

sailist:
It's not very convenient when do batch inference by using original detectron2 api, the solution provided here https://github.com/facebookresearch/detectron2/issues/282#issuecomment-737270926 is good, and I extended it's code to support more forms.

### DIFF
--- a/detectron2/engine/defaults.py
+++ b/detectron2/engine/defaults.py
@@ -273,11 +273,12 @@ class BatchPredictor(DefaultPredictor):
             elif batch.ndim == 4:
                 batch = [i.squeeze() for i in np.split(batch, batch.shape[0])]
             else:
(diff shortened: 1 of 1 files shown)


ashnair1:
@sailist Have you tried using this BatchPredictor with AsyncPredictor?

sailist:
Fine, I'll try it, thanks.

Sorakuhaku:
Can you add unit test to ensure it is working properly?

ppwwyyxx:
As the doc of DefaultPredictor mentions, it is only for simple demo purposes and we do not plan to extend it with new capabilities. Users are expected to use the `model` object following its formats (which supports batch predictions) if more complicated logic is needed.

PR #2931, opened 2021-04-19
Sample image instead of dataset for exporting model

elb3k:
[export_model.py](https://github.com/facebookresearch/detectron2/blob/master/tools/deploy/export_model.py)
 requires to process sample input, and current implementation requires whole dataset to be imported, it is not a easy way for custom models, so instead of dataset, user can pass a sample image.

### DIFF
--- a/tools/deploy/export_model.py
+++ b/tools/deploy/export_model.py
@@ -3,7 +3,6 @@
 import argparse
 import os
 from typing import Dict, List, Tuple
-import cv2
 import onnx
 import torch
 from torch import Tensor, nn
@@ -11,7 +10,7 @@ from torch import Tensor, nn
 import detectron2.data.transforms as T
 from detectron2.checkpoint import DetectionCheckpointer
 from detectron2.config import get_cfg
-from detectron2.data import build_detection_test_loader
+from detectron2.data import build_detection_test_loader, detection_utils
 from detectron2.evaluation import COCOEvaluator, inference_on_dataset, print_csv_format
 from detectron2.export import (
     Caffe2Tracer,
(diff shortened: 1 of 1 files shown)


tools/deploy/export_model.py
```diff
@@ -187,9 +190,29 @@ def eval_wrapper(inputs):
...
+            # whether the model expects BGR inputs or RGB
+            original_image = original_image[:, :, ::-1]
+        height, width = original_image.shape[:2]
+        image = aug.get_transform(original_image).apply_image(original_image)
+        image = torch.as_tensor(image.astype("float32").transpose(2, 0, 1))
+
+        inputs = {"image": image, "height": height, "width": width}
+
+        # Sample ready
+        first_batch = [inputs]
```
ppwwyyxx, line 215: the name `first_batch` only makes sense when getting data from the dataset. please call it `sample_inputs`

tools/deploy/export_model.py
```diff
@@ -187,9 +190,29 @@ def eval_wrapper(inputs):
...
     torch_model.eval()
 
-    # get a sample data
-    data_loader = build_detection_test_loader(cfg, cfg.DATASETS.TEST[0])
-    first_batch = next(iter(data_loader))
+    if args.sample_image is None:
+        # get a sample data
+        data_loader = build_detection_test_loader(cfg, cfg.DATASETS.TEST[0])
+        first_batch = next(iter(data_loader))
+    else:
```
ppwwyyxx, line 197: please move all the logic into a separate function called `get_sample_inputs`

tools/deploy/export_model.py
```diff
@@ -187,9 +190,29 @@ def eval_wrapper(inputs):
...
+        # get a sample data
+        original_image = cv2.imread(args.sample_image)
```
ppwwyyxx, line 199: please use `detection_utils.read_image` to be consistent with how dataset is loaded. It will also support input_format conversion.

elb3k:
@ppwwyyxx all done.

PR #2964, opened 2021-04-24
fix: throw KeyError when there are custom attributes in proposals

QingQiz:
fix: throw KeyError when there are custom attributes in proposals

### DIFF
--- a/detectron2/modeling/proposal_generator/proposal_utils.py
+++ b/detectron2/modeling/proposal_generator/proposal_utils.py
@@ -1,7 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import logging
 import math
-from typing import List, Tuple
+from typing import List, Tuple, Union
 import torch
 
 from detectron2.layers import batched_nms, cat
@@ -129,12 +129,13 @@ def find_top_rpn_proposals(
         results.append(res)
     return results
 
-def add_ground_truth_to_proposals(gt: List[Instances], proposals: List[Instances]):
+
+def add_ground_truth_to_proposals(gt: Union[List[Instances], List[Boxes]], proposals: List[Instances]):
     """
     Call `add_ground_truth_to_proposals_single_image` for all images.
 
     Args:
-        gt(list[Instances]): list of N elements. Element i is a Instances
+        gt(Union[List[Instances], List[Boxes]): list of N elements. Element i is a Instances
             representing the ground-truth for image i.
         proposals (list[Instances]): list of N elements. Element i is a Instances
             representing the proposals for image i.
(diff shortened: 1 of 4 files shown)


ppwwyyxx:
Could you explain what problem is this trying to fix?

QingQiz:
When there are new attributes in proposals predicted by my customized RPN, a KeyError will thrown by 
https://github.com/facebookresearch/detectron2/blob/f50ec07cf220982e2c4861c5a9a17c4864ab5bfd/detectron2/modeling/proposal_generator/proposal_utils.py#L180

---

For example: 

when there is an `alpha` in the output of RPN (`[Instances(num_instances=0, image_height=820, image_width=800, fields=[proposal_boxes: Boxes(tensor([], device='cuda:0', size=(0, 4))), objectness_logits: tensor([], device='cuda:0'), alpha: tensor([], device='cuda:0', size=(0, 2))])]`), a `KeyError` will be thrown by https://github.com/facebookresearch/detectron2/blob/f50ec07cf220982e2c4861c5a9a17c4864ab5bfd/detectron2/modeling/proposal_generator/proposal_utils.py#L180
in https://github.com/facebookresearch/detectron2/blob/f50ec07cf220982e2c4861c5a9a17c4864ab5bfd/detectron2/modeling/proposal_generator/proposal_utils.py#L159
in 
https://github.com/facebookresearch/detectron2/blob/f50ec07cf220982e2c4861c5a9a17c4864ab5bfd/detectron2/modeling/proposal_generator/proposal_utils.py#L133
in
https://github.com/facebookresearch/detectron2/blob/f50ec07cf220982e2c4861c5a9a17c4864ab5bfd/detectron2/modeling/roi_heads/roi_heads.py#L259

detectron2/modeling/proposal_generator/proposal_utils.py
```diff
@@ -167,14 +166,15 @@ def add_ground_truth_to_proposals_single_image(gt_boxes, proposals):
...
+    gt_boxes = gt.gt_boxes
     device = proposals.objectness_logits.device
     # Assign all ground-truth boxes an objectness logit corresponding to
     # P(object) = sigmoid(logit) =~ 1.
     gt_logit_value = math.log((1.0 - 1e-10) / (1 - (1.0 - 1e-10)))
     gt_logits = gt_logit_value * torch.ones(len(gt_boxes), device=device)
 
     # Concatenating gt_boxes with proposals requires them to have the same fields
-    gt_proposal = Instances(proposals.image_size)
+    gt_proposal = Instances(proposals.image_size, **gt.get_fields())
```
ppwwyyxx, line 177: This still fails if `gt` has fields not predicted by the RPN (e.g. `gt_masks`, `gt_boxes`), or your custom RPN predicts fields that are not in gt.

The correct way would be to: for every field in `proposals` are not `proposal_boxes/objectness_logits`, use the field with the same name in `gt`. If the field does not exist in `gt`, throws an error.

detectron2/modeling/proposal_generator/proposal_utils.py
```diff
@@ -129,36 +129,35 @@ def find_top_rpn_proposals(
         results.append(res)
     return results
 
-
-def add_ground_truth_to_proposals(gt_boxes, proposals):
+def add_ground_truth_to_proposals(gt: List[Instances], proposals: List[Instances]):
```
ppwwyyxx, line 132: To maintain backward compatibility, we should allow the function to take `Union[List[Instances], List[Boxes]]`

ppwwyyxx (commented):
The use case makes sense to me. Please also add a unittest in `test_rpn.py` that tests the new behavior for this function.

Issue #2975, opened 2021-04-26
PointRend: no 'pred_masks' field in instances when there are no predictions


MiroPsota:
[This commit](https://github.com/facebookresearch/detectron2/commit/c1689c7d7b1c1af0157cd90b84f296b7b2ce2124) changes behavior of `PointRendMaskHead` when there are no predictions on image. It does not add field `pred_masks` as before this commit. I expect this is not wanted behavior, as MaskRCNN also adds this field.

I think it's because there's no call `mask_rcnn_inference` now, when there are no predictions.
[before](https://github.com/facebookresearch/detectron2/blob/1dac147f824a3c8d908ff253cb8027f15844eb2f/projects/PointRend/point_rend/mask_head.py#L197)
[now](https://github.com/facebookresearch/detectron2/blob/master/projects/PointRend/point_rend/mask_head.py#L307)

I can provide MWE if needed.

ppwwyyxx:
@bowenc0221 any idea why
```
        if len(pred_classes) == 0:
            return instances
```
was added?

bowenc0221:
```
        if len(pred_classes) == 0:
            return instances
```
was added to make it compatible with Implicit PointRend. If there is no instance predicted, the parameter head will be an empty tensor and cause an error when doing convolution.

Any idea to better resolve this?

ppwwyyxx:
torch's convolution ops are able to deal with tensors with batch size 0 (which is why this can work in mask r-cnn). do you have more specific examples?

bowenc0221:
Implicit PointRend takes this batch size 0 tensor and predicts the `weight` of some group convolution which will fail? Because the group size = # of instance which is 0.

And because Implicit PointRend does not predict `coarse_mask`, we do not have this `mask_logits` so I removed `mask_rcnn_inference`.

We can fix it with something like this?
```python
        if len(pred_classes) == 0:
            return mask_rcnn_inference(torch.zeros((0, 28, 28)), instances)
```

MiroPsota:
This seems to work:
```
        if len(pred_classes) == 0:
            mask_rcnn_inference(torch.zeros((0, 1, 28, 28)), instances)
            return instances
```

ppwwyyxx:
It would be good to make sure all components work with empty inputs - they are so common in detection models. For the case in implicit pointrend, replacing 1x1 conv with a matmul will support empty size and also reads more clear to me:
```python
import torch
from torch.nn import functional as F

num_instances = 0
[...]

PR #2998, opened 2021-05-02
Add Implicit PointRend

bowenc0221:
Add Implicit PointRend model described in "Pointly-Supervised Instance Segmentation".

### DIFF
--- a/projects/PointRend/README.md
+++ b/projects/PointRend/README.md
@@ -152,3 +152,16 @@ If you use PointRend, please use the following BibTeX entry.
   year={2019}
 }
 ```
+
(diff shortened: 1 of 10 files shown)


bowenc0221:
@ppwwyyxx Convolution is replaced with matrix multiplication according to https://github.com/facebookresearch/detectron2/issues/2975#issuecomment-827379225

ppwwyyxx:
I think now #2975 can be fixed? subdivision should return empty `pred_masks` when there is no detections

bowenc0221:
> I think now #2975 can be fixed? subdivision should return empty `pred_masks` when there is no detections

Fixed in 038a3b2

Issue #2901, opened 2021-04-12
Can not process video in the colab example.


josenxx:
## 📚 Documentation
I try to run the colab example and it breaks in the video part.
* Link to the relevant documentation/comment:
here is error message
# Run frame-by-frame inference demo on this video (takes 3-4 minutes) with the "demo.py" tool we provided in the repo.
!git clone https://github.com/facebookresearch/detectron2
!python detectron2/demo/demo.py --config-file detectron2/configs/COCO-PanopticSegmentation/panoptic_fpn_R_101_3x.yaml --video-input video-clip.mp4 --confidence-threshold 0.6 --output video-output.mkv \
  --opts MODEL.WEIGHTS detectron2://COCO-PanopticSegmentation/panoptic_fpn_R_101_3x/139514519/model_final_cafdb1.pkl
* What is missing or inaccurate?
*May be there is an error about video encoder.
change：fourcc=cv2.VideoWriter_fourcc(*"x264")
to：fourcc=cv2.VideoWriter_fourcc(*"mp4v"), # x264 ==> mp4

ppwwyyxx:
Yes, the opencv on colab is missing x264 codec so a workaround is to change it to use mp4. Added a note to colab.

Ideally there should be a utility function that first checks if x264 is available (by writing to a temp video file). Leaving this open as an enhancement.

ArtistBanda:
@ppwwyyxx could this be solved just for colab like:
```python
if "google.colab" in sys.modules:
     [required changes...]
```
Ideally, it should be able to catch the exception raised by C++ but I don't know how to do that, if you could guide me or maybe I could apply the above fix?

P.S : This is my first time contributing to detectron2

ppwwyyxx:
solved by #3077

Ayaan-Sharif:
`import locale
def getpreferredencoding(do_setlocale = True):
    return "UTF-8"
locale.getpreferredencoding = getpreferredencoding  `
this works for me google colab

Ayaan-Sharif:
`import locale
def getpreferredencoding(do_setlocale = True):
    return "UTF-8"
locale.getpreferredencoding = getpreferredencoding`

PR #3077, opened 2021-05-22
Added mp4 support and check for x264 in demo.py

ArtistBanda:
fix issue #2901 

Adds a check for `x264` support and shifts to `mp4v` codec if not available.

Also tested on `google.colab`.

### DIFF
--- a/demo/demo.py
+++ b/demo/demo.py
@@ -2,8 +2,10 @@
 import argparse
 import glob
 import multiprocessing as mp
+import numpy as np
 import os
 import time
+import warnings
 import cv2
 import tqdm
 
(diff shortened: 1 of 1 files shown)


demo/demo.py
```diff
@@ -70,6 +72,24 @@ def get_parser():
     return parser
 
 
+def test_opencv_video_format(codec, file_ext):
+    filename = "test_file" + file_ext
```
ppwwyyxx, line 76: It should be inside a temporary directory (use `tempfile.TemporaryDirectory`). then also remove `os.remove`

demo/demo.py
```diff
@@ -70,6 +72,24 @@ def get_parser():
...
+        fps=float(30),
+        frameSize=(10, 10),
+        isColor=True,
+    )
+    [writer.write(np.zeros((10, 10, 3), np.uint8)) for _ in range(30)]
+    writer.release()
+    if os.path.isfile(filename):
+        os.remove(filename)
+        return True
+    warnings.warn("x264 codec not available switching to mp4v")
```
ppwwyyxx, line 89: this warning shouldn't be here because it depends on what arguments this function is called with. either remove it or move it inside `__main__`

ArtistBanda: Yeah, I thought of the same but thinking about the scope of this function I put it there but will change it as it may be confusing in the future.

demo/demo.py
ArtistBanda: Sure, will do that.

ppwwyyxx:
Thanks!

PR #3171, opened 2021-06-22
Add config option for automatic subdir creation

markusschoen:
I added a config option WRITE_OUTPUT_TO_SUBDIR, which creates a sub-directory in the set OUTPUT_DIR based on the current date-time and config file name.
This is handy when running multiple train runs without having to specify a different OUTPUT_DIR for each run.
The default behavior is unchanged, since the config option is disabled by default.


ppwwyyxx:
We will not support this feature. Users are already able to specify the directory by a string, and this provides enough flexibility for implementing other custom directory names.

Issue #2114, opened 2020-10-07
Feature: a hook to automatically save the best model during training


xjinai:
## 🚀 Feature
I'm using train_net.py to train model using custom dataset. It worked well, especially with multiple GPUs. Thank you for your great work!

I set SOLVER.CHECKPOINT_PERIOD (e.g. 1000) to save the model every 1000 iterations. I want Detectron2 to automatically save the best model as the training goes, so that I can use the best model later for inference and evaluation. Currently this feature is missing in Detectron2. 

## Motivation & Examples

I want Detectron2 to save the best model as the training goes, so that I can use the best model later for inference and evaluation. This feature exists in other object detection frameworks, for example, Darknet. I found that it is very useful since the model may overfit with the specified number of iterations. Therefore, the model_final.pth saved by Detectron2 is typically not the best model. If the best model is saved automatically during training, I can read it later and apply it for inference and evaluation.

A possible implementation:
1) Save the model at the first evaluation period to a file named model_best.pth.
2) At every following evaluation period, compare AP value (on the validation set) with the AP value of the last evaluation period. If AP is better, then automatically update the saved best model (model_best.pth). 

## Note

I believe that this feature is relevant to and would be highly appreciated by many users since saving and applying best model is a common practice for object detection.

ppwwyyxx:
Duplicate of #645

xjinai:
Thank you for your response. 

I looked at #645 before. It is not exactly what I want. What I want is saving the best model during training **as the default behavior**, no matter if I run training from train_net.py or tutorial in Colab notebook. Please consider to add this feature. 

Regarding #645, can you please elaborate on how to add the logic "using the return value of do_test" in plain_train_net.py? Does it work with custom dataset?

ppwwyyxx:
I think it is not reasonable to make it automatically a default because 
* there is no definition of "best" model unless users first choose a metric somehow. So it will require a new user input somewhere
* these example training scripts are examples, as written on the top of those files. They are not meant to support every nice features
* in research, using the best model is not a common thing to do (and sometimes not the right thing to do)

`do_test` in `plain_train_net` returns the metrics. So it works for any dataset as long as there are some metrics.

xjinai:
@ppwwyyxx: Implementing a reusable hook that can be manually added by users to a trainer sounds like a good idea. It would be helpful if an example code (for example, to address my use case above) can be provided. Thank you for your consideration. 

[...]

PR #3222, opened 2021-07-05
Best Checkpointing Hook

levan92:
Implemented hook for saving best model, as discussed in #2114

### DIFF
--- a/detectron2/config/defaults.py
+++ b/detectron2/config/defaults.py
@@ -531,10 +531,6 @@ _C.SOLVER.WARMUP_METHOD = "linear"
 # Save a checkpoint after every this number of iterations
 _C.SOLVER.CHECKPOINT_PERIOD = 5000
 
-# Used with `hooks.BestCheckpointer`. Checkpoints every best model based on given eval metric.
-_C.SOLVER.BEST_CHECKPOINTER = CN({"ENABLED": False})
-_C.SOLVER.BEST_CHECKPOINTER.METRIC = "bbox/AP50"
-
 # Number of images per batch across all machines. This is also the number
 # of training images per step (i.e. per iteration). If we use 16 GPUs
 # and IMS_PER_BATCH = 32, each GPU will see 2 images per batch.
--- a/detectron2/config/defaults.py
+++ b/detectron2/config/defaults.py
@@ -531,6 +531,10 @@ _C.SOLVER.WARMUP_METHOD = "linear"
 # Save a checkpoint after every this number of iterations
 _C.SOLVER.CHECKPOINT_PERIOD = 5000
 
+# Used with `hooks.BestCheckpointer`. Checkpoints every best model based on given eval metric.  
+_C.SOLVER.BEST_CHECKPOINTER = CN({"ENABLED": False})
+_C.SOLVER.BEST_CHECKPOINTER.METRIC = "bbox/AP50"
+
 # Number of images per batch across all machines. This is also the number
 # of training images per step (i.e. per iteration). If we use 16 GPUs
 # and IMS_PER_BATCH = 32, each GPU will see 2 images per batch.
--- a/detectron2/config/defaults.py
+++ b/detectron2/config/defaults.py
@@ -531,7 +531,7 @@ _C.SOLVER.WARMUP_METHOD = "linear"
 # Save a checkpoint after every this number of iterations
 _C.SOLVER.CHECKPOINT_PERIOD = 5000
 
-# Used with `hooks.BestCheckpointer`. Checkpoints every best model based on given eval metric.  
+# Used with `hooks.BestCheckpointer`. Checkpoints every best model based on given eval metric.
 _C.SOLVER.BEST_CHECKPOINTER = CN({"ENABLED": False})
 _C.SOLVER.BEST_CHECKPOINTER.METRIC = "bbox/AP50"
 
--- a/detectron2/engine/hooks.py
+++ b/detectron2/engine/hooks.py
@@ -245,11 +245,12 @@ class BestCheckpointer(HookBase):
         self.best_metric = None
         self.best_iter = None
 
-    def _update_best(self, val, iter):
+    def _update_best(self, val, iteration):
         if math.isnan(val) or math.isinf(val):
-            return
+            return False
         self.best_metric = val
-        self.best_iter = iter
+        self.best_iter = iteration
+        return True
 
     def _best_checking(self):
         metric_tuple = self.trainer.storage.latest().get(self._val_metric)
@@ -262,8 +263,7 @@ class BestCheckpointer(HookBase):
             latest_metric, metric_iter = metric_tuple
 
         if self.best_metric is None:
-            self._update_best(latest_metric, metric_iter)
-            if self.save_first:
+            if self._update_best(latest_metric, metric_iter) and self.save_first:
                 additional_state = {"iteration": metric_iter}
                 self._checkpointer.save(f"{self._file_prefix}", **additional_state)
                 self._logger.info(
--- a/detectron2/engine/hooks.py
+++ b/detectron2/engine/hooks.py
@@ -246,7 +246,13 @@ class BestCheckpointer(HookBase):
         self.best_iter = iter
     
     def _best_checking(self):
-        latest_metric, metric_iter = self.trainer.storage.latest().get(self._val_metric)
+        metric_tuple = self.trainer.storage.latest().get(self._val_metric)
+        if metric_tuple is None:
+            self._logger.warning(f'Given val metric {self._val_metric} does not seem to be computed/stored. Will not be checkpointing based on it.')
+            return
+        else:
+            latest_metric, metric_iter = metric_tuple
+        
         if self.best_metric is None:
             self._update_best(latest_metric, metric_iter)
         elif self._compare(latest_metric, self.best_metric):
--- a/detectron2/engine/hooks.py
+++ b/detectron2/engine/hooks.py
@@ -4,13 +4,14 @@
 import datetime
 import itertools
 import logging
+import math
 import os
 import tempfile
 import time
-import math
 from collections import Counter
 import torch
-from fvcore.common.checkpoint import PeriodicCheckpointer as _PeriodicCheckpointer, Checkpointer
+from fvcore.common.checkpoint import Checkpointer
+from fvcore.common.checkpoint import PeriodicCheckpointer as _PeriodicCheckpointer
 from fvcore.common.param_scheduler import ParamScheduler
 from fvcore.common.timer import Timer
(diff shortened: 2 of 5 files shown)


levan92:
Hmm sorry, not sure why there is a test failure, issue seems to stem from `detectron2.data.build.DatasetFromList` but that wasn't changed in this PR

detectron2/engine/hooks.py
```diff
@@ -199,6 +201,81 @@ def after_step(self):
         self.step(self.trainer.iter)
 
 
+class BestCheckpointer(HookBase):
+    """
+    Checkpoints best weights based off given metric.
+
+    This hook should be used in conjunction to and executed after `EvalHook`.
```
ppwwyyxx, line 208:
```suggestion
    This hook should be used in conjunction to and executed after the hook that produces the metric, e.g. `EvalHook`.
```

detectron2/engine/hooks.py
```diff
@@ -199,6 +201,81 @@ def after_step(self):
...
+        file_prefix: str = "model_best",
+        append_iter: bool = False,
+    ) -> None:
+        """
+        Args:
+            eval_period (int): the period `EvalHook` is set to run.
+            checkpointer: the checkpointer object used to save checkpoints.
+            val_metric (str): chosen validation metric to track for best checkpointing
+            file_prefix (str): the prefix of checkpoint's filename, defaults to "model_best"
+            append_iter (bool): if True, checkpoint file name will have iteration num appended, else write to the same checkpoint file.
```
ppwwyyxx, line 225: There is no harm to save iter (it's saved in all other checkpoints anyway). Let's use True and remove this option.
However I don't think we should add iter to the saved file name - it will then save __many__ best models instead of one.

detectron2/config/defaults.py
```diff
@@ -531,6 +531,10 @@
 # Save a checkpoint after every this number of iterations
 _C.SOLVER.CHECKPOINT_PERIOD = 5000
 
+# Used with `hooks.BestCheckpointer`. Checkpoints every best model based on given eval metric.
+_C.SOLVER.BEST_CHECKPOINTER = CN({"ENABLED": False})
+_C.SOLVER.BEST_CHECKPOINTER.METRIC = "bbox/AP50"
+
```
ppwwyyxx, line 537: please remove new config options as we expect users to add the hook to their training script directly

detectron2/engine/hooks.py
```diff
@@ -199,6 +201,81 @@ def after_step(self):
...
+                additional_state = {"iteration": metric_iter}
+                save_name = f"{self._file_prefix}_{metric_iter:07d}"
+            else:
+                save_name = f"{self._file_prefix}"
+                additional_state = {}
+            self._checkpointer.save(save_name, **additional_state)
+            if self.best_metric is None:
+                self._logger.info(
+                    f"Saved first model with latest eval score for {self._val_metric} at {latest_metric:0.5f}"
+                )
```
ppwwyyxx, line 249: Saving the first model looks useless. Could you change the behavior to not save?

detectron2/engine/hooks.py
```diff
@@ -199,6 +201,81 @@ def after_step(self):
...
+    Checkpoints best weights based off given metric.
+
+    This hook should be used in conjunction to and executed after `EvalHook`.
+    """
+
+    def __init__(
+        self,
+        eval_period: int,
+        checkpointer: Checkpointer,
+        val_metric: str = "bbox/AP50",
```
ppwwyyxx, line 215:
```suggestion
        val_metric: str,
```

detectron2/engine/hooks.py
```diff
@@ -199,6 +201,81 @@ def after_step(self):
...
+        file_prefix: str = "model_best",
+        append_iter: bool = False,
+    ) -> None:
+        """
+        Args:
+            eval_period (int): the period `EvalHook` is set to run.
+            checkpointer: the checkpointer object used to save checkpoints.
+            val_metric (str): chosen validation metric to track for best checkpointing
```
ppwwyyxx, line 223:
```suggestion
            val_metric (str): chosen validation metric to track for best checkpoint, e.g. "bbox/AP50"
```

detectron2/engine/hooks.py
```diff
@@ -199,6 +201,81 @@ def after_step(self):
...
 
 
+class BestCheckpointer(HookBase):
+    """
+    Checkpoints best weights based off given metric.
+
+    This hook should be used in conjunction to and executed after `EvalHook`.
+    """
+
+    def __init__(
```
ppwwyyxx, line 211: please add an option `mode="max"` similar to the one in https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/ModelCheckpoint, with two possible values "min" and "max"

detectron2/engine/hooks.py
```diff
@@ -199,6 +201,81 @@ def after_step(self):
...
+            if self.best_metric is None:
+                self._logger.info(
+                    f"Saved first model with latest eval score for {self._val_metric} at {latest_metric:0.5f}"
+                )
+            else:
+                self._logger.info(
+                    f"Saved best model as latest eval score for {self._val_metric} at {latest_metric:0.5f} is better than last best score at {self.best_metric:0.5f} @ {self.best_iter} steps"
+                )
+            if math.isnan(latest_metric):
+                latest_metric = -1.0
```
ppwwyyxx, line 255: a metric can be negative (e.g. log likelihood). Please handle NaNs/Infs more properly: check and never treat them as the best

ppwwyyxx (commented):
Thanks! looks pretty reasonable overall and please disregard the test failure about `DatasetFromList`

detectron2/engine/hooks.py
levan92: oh right, sorry! totally skipped my mind that the metric could go the other direction :facepalm:

detectron2/engine/hooks.py
levan92: I've modified it such that it only updates the `best_metric` during the "first model" but does not save it. 

However, I do have some lingering concerns for this -- should we just save it as long as the current metric value is valid (not NaN nor infs), even if it is the first model? Seems like it is implemented this way in Keras as well, as seen [here](https://github.com/keras-team/keras/blob/f43c792c129ffba27943c00f11c73c7d8db2838f/keras/callbacks.py#L1412). It initialises the best variable as either negative or positive infinity, then compares with current value and saves even if it is the "first call". We could implement that in a similar fashion here.

Kevin-Delnoije: Would it be a good idea to add a starting threshold as a keyword argument defaulted to None?
Often when I train models on my dataset I know what to expect from the metric and if it performs a lot worse then I don't want to save it.

levan92: @ppwwyyxx I do realise that sometimes saving the model after the first evaluation is needed. It is always possible the best model is reached by the first evaluation run, albeit not ideal, and especially if the first evaluation happens after many steps. 

I will pull this commit into this PR if you feel it is alright: https://github.com/levan92/detectron2/commit/59bebb201f59519e8db80f81c0ad2f06347053e0 It exposes a `save_first` argument which default to `False`.

ppwwyyxx: That argument makes total sense to me. We can let it be the default behavior (i.e. also save the first evaluation).

ppwwyyxx:
Could you add a unittest in `tests/test_engine.py` for the hook? 

In the unittest you can fake an evaluate function like what's done in `test_eval_hook()` in `tests/test_engine.py`, e..g:
```python
metrics = iter([0.3, 0.4, 0.35, 0.5])
hooks.EvalHook(20, lambda: {"metrics": next(metrics)})
```

levan92:
> Sorry for the delay. Could you add a unittest in `tests/test_engine.py` for the hook?
> 
> In the unittest you can fake an evaluate function like what's done in `test_eval_hook()` in `tests/test_engine.py`, e..g:
> 
> ```python
> metrics = iter([0.3, 0.4, 0.35, 0.5])
> hooks.EvalHook(20, lambda: {"metrics": next(metrics)})
> ```

Added in 935939b

ppwwyyxx (approved):
Thanks!

levan92:
Sorry @ppwwyyxx, found a minor bug which throws when the eval score is nan in the first evaluation. Just pushed the fix for it.

AKMourato:
Hi @levan92, can you give an example of how to use it with the new baselines configs?
Just adding:
```
hooks.BestCheckpointer(test_period, checkpointer, metric_name, mode=mode, save_first=save_first)])
```
is not sufficient I guess?
Thanks

levan92:
Hi @AKMourato, it should be sufficient if you're hardcoding the arguments/passing them in through other ways. 

However, if you want to control these through config files, you will need to do the additional work of adding them into your configs. I've included below an example of how I do it with the normal config style (yaml). But if you're talking about the new_baselines (which uses the "lazy" config style), it should be even easier, just that you may need to define your own `train` config dict (see `configs/common/train.py`) and modify `lazyconfig_train_net.py` accordingly.

Also, if you're doing multi-gpu training, you might want to condition it with a `if comm.is_main_process()`. 

In `train_net.py`:
```python3
def add_custom_configs(cfg: CfgNode):
    _C = cfg
    _C.SOLVER.BEST_CHECKPOINTER = CfgNode({"ENABLED": False})
    _C.SOLVER.BEST_CHECKPOINTER.METRIC = "bbox/AP50"
    _C.SOLVER.BEST_CHECKPOINTER.MODE = "max"
```

```python3
def setup(args):
    """
    Create configs and perform basic setups.
    """
    cfg = get_cfg()
    add_custom_configs(cfg)
    cfg.merge_from_file(args.config_file)
    cfg.merge_from_list(args.opts)
    cfg.freeze()
    default_setup(cfg, args)
    return cfg
```

Copy and override the `build_hooks` method of `DefaultTrainer` to include this after the `EvalHook` [line](https://github.com/facebookresearch/detectron2/blob/55c0078b1ffa6fca21834db7b3cf85c48d66109d/detectron2/engine/defaults.py#L457)
```python3
if cfg.SOLVER.BEST_CHECKPOINTER and comm.is_main_process():
    ret.append(hooks.BestCheckpointer(cfg.TEST.EVAL_PERIOD, self.checkpointer, cfg.SOLVER.BEST_CHECKPOINTER.METRIC, mode=cfg.SOLVER.BEST_CHECKPOINTER.MODE, save_first=True))
```

ZahraAnam:
Hi! @levan92 I tried this code from yours but it didn't save any model as best, did it work for you?

sjdch:
我是小白，想保存最佳的ap,ap50,ap75等信息，应该修改哪些文件，如何修改？谢谢您的回复

Issue #3225, opened 2021-07-06
Improve documentation concerning the new config files


orbiskcw:
### In short
**Concerning:** https://detectron2.readthedocs.io/en/latest/tutorials/configs.html
**Problem**: Documentation does not seem to have been updated to reflect the new config files (`.py` rather than `.yaml`) 
**Solution**: Update the documentation

### Problem description
FAIR recently published [new Mask R-CNN baselines](https://ai.facebook.com/blog/advancing-computer-vision-research-with-new-detectron2-mask-r-cnn-baselines/) and this was my first introduction to the new config file that no longer relies on YAML files but on 'raw' `.py` files. I am trying to load the new baselines using the config files mentioned in the `MODEL_ZOO` (see [this table](https://ai.facebook.com/blog/advancing-computer-vision-research-with-new-detectron2-mask-r-cnn-baselines/)). For example: 
```python
from detectron2 import model_zoo
model = model_zoo.get("new_baselines/mask_rcnn_regnetx_4gf_dds_FPN_400ep_LSJ.py", trained=True)
```

This gives
```
RuntimeError: new_baselines/mask_rcnn_regnetx_4gf_dds_FPN_400ep_LSJ not available in Model Zoo!
```

I have installed Detectron2 using the [installation instructions](https://detectron2.readthedocs.io/en/latest/tutorials/install.html#build-detectron2-from-source). When looking up [the documentation on configs](https://detectron2.readthedocs.io/en/latest/tutorials/configs.html), it seems that this has not been updated to reflect the new configs and still solely mentions YAML files. 

### Proposed solution
It could be that the `CONFIG_PATH_TO_URL_SUFFIX` dictionary in `_ModelZooUrls` class still has to be updated and that this is actually a bug (see [here](https://github.com/facebookresearch/detectron2/blob/a17b768895a1207c540e909225fa5a9389564927/detectron2/model_zoo/model_zoo.py#L12)), but I find it hard to estimate wheter this is meant behavior (i.e. the new config file should be loaded differently) or a bug due to my limited understanding of the new config files. Either way, I therefore feel like the documentation on readthedocs should be updated to reflect the change from `.yaml` to `.py`.

cwognum:
I think this is actually a bug 🐛

@ppwwyyxx Could you confirm whether this is a bug or expected behavior? I noticed you authored the last commit in the `model_zoo.py` file and made the changes there for the new `.py` config files.

VishalBalaji321:
[...]

PR #3228, opened 2021-07-07
Added the mapping from config to model checkpoint for the new baselines

cwognum:
Related to #3225, but does **not** solve it. 

# Overview of changes
FAIR recently published new [Mask R-CNN baselines](https://ai.facebook.com/blog/advancing-computer-vision-research-with-new-detectron2-mask-r-cnn-baselines/). The new config files have not yet been mapped to the corresponding model checkpoint files in the `model_zoo._ModelZooUrls` class. This was a relatively easy fix that simply follows the table given in the [MODEL_ZOO table](https://github.com/facebookresearch/detectron2/blob/master/MODEL_ZOO.md#new-baselines-using-large-scale-jitter-and-longer-training-schedule). If we run the following code snippet below:
```python
model = model_zoo.get("new_baselines/mask_rcnn_regnetx_4gf_dds_FPN_400ep_LSJ.py", trained=True)
```

This used to give: 
```
RuntimeError: new_baselines/mask_rcnn_regnetx_4gf_dds_FPN_400ep_LSJ not available in Model Zoo!
``` 

With these new changes, we get what we would expect and actually download and load the model: 
```
model_final_f1362d.pkl: 173MB [01:13, 2.37MB/s]                                                                                                                                                                                                                                
The checkpoint state_dict contains keys that are not used by the model:
  proposal_generator.anchor_generator.cell_anchors.{0, 1, 2, 3, 4}
```

I am not entirely sure whether the warning is expected or something that requires fixing.

# Additional proposal
I personally am not a big fan of the naming. "new_baslines" seems too general to me and deviates from the already established naming conventions. This makes it less intuitive and not very robust to future updates. I would suggest moving the configs to `configs/COCO-InstanceSegmentation`. Similarly, I would then also move the checkpoint files accordingly to `https://dl.fbaipublicfiles.com/detectron2/COCO-InstanceSegmentation/`. I did not implement these changes yet because I cannot change where the model files are stored and because I can imagine there to be an internal motivation not to change the naming anymore.

### DIFF
--- a/detectron2/model_zoo/model_zoo.py
+++ b/detectron2/model_zoo/model_zoo.py
@@ -48,6 +48,19 @@ class _ModelZooUrls(object):
         "COCO-InstanceSegmentation/mask_rcnn_R_101_DC5_3x": "138363294/model_final_0464b7.pkl",
         "COCO-InstanceSegmentation/mask_rcnn_R_101_FPN_3x": "138205316/model_final_a3ec72.pkl",
         "COCO-InstanceSegmentation/mask_rcnn_X_101_32x8d_FPN_3x": "139653917/model_final_2d9806.pkl",  # noqa
+        # New baselines using Large-Scale Jitter and Longer Training Schedule
+        "new_baselines/mask_rcnn_R_50_FPN_100ep_LSJ": "42047764/model_final_bb69de.pkl",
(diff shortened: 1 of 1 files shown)


cwognum:
The CI fails, but this has to do with `detectron2.data.build.DatasetFromList`. The error seems to also occur for the latest commit on the master branch.

PR #3237, opened 2021-07-09
Fix optional arguments for configuration

nijkah:
I found that "--opts", described in GETTING_STARTED.md, only works in demo/demo.py

``` python
# demo/demo.py
parser.add_argument(
        "--opts",
        help="Modify config options using the command-line 'KEY VALUE' pairs",
        default=[],
        nargs=argparse.REMAINDER,
)
```

But, default_argument_parser could not find "--opts" arguments. 
Because, it is defined as below.
``` python
# detectron2/engine/defaults.py
parser.add_argument(
        "opts",
        help="Modify config options using the command-line 'KEY VALUE' pairs",
        default=[],
        nargs=argparse.REMAINDER,
)
```
And I thought "opts" is not clear for standard python arguments, so I matched this.


nijkah:
I could understand that why '--' is removed from default_argument_parser after checking training command line examples.
So maybe it is due to allow compactness to set configurations for users.

I close this pr.

PR #3371, opened 2021-08-16
fix device error in BoxMode.convert()

csuhan:
Thanks for your contribution!


ppwwyyxx:
Thanks! Could you add a unittest?

There is a related test here but not sufficient https://github.com/fairinternal/detectron2/blob/00a7ccc817629f8ee43e249503c7fc5a9a16f164/tests/structures/test_boxes.py#L55

PR #3311, opened 2021-08-02
Minimal setup.py for DensePose project

pjbull:
## Summary

Currently DensePose cannot be installed and referenced in other code bases as a Python package. It would be useful to be able to install densepose as a package. 

This PR adds a minimal `setup.py` file that allows outside developers to include densepose as a dependency by running:
```
pip install git+https://github.com/drivendataorg/detectron2@master#subdirectory=projects/DensePose
```

It includes the extra dependencies that are not part of detectron2 as requirements and also installs `detectron2`. It also adds an `__init__.py` to the utils folder so that is importable when installed with pip.

## Open questions:
 1. Should we provide a version number? We currently do not so pip interprets the version as `0.0.0`.
 2. Currently the [package name is `DensePose`](https://github.com/facebookresearch/detectron2/pull/3311/files#diff-8ada752fb4e04a934789193ff59b1e77491221bfdb46f99638f7a0b00e29f2d2R17) to match the name of the project folder, but I think its more conventional to match the actual module and use `densepose` as the name. Is there a preference?
 3. Should we add these install instructions to the README?

-------------

## Example usage

Here is the full output log of a successful `pip install` from our branch into a new environment that only had `torch` installed:

```
❯ pip install git+https://github.com/drivendataorg/detectron2@densepose-add-setup-py#subdirectory=projects/DensePose

 ... // clipped install logs

[...]

PR #3382, opened 2021-08-19
Make vis module importable, remove required torch and torchvision

pjbull:
In PR #3311 we made densepose a pip-installable Python package. However, we forgot to include an `__init__.py` so that the `vis` module is importable. This PR adds that `__init__.py`.

It was also mentioned in that PR that there is a preference not to require torch and torchvision requirements versions:
https://github.com/facebookresearch/detectron2/pull/3311#issuecomment-892513481

This PR also removes those requirements from the `setup.py` and lists them instead in the prerequisites in the documentation.


pjbull:
@patricklabatut @vkhalidov Small updates here to follow up on #3311.

vkhalidov (approved):
Great, thanks for contributing this!

PR #3400, opened 2021-08-23
Fixed an RuntimeError with interference

harishsg99:
Thanks for your contribution!.
Fixed the following runtime Error : numpy() fails when visualizing an output element of a input batch during inference


harishsg99 (commented):
Commit works fine with my local deployment also

PR #3454, opened 2021-09-06
Fix reduction of cross_entropy

sukjunhwang:
Reduction of cross_entropy is always set to "mean".
Fixed to receive the reduction.


ppwwyyxx (approved):

ppwwyyxx:
Thanks for the fix!

PR #3469, opened 2021-09-12
Filter out a UserWarning in torch-1.9

Johnqczhang:
Filter out the following `UserWarning` triggered by the floor division in the original code when using pytorch-1.9.0. Replace with the [torch.div()](https://pytorch.org/docs/stable/generated/torch.div.html#torch.div) API as suggested.

```
UserWarning: floor_divide is deprecated, and will be removed in a future version of pytorch. It currently rounds toward 0 (like the 'trunc' function NOT 'floor'). This results in incorrect rounding for negative values.
To keep the current behavior, use torch.div(a, b, rounding_mode='trunc'), or for actual floor division, use torch.div(a, b, rounding_mode='floor'). (Triggered internally at  /opt/conda/conda-bld/pytorch_1623448224956/work/aten/src/ATen/na
tive/BinaryOps.cpp:467.)
```

Thanks for your contribution!

### DIFF
--- a/detectron2/structures/image_list.py
+++ b/detectron2/structures/image_list.py
@@ -96,7 +96,7 @@ class ImageList(object):
         if size_divisibility > 1:
             stride = size_divisibility
             # the last two dims are H,W, both subject to divisibility requirement
-            max_size = (max_size + (stride - 1)) // stride * stride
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
According to the CI this breaks pytorch1.7 support, so we cannot merge this for now.

PR #3540, opened 2021-09-30
Huggingface Gradio Web Demo

AK391:
easy to use web demo using gradio on huggingface spaces


ppwwyyxx:
Thanks for making this! The demo looks interesting but we don't think it should be a headline in the front page README. Maybe post it on twitter or github discussions to share the work.

PR #3549, opened 2021-10-03
add color field in instances and visualize based on preset colors

wenliangzhao2018:
Summary: Currently D2 (https://github.com/facebookresearch/detectron2/commit/11528ce083dc9ff83ee3a8f9086a1ef54d2a402f) and D2 (https://github.com/facebookresearch/detectron2/commit/11528ce083dc9ff83ee3a8f9086a1ef54d2a402f)GO uses a box IoU based method to assign id and color for each instance. It is useful but not flexible. This diff allows user to define their own preset colors for each instance. By doing this, some advanced tracking method can be applied before get into the Visualizer.

Differential Revision: D31366663


wenliangzhao2018:
The lint error is here: https://github.com/facebookresearch/detectron2/pull/3549/checks?check_run_id=3780584173. Not very clear to me what's wrong.

PR #3686, opened 2021-11-11
save grad_scaler state dict if use amp

xiaohu2015:
saving grad_scaler is better when resume training. Although the AMPTrainer implemented state_dict method,  in fact, the grad_scaler state_dict is not saved in checkpoints.

### DIFF
--- a/detectron2/engine/defaults.py
+++ b/detectron2/engine/defaults.py
@@ -390,6 +390,9 @@ class DefaultTrainer(TrainerBase):
             optimizer=optimizer,
             trainer=weakref.proxy(self),
         )
+        if cfg.SOLVER.AMP.ENABLED:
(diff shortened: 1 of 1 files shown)


detectron2/engine/defaults.py
```diff
@@ -390,6 +390,9 @@ def __init__(self, cfg):
             optimizer=optimizer,
             trainer=weakref.proxy(self),
         )
+        if cfg.SOLVER.AMP.ENABLED:
+            self.checkpointer.add_checkpointable('grad_scaler', self._trainer.grad_scaler)
+
```
ppwwyyxx, line 395: To solve checkpointing issues for both optimizer and grad_scaler, I think it's better to implement a correct `state_dict()` for `DefaultTrainer`.

It can be implemented by merging `super().state_dict()` and `self._trainer.state_dict()`.

ppwwyyxx (approved):

Issue #3404, opened 2021-08-25
Visualizations on W&B


AyushExel:
## 🚀 Feature
Allow visualization of training progress using media panels and tables. 

## Motivation & Examples
This is based on this [issue](https://github.com/facebookresearch/detectron2/issues/774). 
I'm an engineer at W&B and I've been working with object detection tasks. I regularly use some of the following visualizations on W&B dashboards. Would the following features be useful to other detection users? I'd like to know what the maintainers think.
### **Bounding box & Segmentation maps debugger**

W&B supports interactive media panels, where you can track how training progresses by adjusting the steps, confidence scores, and classes of predictions in real-time.
![ezgif com-gif-maker (2)](https://user-images.githubusercontent.com/15766192/130792944-77c59dbd-861b-4161-9c62-5f71e44bac7e.gif)

Try it live [here](https://wandb.ai/cayush/ert/runs/xc4m2xip?workspace=user-cayush)

The Media panel also supports debugging segmentation maps. 
![semantic segmentation (1)](https://user-images.githubusercontent.com/15766192/130792997-887b706c-13b2-4f18-8e7b-10ded88debfa.gif)

### Dataset visualization and versioning

With W&B tables, you can visualize, query, and filter your datasets in your browser.

Quickly compare results across different training epochs, datasets, hyperparameter choices, model architectures etc. For example, [take a look at this comparison of two models on the same test images →](https://wandb.ai/stacey/evalserver_answers_2/artifacts/results/eval_Daenerys/c2290abd3d7274f00ad8/files/eval_results.table.json#b6dae62d4f00d31eeebf$eval_Bob)

<img width="1142" alt="Screen Shot 2021-04-29 at 8 55 25 PM (1)" src="https://user-images.githubusercontent.com/15766192/130793194-eb6a77f3-11df-42a5-ae30-9f9da7b36bcf.png">

**Model versioning and DAGs**
![ezgif com-gif-maker (1) (1)](https://user-images.githubusercontent.com/15766192/130798266-0647a420-f5bd-442e-b556-d022f65fbdf3.gif)

We can have the user set 

- the model logging period
- and the desired metric they'd like to optimize.

Based on that, we'll log models after every  `model logging period` with alias `best` if the current model performs best on the `desired metric`

AyushExel:
@ppwwyyxx what do you think about this? Would this be a useful add-on?

nbardy:
Would be useful for me! I ended up adding it myself to debug some models.

[...]

PR #3716, opened 2021-11-23
[WIP] PoC for Periodic Prediction and Weights and Biases Integrations

AyushExel:
Issue https://github.com/facebookresearch/detectron2/issues/3404
As discussed before in the issue, this is a proof of concept for W&B integration. This is a design proposal, the exact implementation is subject to change
## This PR adds:
* `PeriodicPredictor`: When initialized, performs prediction on the specified split of the dataset and stored it in the event storage. Tries to predict on test set but uses train set if test set is not defined. `PeriodicPredictor` is designed to be independent of W&B and can be used by a Hook/Writer for interacting with predictions.
* `WandbWriter`: W&B plugin to manage experiment metrics, visualizations, and debugging. The tables integration allow you to inspect the model performance on each class across iterations. All **images are deduped** which means the data is only logged once. **Supports DDP**

Here are some Example runs Generated from this PR:
* Detection - [Dashboard](https://wandb.ai/cayush/Detectron2-example/runs/jmwxl592?workspace=user-cayush), [Tables view](https://wandb.ai/cayush/Detectron2-example/artifacts/run_table/run-jmwxl592-coco_2017_val_0/9d40c918f356449f4572/files/coco_2017_val_0.table.json)
* Semantic segmentation(PointRend) - [Dashboard](https://wandb.ai/cayush/Detectron2-example/runs/2cqsz4bx?workspace=user-cayush) [Tables view](https://wandb.ai/cayush/Detectron2-example/artifacts/run_table/run-2cqsz4bx-cityscapes_fine_sem_seg_train_0/d5ed070992749df180c2/files/cityscapes_fine_sem_seg_train_0.table.json)
Notes:
* All media logged are interactive
* We can expand interactive media support for others tasks like panoptic segmentation.
* Both the periodic predictor and W&B writer are independent, i.e, the imaged will be logged only if the user initializes the predictor. Otherwise, only metrics will be logged

## Some examples from dashboard view:
* ## Tables
Inspect, filter, query, and compare your model prediction to better debug
ezgif com-gif-maker (1)

* ## Media Panel
ezgif com-gif-maker

cc @ppwwyyxx could you please look through the design and see if this is something you'd be interested in?


ppwwyyxx:
Thanks for the effort! I took a brief look and here are some high-level comments:

1. CfgNode is not going to support new features per https://github.com/facebookresearch/detectron2/blob/dc2a1dcd286f81e866b05f019cacdbc33c85f51c/detectron2/config/defaults.py#L4-L6 
, so we'll not add new keys to it. Users are expected to use code (or the new config system which is also mostly code) to directly access new features, e.g. by calling `trainer.register_hook` with a wandb hook, etc.

For the same reason, please don't add any class/functions that takes cfg as an argument. Let them take the actual list of arguments explicitly instead. This way the [new config system](https://detectron2.readthedocs.io/en/latest/tutorials/lazyconfigs.html) can call them directly.

2. As mentioned in https://github.com/facebookresearch/detectron2/issues/3404#issuecomment-954566262 it seems easier to produce prediction visualizations in a custom evaluator: it has an easy API that directly gives you all inference inputs & outputs. Compared to the new hook in this PR, using an evaluator:
* avoids creating a new dataloader by yourself. the existing eval logic can directly apply 
* can share the inference cost with another evaluator (e.g. those that compute metrics)

The end user's code may then look like:
```
inference_on_dataset(
  model, my_dataloader, 
  [OtherEvaluator(), WandbVisualizer(arg1, arg2, ...)]
)
```

3. How data is written: now predictions are saved in `EventStorage`, similar to how we are now handling "put_image". However this is actually a design problem that we attempt to fix: we hope to avoid the pattern where `EventStorage` becomes a global dumpground where one piece of code stores some arbitrary data, and another piece of code takes it out. We haven't solved this issue yet, which is why `put_image` is still the way it is.  

    However for wandb, it seems you can directly call `wandb.log...` to log the predictions to wandb when they are available (e.g. inside the evaluator), without having to save them to a `EventStorage`. So it seems the below would work (correct me if I'm wrong:
```
class WandbVisualizer(DatasetEvaluator):
   def process(inputs, outputs):
       image = create_vis(inputs, outputs)
       wandb.log(image)
```

additionally, to save the scalars that we're now saving to tensorboard, the wandb hook like what's in this PR can still be used, e.g. like:
```
class WandbMetrics(EventWriter):
   def write(self):
       for k, v in storage.latest_scalars():
           wandb.log(k=v)
```

Feel free to let me know what you think of the above.

AyushExel:
@ppwwyyxx thanks for the quick response. Appreciate it.
1. Got it. It should be simple to remove dependency from cfgNode. Will do
2. Sounds interesting. I'll have to look in detail as I'm less familiar with plugging into Evaluator class. I'll reach out after going through some examples.
3. Sounds good. We won't need to dump stuff into EventStorage if 2. is done right as you said. Btw, the current implementation also handles and tb scalars as you mentioned.

AyushExel:
@ppwwyyxx Some initial thoughts on being able to share the inference cost. I was thinking of this approach
* In case user doesn't want to perform evaluation(hasn't defined the test set) but wants to log predictions as set intervals:
      if user sets `log_prediction` then `inference_on_dataset` can be called in the `WandbWriter` logger.

* But in case we want to share the cost when the test set is defined, things might get more complicated as `inference_on_dataset` is called inside the [test function ](https://github.com/facebookresearch/detectron2/blob/main/detectron2/engine/defaults.py#L617). Are there any plans to incorporate an optional argument with a default test function that allows user to pass a Visualizer? Or if there's any other clean workaround.
I imagine something like this.
```
if visualizer is not None:
    evaluator_list.append(visualizer)
results_i = inference_on_dataset(model, data_loader, evaluator_list)
```

AyushExel:
@ppwwyyxx  can you take a look again, please? Thanks for your suggestion, I like the `DatasetEvaluator` version better. 
I still have to update docstrings etc, but just wanted to check if I'm in the right direction. Usage is mentioned below
### Usage:
```
class MyTrainer(DefaultTrainer):
     # if you want to visualize 100 instances of test set
    @classmethod
    def build_evaluator(cls, cfg, dataset_name, output_folder=None):
        return [WandbVisualizer(dataset_name, size=100)]

    # if you want to track expeirment metrics/configs/media
    def build_writers(self):
        writers =  super().build_writers()
        writers.append(WandbWriter(config=self.cfg))
        return writers
```

Misc question:
* Is it possible to only run Evaluation on a subset of a dataset? The motivation for this - what if I'm only using `WandbVisualizer` as an evaluator and I know ahead of time that only the first 100 batches are required.

ppwwyyxx:
Thanks and the new implementation and usage looks pretty clean and straightforward. I'll provide more detailed comments once you're ready.

detectron2/utils/events.py
```diff
@@ -484,3 +492,41 @@ def clear_histograms(self):
...
+        """
+        if wandb is None:
+            raise Exception("WandbWriter requires wandb. please install using `pip install wandb`")
+
+        self._window_size = window_size
+        self._run = wandb.init(
+            project=project,
+            config=config,
+            **kwargs
+        )
```
ppwwyyxx, line 519: Like evaluator I think it also needs to check if `wandb.run` is already set? Maybe user already call `wandb.init` before training - that seems to be a common pattern if I'm not mistaken

detectron2/utils/visualizer.py
```diff
@@ -1228,3 +1233,311 @@ def get_output(self):
...
+        if self._evalset_table is None:
+            self._evalset_table = self._build_evalset_table(parsed_output.keys())
+
+
+    def evaluate(self):
+        comm.synchronize()
+        table_rows = comm.gather(self._evalset_table_rows)
+        if comm.is_main_process():
+            if self._run is None:
+                self._run = wandb.init(project=self.dataset_name) if not wandb.run else wandb.run
```
ppwwyyxx, line 1280: where will it log to when `not wandb.run` ?
If there isn't a reasonable place to log to, we shouldn't just create a random directory or some random project.

We could in `__init__` check if `wandb.run` is set, and raise an error. That means users are responsible for calling `wandb.init` or `WandbWriter` before using this evaluator.

An alternative is to let this class accept kwargs as well like `WandbWriter` - but then I think the semantics becomes unclear when both classes are used: one of them (not sure which) will actually do init and the other will completely ignore the kwargs.

detectron2/utils/events.py
```diff
@@ -6,11 +6,19 @@
...
 
 from detectron2.utils.file_io import PathManager
+from detectron2.config import CfgNode
+
+try:
+    import wandb
+
+    assert hasattr(wandb, '__version__')  # verify package import not local dir
+except (ImportError, AssertionError):
+    wandb = None
```
ppwwyyxx, line 21: "Only import when needed" is how we deal with optional dependencies. So this better goes into the functions below (and there is no need to catch error then) .

detectron2/utils/events.py
AyushExel: Oh yep. Good catch. Thanks!

detectron2/utils/visualizer.py
AyushExel: How about instead of all kwargs we let the `WandbVisualizer` take an argument `output_dir` like `COCOEvaluator` class and use that as output folder if wandb isn't explicitly initialized by the user?

detectron2/utils/events.py
AyushExel: sounds good. will do

detectron2/utils/visualizer.py
ppwwyyxx: That also makes sense, but are you proposing to do this for both `WandbVisualizer` and `WandbWriter`, or just one? Consistency of usage is important and I can't see why one class accepts `**kwargs` for `wandb.init`, while another accepts `output_dir`.

AyushExel: Yeah, fair point. 
So, I guess the only option that remains is to make sure a wandb run is initialized manually when Visualizer is used. We cannot rely on `WandbWriter`'s initialization because `build_writers` is called after evaluators are initialized.

AyushExel: Yeah that's fair enough.
I changed the design so the visualizer won't call wandb.init. I relies on user either calling `wandb.init` manually or explicitly passing `wandbWriter`
```
 if comm.is_main_process():
       if self.wandb.run is None:
           raise Exception("wandb run is not initialized. Either call wandb.init(...) explicitly or set WandbWriter()")
       self._run = self.wandb.run
```

detectron2/utils/events.py
AyushExel: this is done

detectron2/utils/events.py
AyushExel: done!

detectron2/utils/visualizer.py
AyushExel: @ppwwyyxx in case you missed ^^^

ppwwyyxx (commented):
Thanks! The design looks good in general. Please polish with more docs and ideally a lightweight unittest and we should be good to go.

You could just test the two classes individually with some fake data. There is a tiny dataset available during unittest, that you can setup with https://github.com/facebookresearch/detectron2/blob/main/datasets/prepare_for_tests.sh . Our CI config (.circleci/config.yml, .github/workflows/workflow.yml) will also need to install wandb.

detectron2/utils/visualizer.py
```diff
@@ -1228,3 +1232,313 @@ def get_output(self):
...
+        self._run.use_artifact(eval_art)
+        eval_art.wait()
+        self._evalset_table_ref = eval_art.get(self.dataset_name).data
+
+    def _table_logging(self):
+        """
+        This function returns true if user defined settings enable tables logging implicitly or explicitly
+        """
+        return self.cfg.WANDB.LOG_PREDICTION
+
```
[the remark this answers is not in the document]
AyushExel: yep. Will remove

detectron2/utils/visualizer.py
```diff
@@ -1228,3 +1232,313 @@ def get_output(self):
...
+        E.g.
+        # set the properties separately
+        MetadataCatalog.get("val1").property = ...
+        MetadataCatalog.get("val2").property = ...
+
+        Builds 3 metadata objects for stuff and thing classes:
+        1. self.thing_class_names/self.stuff_class_names -- List[List[str]] of category names
+        2. self.thing_index_to_class/self.stuff_index_to_class-- List[Dict[int, str]]
+        3. self.thing_class_names/self.stuff_class_names -- List[wandb.Classes] for logging images to wandb.Table
```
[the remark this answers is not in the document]
AyushExel: will do

detectron2/utils/visualizer.py
```diff
@@ -1228,3 +1232,313 @@ def get_output(self):
             to the image.
         """
         return self.output
+
+
+class WandbVisualizer(DatasetEvaluator):
+
+    def __init__(self, dataset_name, size = 50) -> None:
+        super().__init__()
```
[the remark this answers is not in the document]
AyushExel: will do

detectron2/utils/visualizer.py
```diff
@@ -1228,3 +1232,313 @@ def get_output(self):
             to the image.
         """
         return self.output
+
+
+class WandbVisualizer(DatasetEvaluator):
```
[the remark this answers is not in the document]
AyushExel: will do

detectron2/utils/visualizer.py
```diff
@@ -1228,3 +1232,313 @@ def get_output(self):
...
+
+        returns:
+            Dict (): parsed predictions
+        """
+        parsed_pred = {}
+        if pred.get("instances") is not None:
+            pred_ins = pred["instances"]
+            parsed_pred["boxes"] = (
+                pred_ins.pred_boxes.tensor.tolist()[:10] if pred_ins.has("pred_boxes") else None
+            )
```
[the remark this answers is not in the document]
AyushExel: Hmm yeah. According to you what would be a good number? I guess by default models output 100 predictions?
I set it to 10 for 2 reasons:
1. 100 predictions seem too much for 1 img in the UI
2. more importantly, gathering large tensors( `comm.gather` ) might slow down. 
what do you think?

tests/tooling/test_wandb.py
```diff
@@ -0,0 +1,83 @@
...
+
+        visualizer = WandbVisualizer("coco_2017_val_100")
+        class CustomTrainer(DefaultTrainer):
+            @classmethod
+            def build_evaluator(cls, cfg, dataset_name, output_folder=None):
+                return [visualizer]
+
+        cfg = get_cfg()
+        cfg.MODEL.DEVICE = "cpu"
+        cfg.merge_from_file(model_zoo.get_config_file("COCO-Detection/faster_rcnn_R_50_C4_1x.yaml"))
```
AyushExel, line 71: @ppwwyyxx I'm trying to write a test for class `WandbVisualizer(DatasetEvaluator)`. How do you generally mock these models? I was testing using this model config but I suppose this is too large for testing purposes and will cause OOM errors. 
Any toy model that mocks bounding box output should work as the integration currently supports only detection and semantic segmentation

ppwwyyxx: We don't have a mechanism for mocked models. I think it's easier to call `visualizer.process(input, output)` manually with some fake prediction data.

detectron2/utils/visualizer.py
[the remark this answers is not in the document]
ppwwyyxx: 1. What we typically do in visualization is to show all predictions with confidence score > 0.7. Showing all predictions on one image is indeed very noisy in the UI and 0.7 is an empirical threshold we use. It can certainly be a configurable argument as well.

2. I think speed isn't an important factor here as predictions are pretty small tensors, and there is already a limit on total #images as `self.size`.

tests/tooling/test_wandb.py
AyushExel: @ppwwyyxx can you please point me to any such fake data being used in the rep?

detectron2/utils/visualizer.py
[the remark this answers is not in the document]
AyushExel: Interesting. 1. sounds good I'll test with this.
2. I was testing with some cityscapes semantic segmentation tasks and gathering data across GPUs got very slow. And there was a warning from `comm.gather` something like `gathering large data 1.x GB `

tests/tooling/test_wandb.py
ppwwyyxx: I meant just manually write down some fake values, e.g. like this: https://github.com/facebookresearch/detectron2/blob/490dfb9132a64945e210fd3b552e2e3221a13b70/tests/data/test_coco_evaluation.py#L24 or https://github.com/facebookresearch/detectron2/blob/490dfb9132a64945e210fd3b552e2e3221a13b70/tests/test_visualizer.py#L40-L61

ppwwyyxx: Also the dataset "coco_2017_val_100" is automatically registered and available during CI's unittests. It provides some real images and annotations if you need (get them by `DatasetCatalog.get("coco_2017_val_100")`). It can be downloaded by https://github.com/facebookresearch/detectron2/blob/main/datasets/prepare_for_tests.sh .

AyushExel: Thanks. This was very helpful. Tests are passing now

AyushExel:
@ppwwyyxx hey the tests are passing. Please have a look when you have some time :)

AyushExel:
@zhanghang1989 I've fixed the linting errors. Please pull

detectron2/evaluation/wandb.py
```diff
@@ -0,0 +1,320 @@
...
+    """
+    WandbVisualizer logs the model predictions over time in form of interactive W&B Tables.
+    Sopported tasks: Bounding Box Detection, Semantic Segmentation
+    """
+
+    def __init__(self, dataset_name, size=50) -> None:
+        """
+        Args:
+            dataset_name (str): Name of the registered dataset to be Visualized
+            size (int): Maximum number of data-points/rows to be visualized
```
ppwwyyxx, line 18:
```suggestion
            dataset_name (str): name of the registered dataset to be visualized
            size (int): maximum number of samples/rows to be visualized
```

ppwwyyxx (commented):
Added some small comments. Otherwise looks good to me. Thanks!

detectron2/evaluation/wandb.py
```diff
@@ -0,0 +1,320 @@
...
+
+        returns:
+            Dict (): parsed predictions
+        """
+        parsed_pred = {}
+        if pred.get("instances") is not None:
+            pred_ins = pred["instances"]
+            parsed_pred["boxes"] = (
+                pred_ins.pred_boxes.tensor.tolist()[:10] if pred_ins.has("pred_boxes") else None
+            )
```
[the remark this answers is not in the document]
AyushExel: Okay making this change. I'll run some manual tests for semantic segmentation masks later when we work on supporting instance/panoptic segmentation. Currently, I've removed the limit of 10 masks

detectron2/evaluation/wandb.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def process(self, inputs, outputs):
+        if self.size > 0 and len(self._evalset_table_rows) >= self.size // (comm.get_world_size()):
+            return
+        parsed_output = self._parse_prediction(outputs[0])
+        parsed_output["file_name"] = inputs[0].get("file_name")
+        table_row = self._plot_table_row(parsed_output)
+        self._evalset_table_rows.append(table_row)
+        if self._evalset_table is None:
+            self._evalset_table = self._build_evalset_table(parsed_output.keys())
```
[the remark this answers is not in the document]
AyushExel: sounds good

detectron2/evaluation/wandb.py
```diff
@@ -0,0 +1,320 @@
...
+                "class_labels": self.stuff_index_to_class[loader_i],
+            }
+        '''
+
+        table_row = [
+            file_name,
+            self.wandb.Image(file_name, boxes=boxes, masks=masks, classes=classes),
+        ]
+        if pred.get("boxes") is not None:
+            table_row.extend(avg_conf_per_class)
```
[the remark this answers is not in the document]
AyushExel: @ppwwyyxx  Thanks for the insight. I guess It'll be a UI problem to display conf. each prediction of each class as unique value so I used this less useful metric. At the most, it can provide info if a model is doing bad in a particular class. 
It looks like [this](https://wandb.ai/cayush/pr_review/reports/Shared-panel-22-01-05-14-01-11--VmlldzoxNDA1ODAx) in a workspace. 
Do you think we can include some other metrics to make this table more useful for practitioners/researchers?

detectron2/evaluation/wandb.py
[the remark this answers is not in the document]
AyushExel: the new logic checks the last set of images and predictions. That should be fine right? I'm not sure if checking the entire dataset prediction would be the right approach

AyushExel:
Hey, @ppwwyyxx sorry to ping you again. Just wanted to confirm that there's nothing remaining from our side on this. Thanks!

AyushExel:
@ppwwyyxx @zhanghang1989 any updates on this? Are there any blockers?

zhanghang1989:
@wat3rBro I think it is a good feature for users. What're your thoughts?

AyushExel:
@wat3rBro @ppwwyyxx @zhanghang1989 Hey guys, just wanted to check in and see why this is blocked and what we can do to unlock it. I'd be happy to address any feedback on this. 
Based on previous review comments, I was under the impression that this would get merged and I don't know what changed. 
Is there any chance that this PR is still under review and will get merged or should I close this ?

NingYuanLin:
I think it's a good feature and I really need it. I hope the development team will consider and agree to the pull request.

BenSpex:
@zhanghang1989 it would be great to merge this

PR #3770, opened 2021-12-08
Fix rrpn proposal validation

nijkah:
https://github.com/facebookresearch/detectron2/blob/08f617cc2c9276a48e7e7dc96ae946a5df23af3f/detectron2/modeling/proposal_generator/proposal_utils.py#L104
```python
        if not valid_mask.all():
            if training:
                raise FloatingPointError(
                    "Predicted boxes or scores contain Inf/NaN. Training has diverged."
                )
            boxes = boxes[valid_mask]
            scores_per_img = scores_per_img[valid_mask]
            lvl = lvl[valid_mask]
```

Unlike function **find_top_rpn_proposals**, **find_top_rrpn_proposals** does not include valid masking for level indices.
When the program goes into if condition, it gives an undesirable error says that "shape does not match".

I add a handling code same with **find_top_rpn_proposals**.

### DIFF
--- a/detectron2/modeling/proposal_generator/rrpn.py
+++ b/detectron2/modeling/proposal_generator/rrpn.py
@@ -92,17 +92,23 @@ def find_top_rrpn_proposals(
     for n, image_size in enumerate(image_sizes):
         boxes = RotatedBoxes(topk_proposals[n])
         scores_per_img = topk_scores[n]
+        lvl = level_ids
+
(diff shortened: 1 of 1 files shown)


ppwwyyxx (approved):

nijkah:
@ppwwyyxx 
Is there something that I have to check for merging?

I couldn't find any information for final decision of Merge Pull Request.

ppwwyyxx:
@zhanghang1989 Could you merge this bug fix (assuming tests will pass)? Thanks!

PR #3851, opened 2022-01-03
Bug fix: Handle empty instances in FCOS.

kdexd:
### 🐛 One line bug info: [FCOS implementation in current master](https://github.com/facebookresearch/detectron2/blob/31ec19b3132a3ac609600802dd37b2b40a76b5c9/detectron2/modeling/meta_arch/fcos.py) is unable to handle empty instances. 

This bug went unnoticed because: (a) images with empty instances are usually filtered while loading COCO annotations, and (b) FCOS should not encounter empty instances with its default training hyper-parameters. However, if I switch to a more aggressive large-scale jitter (LSJ) cropping augmentation, the model may encounter an image crop without any boxes in it. This crashes training abruptly due to a size mismatch in the pairwise anchor matching matrix.

### Bug fix

This PR lets FCOS handle empty instances by adding a dummy `[0, 0, 0, 0]` box labeled as background (ID = `num_classes`), similar to how it is handled in `RetinaNet` class. Training FCOS with LSJ augmentation does not crash anymore.

**Additional refactor:** While I was working on a fix, I noticed some inconsistencies in variable representations and naming conventions. For example, the `pairwise_match` variable was a `(R: anchor points, M: GT boxes)` matrix — this is a transposed representation of what a `match_quality_matrix` represents in `RetinaNet` and `GeneralizedRCNN`, `(M: GT boxes, R: anchor points)`. I refactored the code to make it more uniform with D2 conventions and make FCOS logic flow similar to RetinaNet, given that RetinaNet was the primary baseline in FCOS paper. My changes include:
  - Refactoring `pairwise_match` to a `match_quality_matrix` with its representation consistent with the rest of meta architectures. Moreover, variables renaming like (`matched_boxes` –> `matched_gt_boxes`, `label` —> `gt_labels`, and `gt_index` —> `matched_indices`) make the code more consistent with the naming convention in rest of meta architectures. 
  - Update: After @ppwwyyxx , I added this explanation as a comment instead of refactoring code: ~Use a `Matcher` instead of simply doing `pairwise_match.max()`. Original code was replacing indices of unmatched anchors as `-1` and accessing GT labels/boxes by doing `gt_index.clip(0)`, which felt non-ideal.~
  - Change the internal method `match_anchors` to compute and return per-instance `match_quality_matrix` (similar to using `pairwise_iou` in R-CNN). This modifies the old behavior of returning `matched_indices`. Since this method is used internally in `FCOS`, I renamed the method to `_match_anchors`.

### Any API changes?

The call signature of `FCOS` is unchanged. `FCOS.match_anchors()` is now `FCOS._match_anchors()` with a different return value, but it was only used internally by `FCOS.label_anchors()`.

### Verification

I verified my changes by one full training run of FCOS with ResNet-50-FPN on COCO detection, using the builtin config. The validation curves overlap very closely (orange: `master` branch, blue: with my changes).

image

I should have fixed the training seed in both runs... so additionally, I manually checked the equality of `pairwise_match.tranpose(0, 1)` and my `match_quality_matrix`for first 20 iterations by fixing `train.seed = 0` in config. Everything matches exactly.

### DIFF
--- a/detectron2/modeling/meta_arch/fcos.py
+++ b/detectron2/modeling/meta_arch/fcos.py
@@ -4,7 +4,7 @@ import logging
 from typing import List, Optional, Tuple
 import torch
 from fvcore.nn import sigmoid_focal_loss_jit
-from torch import Tensor, nn
+from torch import nn
 from torch.nn import functional as F
 
 from detectron2.layers import ShapeSpec, batched_nms
@@ -239,9 +239,7 @@ class FCOS(DenseDetector):
         }
 
     def compute_ctrness_targets(
-        self,
-        anchors: List[torch.Tensor],
-        gt_boxes: List[torch.Tensor],
+        self, anchors: List[Boxes], gt_boxes: List[torch.Tensor]
     ):
         anchors = Boxes.cat(anchors).tensor  # Rx4
         reg_targets = [self.box2box_transform.get_deltas(anchors, m) for m in gt_boxes]
@@ -256,7 +254,10 @@ class FCOS(DenseDetector):
         return torch.sqrt(ctrness)
 
     def forward_inference(
-        self, images: ImageList, features: List[Tensor], predictions: List[List[Tensor]]
+        self,
+        images: ImageList,
+        features: List[torch.Tensor],
+        predictions: List[List[torch.Tensor]],
     ):
         pred_logits, pred_anchor_deltas, pred_centerness = self._transpose_dense_predictions(
             predictions, [self.num_classes, 4, 1]
@@ -281,8 +282,8 @@ class FCOS(DenseDetector):
     def inference_single_image(
         self,
         anchors: List[Boxes],
-        box_cls: List[Tensor],
-        box_delta: List[Tensor],
+        box_cls: List[torch.Tensor],
+        box_delta: List[torch.Tensor],
         image_size: Tuple[int, int],
     ):
         """
--- a/detectron2/modeling/meta_arch/fcos.py
+++ b/detectron2/modeling/meta_arch/fcos.py
@@ -8,6 +8,7 @@ from torch import Tensor, nn
 from torch.nn import functional as F
 
 from detectron2.layers import ShapeSpec, batched_nms
+from detectron2.modeling.matcher import Matcher
 from detectron2.structures import Boxes, ImageList, Instances, pairwise_point_box_distance
 from detectron2.utils.events import get_event_storage
 
@@ -18,7 +19,6 @@ from .dense_detector import DenseDetector
 from .retinanet import RetinaNetHead
 
 __all__ = ["FCOS"]
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


xiaohu2015:
@kdexd hi，I also fix this bug https://github.com/facebookresearch/detectron2/pull/3910, but different with you.

detectron2/modeling/meta_arch/fcos.py
```diff
@@ -96,77 +102,73 @@ def forward_training(self, images, features, predictions, gt_instances):
...
-    @torch.no_grad()
-    def label_anchors(self, anchors, gt_instances):
-        """
-        Same interface as :meth:`RetinaNet.label_anchors`, but implemented with FCOS
-        anchor matching rule.
+            # Get matches and their labels using match quality matrix.
+            match_quality_matrix = match_quality_matrix.to(torch.float32) * (
+                1e8 - gt_areas[:, None]
+            )
+            matched_idxs, anchor_labels = self.anchor_matcher(match_quality_matrix)
```
ppwwyyxx, line 155: Though in general I like showing that RetinaNet and FCOS are very similar, I hope to not achieve this by complicating FCOS. After all it does deserve the credit of simplicity. 

Personally I think the introduction of anchor/box_transform is not complicating FCOS because it does make many RetinaNet code reusable, and make this implementation easier to understand than its original code IMHO. But introducing an extra concept of "anchor matcher" that can be achieved with a trivial one-liner is unnecessarily complicating the code.

Perhaps it's better to keep the original two lines here, but with a comment saying it is equivalent to applying `Matcher(1e-5, [0, 1])`. This way we keep the simplicity of FCOS while also drawing the connection to RetinaNet.

ppwwyyxx (commented):
Thanks for fixing the bug and also bringing the style of FCOS closer to RetinaNet!

There are some unittests (`test_empty_data`) in `test_model_e2e.py` that check if model supports empty instances. Could you add FCOS there? Just
```
class FCOSE2ETest(InstanceModelE2ETest, unittest.TestCase):
    CONFIG_PATH = "COCO-Detection/fcos....py"
```
is probably sufficient.

kdexd:
I also decided to add the internal method as `FCOS._match_anchors` which is only called inside `FCOS.label_anchors`. I am extending FCOS to do instance segmentation in my current project, and I need to override `FCOS.label_anchors` to return matched masks as well — it results in a lot of code duplication for computing `match_quality_matrix`. It is worth having a callable that returns a `match_quality_matrix`, analogous to `pairwise_iou` in R-CNN. Let me know what you think.

Since this will end up in `master`, I have started a training job to verify numbers (job same as that in my main comment). I will update this thread when it converges!

kdexd:
My tests are failing because of two reasons —

1. Linting tests are failing on files that are outside this PR.
2. Newly added FCOS tests are failing because they do not directly support reading lazy configs (if I understood the logs correctly)?

@ppwwyyxx: what is your suggestion? I am brute-forcing as many solutions as I can:
- Add YAML config for FCOS
- Add Lazy config handling in tests in this PR.
- Run FCOS with forced empty boxes / LSJ augmentation and post behavior here; work on tests in separate PR.
- ..? 🤔

ppwwyyxx:
Yeah please feel free to ignore linter :cry: 

This should let it support .py configs:
```
--- i/detectron2/utils/testing.py
+++ w/detectron2/utils/testing.py
@@ -4,6 +4,7 @@ import numpy as np
 import torch

+from detectron2.config import CfgNode, instantiate
 from detectron2.data import DatasetCatalog
 from detectron2.data.detection_utils import read_image
 from detectron2.modeling import build_model
@@ -21,9 +22,12 @@ def get_model_no_weights(config_path):
     Like model_zoo.get, but do not load any weights (even pretrained)
     """
     cfg = model_zoo.get_config(config_path)
-    if not torch.cuda.is_available():
-        cfg.MODEL.DEVICE = "cpu"
-    return build_model(cfg)
+    if isinstance(cfg, CfgNode):
+        if not torch.cuda.is_available():
+            cfg.MODEL.DEVICE = "cpu"
+        return build_model(cfg)
+    else:
+        return instantiate(cfg.model)
```

ppwwyyxx:
Thanks, the tests have passed. (macos + pytorch1.9 failure will be fixed in #3909)

ppwwyyxx (approved):

detectron2/modeling/meta_arch/fcos.py
```diff
@@ -96,77 +96,101 @@ def forward_training(self, images, features, predictions, gt_instances):
...
+        pairwise_dist = pairwise_dist.max(dim=2).values
+        match_quality_matrix &= (pairwise_dist > lower_bound[None, :]) & (
+            pairwise_dist < upper_bound[None, :]
+        )
+        # Match the GT box with minimum area, if there are multiple GT matches.
+        gt_areas = gt_boxes.area()  # (M, )
 
-            matched_indices.append(matched_idx)
-        return matched_indices
+        # Get matches and their labels using match quality matrix.
```
ppwwyyxx, line 150: this line of comment seems unrelated

kdexd: Oops, an artifact of prior commit. Thanks for spotting!

kdexd:
After the recent refactor (removing anchor matcher and re-adding `FCOS._match_anchors`, I am getting 39.52 AP using built-in config (COCO detection with R-50-FPN, 1x schedule).

image

kdexd:
I rechecked everything to verify correctness and I found that I added an incorrect type hint in `compute_ctrness_targets` — `anchors` is `List[Boxes]` and not `List[torch.Tensor]`.

ppwwyyxx:
@zhanghang1989 could you merge this bug fix? Thanks!

PR #3860, opened 2022-01-06
Add cache option to COCOEvaluator

alexander-soare:
Addresses https://github.com/facebookresearch/detectron2/issues/3859

### DIFF
--- a/detectron2/evaluation/coco_evaluation.py
+++ b/detectron2/evaluation/coco_evaluation.py
@@ -50,6 +50,7 @@ class COCOEvaluator(DatasetEvaluator):
         max_dets_per_image=None,
         use_fast_impl=True,
         kpt_oks_sigmas=(),
+        allow_cached_coco=True
     ):
         """
         Args:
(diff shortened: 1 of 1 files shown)


alexander-soare:
@zhanghang1989 Thanks for handling this PR. Just had a look at the test failure and noticing that the errors are not do do with any of the files changed in my commit. Any ideas on how to move forward?

ppwwyyxx:
`allow_cached` or `allow_caching` would be a better name for the argument since the class already has `COCO` in its name

alexander-soare:
@ppwwyyxx the reason I appended `_coco` is that it's not clear from the context what is being cached. Whereas in `convert_to_coco_json` I think it's fine to have it as just `allow_cached` because it's obvious what it means. Happy to change it though. Since this is closed, should I drop a new PR?

ppwwyyxx:
Thanks for the explanation. It has convinced me that the current PR is good as-is.

PR #3874, opened 2022-01-10
add scipy dependency

wenliangzhao2018:
Thanks for your contribution!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -184,7 +184,7 @@ setup(
         "omegaconf>=2.1",
         "hydra-core>=1.1",
         "black==21.4b2",
-	"scipy>=1.7.1",
+	"scipy>=1.5.1",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -184,6 +184,7 @@ def get_model_zoo_configs() -> List[str]:
         "omegaconf>=2.1",
         "hydra-core>=1.1",
         "black==21.4b2",
+	"scipy>=1.5.1",
```
ppwwyyxx, line 187: scipy is not a pure-python package. As Line160 mentioned, it's better to avoid declaring hard dependency on such packages. You can add it to Line194 as an optional dependency, since it's only required by a single feature for now.

The open source CI is already set to install all optional dependencies, so the tests should pass after that.

PR #3909, opened 2022-01-21
Make detectron2 usable without building C extensions

ppwwyyxx:
The C code in detectron2 are:
* deformable conv & rotated boxes ops
* fast COCO evaluation

So all the core features should be able to work without building any C code. 

This PR make D2 usable without compiling C extensions by making some import statements optional. This will make it easier to use D2 where the compilation toolchain & build system is different.

(There is an open internal task about this as well)

### DIFF
--- a/detectron2/evaluation/fast_eval_api.py
+++ b/detectron2/evaluation/fast_eval_api.py
@@ -5,7 +5,7 @@ import numpy as np
 import time
 from pycocotools.cocoeval import COCOeval
 
-from detectron2.utils.develop import create_dummy_class
+from detectron2 import _C
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


detectron2/evaluation/fast_eval_api.py
```diff
@@ -119,3 +119,10 @@ def accumulate(self):
         self.eval["scores"] = np.array(self.eval["scores"]).reshape(self.eval["counts"])
         toc = time.time()
         logger.info("COCOeval_opt.accumulate() finished in {:0.2f} seconds.".format(toc - tic))
+
+
+try:
+    from detectron2 import _C
+except ImportError:
+    _msg = "detectron2 is not compiled successfully, please build following the instructions!"
+    COCOeval_opt = create_dummy_class("COCOeval_opt", "detectron2._C", _msg)
```
zhanghang1989, line 128: Can we fall back to non-accelerated version?

ppwwyyxx: Added fallback with official COCO evaluation.

Issue #3515, opened 2021-09-23
cv2.findContours() produces overlapping prediction masks


jonnyevans3210:
## Instructions To Reproduce the Issue:
Train any model on a very small number of images before testing. When using Visualizer from detectron2.utils.visualizer, when visualising test images you will see (on occasion) overlapping segmentation masks. Such as:

![image](https://user-images.githubusercontent.com/87654835/134494069-7d76347d-a0cc-4335-87ed-6e74754161dc.png)

Note that the small bits are not holes, they are masks overlapping themselves. 

## Expected behavior:
All images visualised to not have any overlaps. Like so:
![image](https://user-images.githubusercontent.com/87654835/134495043-9930d1e6-91a3-4260-9c26-4f54c46ed87e.png)

## Cause and possible solution
The issue originates from the mask_to_polygon function [here](https://github.com/facebookresearch/detectron2/blob/main/detectron2/utils/visualizer.py#L116) function. Specifically, the cv2.findContours() can erroneously output contours that are subsets of one another. In order to achieve the expected behaviour above, I overwrote the mask_to_polygon function to remove any contours that are subsets of others. The code is below. It is fast (doesn't add much time to the inference at all <0.1seconds per image), but clearly it is not very efficient currently and could be improved.
 
```    
def mask_to_polygons(self, mask):
      # cv2.RETR_CCOMP flag retrieves all the contours and arranges them to a 2-level
      # hierarchy. External contours (boundary) of the object are placed in hierarchy-1.
      # Internal contours (holes) are placed in hierarchy-2.
      # cv2.CHAIN_APPROX_NONE flag gets vertices of polygons from contours.
      mask = np.ascontiguousarray(mask)  # some versions of cv2 does not support incontiguous arr
      res, hierarchy = cv2.findContours(mask.astype("uint8"), cv2.RETR_CCOMP, cv2.CHAIN_APPROX_NONE)

      if hierarchy is None:  # empty mask
          return [], False

      #TODO speed this search up
      new_polys = res.copy()
      indices_to_keep = list(range(len(res)))
      for i,r1 in enumerate(res):
          for j,r2 in enumerate(res):
              if i == j:
                  continue
              if np.in1d(r2.ravel(), r1.ravel()).all():
                  if len(r2) > len(r1):
                      if i in indices_to_keep:
                          indices_to_keep.remove(i)
                  elif len(r1) > len(r2):
                      if j in indices_to_keep:
                          indices_to_keep.remove(j)

[...]

PR #4003, opened 2022-02-28
Fix self overlapping masks

Ivan-Yovchev:
This is an attempt to fix issue https://github.com/facebookresearch/detectron2/issues/3515

# Explanation
Currently the Visualizer sometimes produces masks where a section of the mask is "doubled" (see example images in the referenced issue).

The current pipeline for drawing an item mask is the following. First  [`findContours`](https://docs.opencv.org/4.x/d3/dc0/group__imgproc__shape.html#gadf1ad6a0b82947fa1fe3c3d497f260e0) is applied to the binary prediction mask in order to extract the perimeter of each segment in the mask (**Note**: this can result in multiple contours including nested ones). The resulting polygon(s) are passed to the [`overlay_instances`](https://github.com/facebookresearch/detectron2/blob/6999554d5a02e0c9c7a22c1abd8f6eb6812fad0a/detectron2/utils/visualizer.py#L607) method where for **each** segment (i.e. extracted perimeter/polygon) the drawing method [`draw_polygon`](https://github.com/facebookresearch/detectron2/blob/6999554d5a02e0c9c7a22c1abd8f6eb6812fad0a/detectron2/utils/visualizer.py#L1114) is called. Internally the method creates a matplolib `Polygon` class with `fill=True`. This means that while the [`GenericMask`](https://github.com/facebookresearch/detectron2/blob/6999554d5a02e0c9c7a22c1abd8f6eb6812fad0a/detectron2/utils/visualizer.py#L59) class contains information about having holes in the mask said information is not used in the drawing pipeline and instead the entire polygon is filled.

image

The "doubling" of certain areas of the mask can then be explained given the perimeter extraction method together with the drawing method. Consider the example image above which represents a roughly rectangular object with a hole in the mask. The aforementioned `findContours` method would result in two perimeters (i.e. one inside the other) as there are two edges (one edge from black to white and one from white to black). The drawing method would then draw a polygon **for each** perimeter with `fill=True`. The final result then is a "doubling" of the mask where the hole was supposed to be. See the example image below

image

# Proposed solution
In order to avoid these issues, instead of drawing a polygon for each contour the proposed solution directly draws the binary mask using the already existing `draw_binary_mask` method which takes into account the information regarding holes in the mask. An example image using the proposed solution can be seen below

image

**Note**: The issues described above seem to apply only in the case of Instance Segmentation as the `draw_binary_mask` method **is called** in the case of Semantic Segmentation or KeyPoints Detection.

### DIFF
--- a/detectron2/utils/visualizer.py
+++ b/detectron2/utils/visualizer.py
@@ -691,8 +691,7 @@ class Visualizer:
                 self.draw_box(boxes[i], edge_color=color)
 
             if masks is not None:
-                for segment in masks[i].polygons:
-                    self.draw_polygon(segment.reshape(-1, 2), color, alpha=alpha)
+                self.draw_binary_mask(masks[i].mask, color, alpha=alpha)
 
             if labels is not None:
                 # first get a box


diogo-sr:
Hi. Is there any update about the status of this pull request or the fix of this bug in general?

Issue #4096, opened 2022-03-22
UnboundLocalError attempting to reference non-existent image


alexander-soare:
[Here](https://github.com/facebookresearch/detectron2/blob/d1e04565d3bec8719335b88be9e9b961bf3ec464/detectron2/modeling/test_time_augmentation.py#L193-L202), there is a reference to `image` on line 200 but this won't necessarily be defined.

PR #4097, opened 2022-03-22
Fix bug: UnboundLocalError for

alexander-soare:
Fixes https://github.com/facebookresearch/detectron2/issues/4096. Admittedly I haven't done the due diligence of checking the wider context or running tests, but feels like a safe fix to me 🤞


alexander-soare:
Hmm I don't think I caused that linting error

ppwwyyxx (approved):

Issue #3488, opened 2021-09-16
Exporting a model to ONNX; onnx.optimizer 


mdebeer:
Exporting a model to ONNX leads to an error due to `onnx.optimizer` being missing, as these packages have been removed from `onnx >= 1.9` (exists in onnx<=1.8.1) and moved to `onnxoptimizer`. 

Running with onnx==1.8.1 does not solve the issue. It throws an `IndexError: Input  is undefined`, but I have managed to solve it using the latest onnx and onnxoptimizer.

## Instructions To Reproduce the Issue:
Using detectron 0.5 and torch 1.9.0.

Example ONNX model export:
```
#wget http://images.cocodataset.org/val2017/000000439715.jpg -q -O input.jpg

import onnx
import cv2
import torch

from detectron2 import model_zoo
from detectron2.config import get_cfg
from detectron2.export import export_onnx_model
from detectron2.modeling import build_model
from detectron2.checkpoint import DetectionCheckpointer
import detectron2.data.transforms as T

im = cv2.imread("./input.jpg")

cfg = get_cfg()
# add project-specific config 
cfg.merge_from_file(model_zoo.get_config_file("COCO-PanopticSegmentation/panoptic_fpn_R_101_3x.yaml"))
# Find a model from detectron2's model zoo.
cfg.MODEL.WEIGHTS = model_zoo.get_checkpoint_url("COCO-PanopticSegmentation/panoptic_fpn_R_101_3x.yaml")
cfg.MODEL.DEVICE='cpu'

# Build model and prepare input
model = build_model(cfg)
model.eval()
checkpointer = DetectionCheckpointer(model)
checkpointer.load(cfg.MODEL.WEIGHTS)
aug = T.ResizeShortestEdge([cfg.INPUT.MIN_SIZE_TEST, cfg.INPUT.MIN_SIZE_TEST],
                           cfg.INPUT.MAX_SIZE_TEST)
height, width = im.shape[:2]
image = aug.get_transform(im).apply_image(im)
image = torch.as_tensor(image.astype("float32").transpose(2, 0, 1))
inputs = {"image": image, "height": height, "width": width}

# Export to Onnx model
onnxModel = export_onnx_model(cfg, model, [inputs])
onnx.save(onnxModel, "test.onnx")
```

For me, this fails in the onnx.optimizer package with an IndexError, when it appears it should work correctly.

By installing the latest onnx (1.10.1) and onnxoptimizer (0.2.6) packages, I solved it as follows:
[...]

PR #4120, opened 2022-04-01
Add support for ONNX-only and Caffe2 ONNX export

thiagocrepaldi:
## Summary of changes

This PR fixes both ONNX-only and Caffe2 ONNX exporters for the latest versions of this repo and PyTorch.

For ONNX-only, the main issue is that `add_export_config(cfg)` is not exposed when Caffe2 is not compiled along with PyTorch, but for ONNX-only scenarios, such dependency is not needed. Therefore, `add_export_config` is moved from `detectron2/export/api.py` to `detectron2/export/__init__.py`

A second contribution is a new `test_export_onnx.py` test file that export almost the same models as the `test_export_tracing.py` tests.

For the Caffe2-ONNX, the main issue was a dependency on ONNX optimizer pass which is deprecated in newer ONNX versions.
This PR removes such dependency because `fuse_bn_into_conv` optimization pass is already performed by `torch.onnx.export` anyway.

Fixes https://github.com/facebookresearch/detectron2/issues/3488
Fixes https://github.com/pytorch/pytorch/issues/69674 (PyTorch repo)


tests/test_export_onnx.py
```diff
@@ -0,0 +1,243 @@
...
+            if need_unregister:
+                unregister_custom_op_symbolic(fn.symbolic_name, fn.opset_version)
+            return ret
+
+        return wrapper
+
+    return decorator
+
+
+# def resolve_conj(g, self):
```
garymm, line 67: remove?

tests/test_export_onnx.py
```diff
@@ -0,0 +1,243 @@
...
+    "The test is too slow on CPUs and will be executed on CircleCI's GPU jobs.",
+)
+
+NOT_IMPLEMENTED_YET = unittest.skipIf(
+    True,
+    "Test is still in development and not stable for running.",
+)
+
+################################################################################
+# Fake symbolics to allow exporting to succeed
```
garymm, line 38: I'd convert this to draft until https://github.com/pytorch/pytorch/issues/27212 is fixed.
We should add grid_sampler export to PyTorch before asking for review of this PR so that fake symbolics are not needed, and we can actually successfully export.

garymm (changes requested):

tests/test_export_onnx.py
thiagocrepaldi: I was actually trying to run all pipelines only, but apparently it needs maintainer approval. I was thinking in having a "@skipIfNoGridSampler" decorator to keep the test. This way, it would contain the same tests as the torch scripting component.

FrancescoMandru:
Hi, any update on this merge request?

thiagocrepaldi:
> Hi, any update on this merge request?

This PR will probably need to be split in 2. One that fixes the PointRend model (and requires a new symbolic ::grid_sampler) and a second that fixes everything else. The former should take some time, but the second should be fairly quick

FrancescoMandru:
> > Hi, any update on this merge request?
> 
> This PR will probably need to be split in 2. One that fixes the PointRend model (and requires a new symbolic ::grid_sampler) and a second that fixes everything else. The former should take some time, but the second should be fairly quick

Hope this new feature will come soon! If I can help you for something don't esitate to contact me

FrancescoMandru:
Hi, any upsate on this feature?

thiagocrepaldi:
https://github.com/facebookresearch/detectron2/pull/4132 breaks ONNX export. Root cause is related to how `def to()` symbolic is implemented or how detectron creates `to(device)` on their code

thiagocrepaldi:
> I am working on this today. If you know how to add new onnx symbolics to pytorch, you could try working on adding grid_sampler

https://github.com/pytorch/pytorch/pull/76159 adds `grid_sample` to PyTorch, so we should have this working very soon

thiagocrepaldi:
We need https://github.com/pytorch/pytorch/pull/76498 to fix ONNX export after https://github.com/facebookresearch/detectron2/pull/4132/ was merged

We also need https://github.com/pytorch/pytorch/pull/76159 to add `::grid_sapler` symbolic

After both of these, this PR is ready to go

thiagocrepaldi:
@garymm somehow I cannot request another review from you through GH's UI, but I have addressed your comments

tests/test_export_onnx.py
```diff
@@ -0,0 +1,123 @@
...
+    is_onnx_installed)
+
+
+SUPPORTED_ONNX_OPSET = 14
+
+
+# TODO: this test requires manifold access, see: T88318502
+class TestONNXTracingExport(unittest.TestCase):
+    def setUp(self):
+        if not is_onnx_installed():
```
garymm, line 26: can you decorate with `skipIfNoONNX` instead?

tests/test_export_onnx.py
```diff
@@ -0,0 +1,123 @@
...
+    def testCascadeRCNN(self):
+        def inference_func(model, image):
+            inputs = [{"image": image}]
+            return model.inference(inputs, do_postprocess=False)[0]
+
+        self._test_model_zoo_from_config_path(
+            "Misc/cascade_mask_rcnn_R_50_FPN_3x.yaml", inference_func
+        )
+
+    # # bug fixed by https://github.com/pytorch/pytorch/pull/67734
```
garymm, line 59: extra #

tests/test_export_onnx.py
```diff
@@ -0,0 +1,123 @@
...
+            )
+        assert onnx.load_from_string(f.getvalue())
+
+    def _test_model_zoo_from_config_path(self, config_path, inference_func, batch=1):
+        # TODO: Using device='cuda' raises
+        #   RuntimeError: isBool() INTERNAL ASSERT FAILED at
+        #   "/github/pytorch/torch/include/ATen/core/ivalue.h":590,
+        #   please report a bug to PyTorch.
+        #   from /github/pytorch/torch/_ops.py:142 (return of OpOverloadPacket.__call__)
+        model = model_zoo.get(config_path, trained=True, device="cpu")
```
garymm, line 89: I'm confused: does `device="cpu"` mean that it will always run on CPU? If so, then why are we skipping tests when cuda is not available? Even if cuda is available, we won't use it?

tests/test_export_torchscript.py
```diff
@@ -28,6 +28,7 @@
     convert_scripted_instances,
     get_sample_coco_image,
     random_boxes,
+    SLOW_PUBLIC_CPU_TEST
```
garymm, line 31: why?

garymm (changes requested):
Thanks!

detectron2/export/api.py
```diff
@@ -3,233 +3,244 @@
...
-        """
-        if self._predictor is None:
-            self._predictor = ProtobufDetectionModel(self._predict_net, self._init_net)
-        return self._predictor(inputs)
+try:
+    from caffe2.proto import caffe2_pb2
+except ImportError:
+    pass
+else:
+    # Adds Caffe2 dependencies when PyTorch is built with BUILD_CAFFE2=1
```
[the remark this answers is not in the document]
ppwwyyxx: The function `add_export_config` does nothing and is no longer used in the public code. It's only there to not break some internal code. So there is no need to bother with it.

tests/test_export_onnx.py
```diff
@@ -0,0 +1,123 @@
...
+from detectron2.utils.testing import (
+    get_sample_coco_image,
+    SLOW_PUBLIC_CPU_TEST,
+    is_onnx_installed)
+
+
+SUPPORTED_ONNX_OPSET = 14
+
+
+# TODO: this test requires manifold access, see: T88318502
```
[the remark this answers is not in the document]
thiagocrepaldi: pre-existing comment. I have no idea they are talking about here lol

tests/test_export_torchscript.py
thiagocrepaldi: Before `SLOW_PUBLIC_CPU_TEST` was defined here, but I moved to the new place to be used by `test_export_onnx.py`. I didnt use yet as I didnt measure the export time on CI just yet

detectron2/utils/testing.py
```diff
@@ -10,12 +12,33 @@
...
 from detectron2.utils.file_io import PathManager
+from functools import wraps
+from torch.hub import _check_module_exists
 
 
 """
 Internal utilities for tests. Don't use except for writing tests.
 """
 
+SLOW_PUBLIC_CPU_TEST = unittest.skipIf(
```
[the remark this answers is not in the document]
thiagocrepaldi: pre-existing code. It was moved from `tests/test_export_torchscript.py` for possible reuse at `tests/test_export_onnx.py`. First I need to know it is working, then we skip the slow ones

detectron2/export/api.py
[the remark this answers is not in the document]
thiagocrepaldi: I did that on my first revision, but then felt guilty in removing an API from `apy.py` lol

tests/test_export_onnx.py
FrancescoMandru: Based on the Detectron2 documentation, if device is specified, it will overwrite the device specified in the documentation, so yes, I think it will load the model in cpu mode. I don't know if this was your doubt.

tests/test_export_onnx.py
thiagocrepaldi: At first I thought about that too, but preferred the current approach, which is also used in other [tests](https://github.com/facebookresearch/detectron2/blob/main/tests/layers/test_mask_ops.py#L64)

The current approach generates a summary listing all tests that are disabled and including the reason, as shown below. The user is allowed to take an action from this.

```
pytest -sv tests/test_export_onnx.py 
==================================================================================================================================== test session starts ====================================================================================================================================
platform linux -- Python 3.9.12, pytest-7.1.1, pluggy-1.0.0 -- /bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/detectron2/.hypothesis/examples')
rootdir: /detectron2
plugins: cov-3.0.0, hypothesis-6.44.0, typeguard-2.13.3, nbval-0.9.6, hydra-core-1.2.0.dev2
collected 5 items                                                                                                                                                                                                                                                                           

tests/test_export_onnx.py::TestONNXTracingExport::testCascadeRCNN SKIPPED (ONNX is not installed)
tests/test_export_onnx.py::TestONNXTracingExport::testMaskRCNNC4 SKIPPED (ONNX is not installed)
tests/test_export_onnx.py::TestONNXTracingExport::testMaskRCNNFPN SKIPPED (ONNX is not installed)
tests/test_export_onnx.py::TestONNXTracingExport::testMaskRCNNFPN_batched SKIPPED (ONNX is not installed)
tests/test_export_onnx.py::TestONNXTracingExport::testRetinaNet SKIPPED (ONNX is not installed)
```

Annotating `@skipIfNoONNX`, on the other hand, shows no warning. Users would never know about the existence of `tests/test_export_onnx.py` and the need of onnx for additional tests. See below

```
pytest -sv tests/test_export_onnx.py 
==================================================================================================================================== test session starts ====================================================================================================================================
platform linux -- Python 3.9.12, pytest-7.1.1, pluggy-1.0.0 -- /bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/detectron2/.hypothesis/examples')
rootdir: /detectron2
plugins: cov-3.0.0, hypothesis-6.44.0, typeguard-2.13.3, nbval-0.9.6, hydra-core-1.2.0.dev2
collected 0 items
```

Annotating `@skipIfNoONNX` in each test individually has the same behavior as the `setUp` one, but it is tedious to duplicate it everywhere

tests/test_export_onnx.py
thiagocrepaldi: I will try removing the CPU restriction and check whether the CI passes. If it does, I will remove the `device` argument

thiagocrepaldi:
@garymm just pushed a new revision for when you have a chance to look into it

tests/test_export_onnx.py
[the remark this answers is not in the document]
garymm: I'd err on the side of not copying it here if you have no idea what it means.

tests/test_export_onnx.py
```diff
@@ -0,0 +1,118 @@
...
+class TestONNXTracingExport(unittest.TestCase):
+    def setUp(self):
+        if not _check_module_exists("onnx"):
+            self.skipTest("ONNX is not installed")
```
garymm, line 28: can you decorate the class with `skipIf` instead?

garymm (approved):

detectron2/export/__init__.py
```diff
@@ -12,4 +12,9 @@
 from .flatten import TracingAdapter
 from .torchscript import scripting_with_instances, dump_torchscript_IR
 
+
+def add_export_config(cfg):
+    return cfg
```
ppwwyyxx, line 17: The function does nothing and is not used anywhere in detectron2, so there is really no need to call or move it.

It's kept there likely because some internal code calls it - but I'm not sure.

detectron2/utils/testing.py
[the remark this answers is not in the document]
ppwwyyxx: The name is indeed confusing. It would be good to rename it to `skip_on_cpu_ci`. I don't think it's referenced anywhere else.

tests/test_export_onnx.py
```diff
@@ -0,0 +1,114 @@
...
+        import onnx  # noqa: F401
+        f = io.BytesIO()
+        with torch.no_grad():
+            torch.onnx.export(
+                model,
+                inputs,
+                f,
+                training=torch.onnx.TrainingMode.EVAL,
+                operator_export_type=torch.onnx.OperatorExportTypes.ONNX,
+                opset_version=SUPPORTED_ONNX_OPSET,
```
ppwwyyxx, line 76: Could you also make necessary changes to https://github.com/facebookresearch/detectron2/blob/0ad20f13e23f2f4454be6196c1cd38e2171b294c/tools/deploy/export_model.py#L127 such that the example will work? It's likely not in a working state

ppwwyyxx (commented):
I no longer have the permission to merge it as I left facebook. 
But left some comments to the best of my knowledge. Looks good overall as long as tests passed.

tests/test_export_onnx.py
```diff
@@ -0,0 +1,114 @@
...
+from detectron2.utils.env import TORCH_VERSION
+from detectron2.utils.testing import (
+    get_sample_coco_image,
+    SLOW_PUBLIC_CPU_TEST)
+
+
+SUPPORTED_ONNX_OPSET = 14
+
+
+@unittest.skipIf(not _check_module_exists("onnx"), "ONNX not installed.")
```
ppwwyyxx, line 24: Actually, could you include https://github.com/facebookresearch/detectron2/pull/4190 into this PR so the newly added tests can actually run?

tests/test_export_onnx.py
```diff
@@ -0,0 +1,184 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+import io
+import onnx
+import os
+import unittest
```
FrancescoMandru, line 5: Change order here for CI tests

detectron2/utils/testing.py
```diff
@@ -1,7 +1,9 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import io
 import numpy as np
+import os
 import torch
```
FrancescoMandru, line 5: Change order here for CI tests

detectron2/export/__init__.py
thiagocrepaldi: I have seen some online examples with this, so probably better to keep it and add a deprecation warning. What do you think?

tests/test_export_onnx.py
```diff
@@ -0,0 +1,114 @@
...
+from detectron2.export import add_export_config
+from detectron2.export.flatten import TracingAdapter
+from detectron2.modeling import build_model
+from detectron2.utils.env import TORCH_VERSION
+from detectron2.utils.testing import (
+    get_sample_coco_image,
+    SLOW_PUBLIC_CPU_TEST)
+
+
+SUPPORTED_ONNX_OPSET = 14
```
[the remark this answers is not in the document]
thiagocrepaldi: Is pytorch upgrade an option? The newer, the better ONNX support it has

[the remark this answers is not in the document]
thiagocrepaldi: But I will check what is the newer opset we can use with pytorch 1.10 in the mean time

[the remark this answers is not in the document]
FrancescoMandru: I tried with a simple model in colab with `torch==1.10` and it works.

```
import io
import numpy as np

from torch import nn
import torch.utils.model_zoo as model_zoo
import torch.onnx

import torch.nn as nn
import torch.nn.init as init


class SimpleNet(nn.Module):
    def __init__(self, inplace=False):
        super(SimpleNet, self).__init__()

        self.relu = nn.ReLU(inplace=inplace)
        self.conv1 = nn.Conv2d(1, 64, (5, 5), (1, 1), (2, 2))


    def forward(self, x):
        x = self.relu(self.conv1(x))
        return x

# Create the super-resolution model by using the above model definition.
torch_model = SimpleNet()

# set the model to inference mode
torch_model.eval()

# Input to the model
x = torch.randn(1, 1, 224, 224, requires_grad=True)
torch_out = torch_model(x)

# Export the model
torch.onnx.export(torch_model,               # model being run
                  x,                         # model input (or a tuple for multiple inputs)
                  "super_resolution.onnx",   # where to save the model (can be a file or file-like object)
                  export_params=True,        # store the trained parameter weights inside the model file
                  opset_version=14,          # the ONNX version to export the model to
                  do_constant_folding=True,  # whether to execute constant folding for optimization
                  input_names = ['input'],   # the model's input names
                  output_names = ['output'], # the model's output names
                  dynamic_axes={'input' : {0 : 'batch_size'},    # variable length axes
                                'output' : {0 : 'batch_size'}})
```

thiagocrepaldi:
Fixes https://github.com/facebookresearch/detectron2/discussions/4188

tests/test_export_onnx.py
thiagocrepaldi: It does work, as long as `onnx` and `onnxoptimizer` are installed

tests/test_export_onnx.py
```diff
@@ -0,0 +1,109 @@
...
+        adapter_model = TracingAdapter(model, inputs, inference_func)
+        adapter_model.eval()
+        return self._test_model(adapter_model, adapter_model.flattened_inputs)
+
+    def _test_model_from_config_path(self, config_path, inference_func, batch=1):
+        from projects.PointRend import point_rend
+
+        cfg = get_cfg()
+        cfg.DATALOADER.NUM_WORKERS = 0
+        cfg = add_export_config(cfg)
```
ppwwyyxx, line 88: remove this no-op line

thiagocrepaldi:
@ppwwyyxx could you review one more time? Because the CI uses PyTorch 1.10 and older, I had to monkey patch 2 ONNX symbolics to allow the export to suceed.

Another option would be remove all patching and add a `skipIf` pytorch < 1.12 (next version to be released)

detectron2/utils/testing.py
```diff
@@ -1,7 +1,10 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import io
 import numpy as np
+import os
+import unittest
```
FrancescoMandru, line 5: Change order of imports

FrancescoMandru (approved):

thiagocrepaldi:
@FrancescoMandru Thanks, I think you have to approve one workflow to run before we can merge. Is that right? On my end I see this

```
1 workflow awaiting approval
First-time contributors need a maintainer to approve running workflows. [Learn more.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)
7 successful checks
```

.github/workflows/workflow.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Lint
         run: |
           echo "Running isort"
-          isort -c -sp .
```
ppwwyyxx, line 27: There is no need to read the verbose messages and figure out what's wrong manually. Just run `./dev/linter.sh` with correct version of isort/black, and all isort/black issues should be fixed.

thiagocrepaldi: @ppwwyyxx I am using `isort==4.3.21`, but the way `dev/linter.sh` fixes the code is not accepted by `isort -c -vb -sp .` at `workflow.yaml`

It moves `unittest` before torch, for `instance`

ppwwyyxx: It's correct to move `unittest` before `torch` because `unittest` is stdlib.

detectron2/utils/testing.py
```diff
@@ -1,7 +1,10 @@
...
 import torch
+from packaging import version
```
ppwwyyxx, line 7: A guess that this import might be causing some inconsistencies. 
`packaging` is a third-party library so it should be declared in `setup.cfg` as a thirdparty, so isort knows it is a thirdparty.
It should also be declared in setup.py.

isort uses what's importable in the environment to determine whether something is third-party. - so this behavior can be inconsistent across environemtn

thiagocrepaldi: Thank you, I will try that

thiagocrepaldi: @ppwwyyxx Could you approve one more run? hoppefully your suggestion does the trick :)

ppwwyyxx (approved):

FrancescoMandru:
> approved and passes all CI

Thank you for your time and effort @thiagocrepaldi. Waiting for merge in main and then I will play with this repo in order to export MaskRCNN in Onnx format

FrancescoMandru:
Why import is still blocked?

thiagocrepaldi:
> Why import is still blocked?

The repo keep changing, which requires the branch to be updated.
Because this is my first PR, an internal pipeline must be authorized after every change

I am stuck in this infinite loop for weeks :)

FrancescoMandru:
@thiagocrepaldi, little off-topic, does the future onnx export of MaskRCNN output the data already processed like for example in the Coolab notebook or the output is raw? Because there are some export modes in Detectron2 (https://detectron2.readthedocs.io/en/latest/tutorials/deployment.html) where the output needs to be processed.

thiagocrepaldi:
> @thiagocrepaldi, little off-topic, does the future onnx export of MaskRCNN output the data already processed like for example in the Coolab notebook or the output is raw? Because there are some export modes in Detectron2 (https://detectron2.readthedocs.io/en/latest/tutorials/deployment.html) where the output needs to be processed.

Will look into the notebook today, but ideally we would support post-processing as well, unless their transformations cannot be properly represented in ONNX for some reason. Even still, we can try to improve ONNX to enable new scenarios

thiagocrepaldi:
@zhanghang1989 could you help me with the Internal Linter and test failure for this PR? I cannot repro this locally, but should be something simple

FrancescoMandru:
Hi @thiagocrepaldi!

> I am stuck in this infinite loop for weeks :)

Are you still stucked in this infininite loop?

FrancescoMandru:
> Yes, there is an internal linter warning (which causes pipeline failure) that I cannot repro and couldnt get any feedback from Facebook Research to get it fixed

@zhanghang1989 @ppwwyyxx 

Could you help us for this internal bug please?

thiagocrepaldi:
Merging https://github.com/facebookresearch/detectron2/pull/4153 into this PR. It was reviewed and approved too
That should simplify the merge pipeline

ppwwyyxx:
Now I found this PR has many unreviewed changes and frankly I can see some of them are wrong. Could you move back to a state where it only contains reviewed changes? IIRC I only reviewed two PRs (for ONNX export and caffe2 export). FYI, @FrancescoMandru is not a developer of the project and just having their approval is not enough.

I fully understand that the delay in merging is not a good experience. But adding more changes to make a super huge PR is not helping. Let's just wait for maintainers when they have a chance.

thiagocrepaldi:
> Now I found this PR has many unreviewed changes and frankly I can see some of them are wrong. Could you move back to a state where it only contains reviewed changes? IIRC I only reviewed two PRs (for ONNX export and caffe2 export). FYI, @FrancescoMandru is not a developer of the project and just having their approval is not enough.
> 
> I fully understand that the delay in merging is not a good experience. But adding more changes to make a super huge PR is not helping. Let's just wait for maintainers when they have a chance.

Cheers @ppwwyyxx! Sorry for the trouble, my goal was to simplify the merge process by combining both approved PRs #4120/#4153 which were waiting for the merge. In the meantime, I found a couple minor bugs that I decided to append so that we can provide an end-to-end ONNX export implementation that works out of the box.

Instead of reverting it, I've just rebased the PR, splitting it in 5 commits with detailed explanation in each of them.
In summary, the first 2 are the same changes you have approved before (for ONNX and Caffe2). The next three fixes the ONNX export, remove unnecessary output from ONNX graph (and ONNX graph only) and force tracing of heatmaps_to_keypoints(), which is OK because ONNX uses dynamic_axes argument to ensure the graph will accept dynamic inputs

I kindly ask you that you give it a shot in reviewing them, and if you still think that I should split them again, I will do it.

Happy friday

FrancescoMandru:
> > @FrancescoMandru is not a developer of the project and just having their approval is not enough.

Definitely true, I'm trying to help @thiagocrepaldi as far as I can!

detectron2/modeling/meta_arch/retinanet.py
```diff
@@ -353,11 +353,16 @@ def norm(c):
...
+            try:
+                norm_name = str(type(get_norm(norm, 1)))
+                if "BN" in norm_name:
+                    logger.warning(
+                        f"Shared BatchNorm (type={norm_name}) may not work well in RetinaNetHead."
+                    )
+            except ValueError:
+                # https://github.com/pytorch/pytorch/pull/74293
+                if isinstance(norm, str):
+                    logger.warning(f"No BatchNorm was found for '{norm}' for RetinaNetHead.")
```
ppwwyyxx, line 365: Please undo this change. There is no reason to ignore the exception.

ppwwyyxx (changes requested):
https://github.com/facebookresearch/detectron2/pull/4120/commits/cb755feaed29e879fb141a318826789573c0523d#diff-362c43f324b3aaa4d55eae710f24245e361fd9be77cb2ecaded84672ef88b0d5R356-R365 looks good but there is a problematic change that I don't remember have reviewed. Added a comment.

The second commit https://github.com/facebookresearch/detectron2/pull/4120/commits/8f865282759ff21d8d25806fb89ed41ad972f51f looks OK, but like I said caffe2 is deprecated so there is a chance that it may not get accepted. That should be a decision separated from the ONNX-only export so it's better to be a separate PR.

The third commit https://github.com/facebookresearch/detectron2/pull/4120/commits/b35217223016e158046b8afe8cee93a37fcd8693 appears to be fixing something but I cannot see what problem it's trying to fix so I cannot accept it. 
If it is trying to fix a caffe2 export correctness issue please put it in the caffe2-onnx PR, and in that case it's less likely to be accepted because it's more work for a deprecated functionality.
If it's trying to fix a onnx-only export, correctness issue, please put it in this onnx-only export PR and add a test to show what problems it's trying to fix.

Similarly I don't understand what the 4th commit https://github.com/facebookresearch/detectron2/pull/4120/commits/f108953ba8bcf8a3f59b18c5a949bd118f2bb889 is trying to do so I cannot accept it.
It seems to be about a user experience improvement but not a bug -- in that case then it should be a separate PR. Also I expect to see demonstration of the problem first, before jumping to solutions - there is a chance that the extra outputs are useful and should not be removed, but I cannot tell if I don't know what the problem is.

I cannot tell without running some tests myself whether the 5th commit https://github.com/facebookresearch/detectron2/pull/4120/commits/6fabe508680dd2bd61556265aa59d01721bd8a81 is correct or not. I strongly suspect it's incorrect, because the scripted function has a forloop with dynamic length which I assumed ONNX tracing cannot handle. But I don't know about "dynamic_axes" and might be wrong here. Anyway that's a conversion better happen separately as well, because this does not affect the basic (fast/mask r-cnn) models.

In summary, all of the unreviewed changes will need more iterations. Having them together will slow down the merge.

thiagocrepaldi:
Divided into 3 different PRs:

https://github.com/facebookresearch/detectron2/pull/4291
https://github.com/facebookresearch/detectron2/pull/4295
https://github.com/facebookresearch/detectron2/pull/4315

Issue #4083, opened 2022-03-18
Interested in accelerate d2 inference with TorchBlade?


fortianyou:
## 🚀 Feature

To accelerate the detectron2 models with TensorRT via TorchBlade.

## Motivation & Examples

TorchBlade notices that d2 supports exporting models to TorchScript. And TorchBlade can optimize TorchScript automatically with only a few lines of code. Our benchmark shows accelerate detectron2 models with TensorRT via TorchBlade would improve performance up to 2.7x.

![image](https://user-images.githubusercontent.com/8392447/158964255-5a7aed09-2149-4427-ad1c-dfee50f3cd25.png)

## Note

All the scripts and demo cases would be public to (BladeDISC/pytorch_blade)[https://github.com/alibaba/BladeDISC/tree/main/pytorch_blade) soon.

zhanghang1989:
This is great. do you have plan to support [d2go models](https://github.com/facebookresearch/d2go/blob/main/MODEL_ZOO.md#coco-object-detection) as well?

ppwwyyxx:
Looks awesome! Would love to see the demo code.

fortianyou:
> This is great. do you have plan to support [d2go models](https://github.com/facebookresearch/d2go/blob/main/MODEL_ZOO.md#coco-object-detection) as well?

We currently have no clear plan for d2 on mobile yet. But I think our solution could be planted to mobile quickly, especially you already have something like ONNX-TensorRT. Please refer to our method to integrate [TensorRT into BladeDISC/pytorch_blade](https://github.com/alibaba/BladeDISC/tree/main/pytorch_blade/src/compiler/tensorrt). Our design docs of this modular would be update later. Feel free to contact us if you are interested in.

fortianyou:
> Looks awesome! Would love to see the demo code.

Thanks! And sorry for the reply delay. I have updated the tutorials and demo to [BladeDISC/pytorch_blade/benchmark](https://github.com/alibaba/BladeDISC/tree/main/pytorch_blade/benchmark).

cc @ppwwyyxx @zhanghang1989

PR #4132, opened 2022-04-05
enable moving traced model between devices

wat3rBro:
Summary:
For full discussion: https://fb.workplace.com/groups/1405155842844877/posts/5744470455580039

Tracing the `.to(device)` will cause problem when moving the traced torchscript to another device (eg. from cpu to gpu, or even, from `cuda:0` to `cuda:1`). The reason is that `device` is not a `torch.Tensor`, so the tracer just hardcode the value during tracing. The solution is scripting the casting operation.

Here's the code snippet illustrating this:
```
# define the MyModel similar to GeneralizedRCNN, which casts the input to the model's device
class MyModel(nn.Module):
    def __init__(self):
        super().__init__()

        self.conv1 = nn.Conv2d(3, 20, 5)
        self.conv2 = nn.Conv2d(20, 20, 5)

    def forward(self, x):
        # cast the input to the same device as this model, this makes it possible to
        # take a cpu tensor as input when the model is on GPU.
        x = x.to(self.conv1.weight.device)

        x = F.relu(self.conv1(x))
        return F.relu(self.conv2(x))

# export the model by tracing
model = MyModel()
x = torch.zeros([1, 3, 32, 32])
ts = torch.jit.trace(model, x)
print(ts.graph)

# =====================================================
graph(%self.1 : __torch__.MyModel,
      %x : Float(1, 3, 32, 32, strides=[3072, 1024, 32, 1], requires_grad=0, device=cpu)):
  %conv2 : __torch__.torch.nn.modules.conv.___torch_mangle_0.Conv2d = prim::GetAttr[name="conv2"](%self.1)
  %conv1 : __torch__.torch.nn.modules.conv.Conv2d = prim::GetAttr[name="conv1"](%self.1)
  %14 : int = prim::Constant[value=6]() # <ipython-input-2-5abde0efc36f>:11:0
  %15 : int = prim::Constant[value=0]() # <ipython-input-2-5abde0efc36f>:11:0
  %16 : Device = prim::Constant[value="cpu"]() # <ipython-input-2-5abde0efc36f>:11:0
  %17 : NoneType = prim::Constant()
  %18 : bool = prim::Constant[value=0]() # <ipython-input-2-5abde0efc36f>:11:0
  %19 : bool = prim::Constant[value=0]() # <ipython-input-2-5abde0efc36f>:11:0
  %20 : NoneType = prim::Constant()
  %input.1 : Float(1, 3, 32, 32, strides=[3072, 1024, 32, 1], requires_grad=0, device=cpu) = aten::to(%x, %14, %15, %16, %17, %18, %19, %20) # <ipython-input-2-5abde0efc36f>:11:0
  %72 : Tensor = prim::CallMethod[name="forward"](%conv1, %input.1)
  %input.5 : Float(1, 20, 28, 28, strides=[15680, 784, 28, 1], requires_grad=1, device=cpu) = aten::relu(%72) # /mnt/xarfuse/uid-20293/a90d1698-seed-nspid4026533681_cgpid21128615-ns-4026533618/torch/nn/functional.py:1406:0
  %73 : Tensor = prim::CallMethod[name="forward"](%conv2, %input.5)
  %61 : Float(1, 20, 24, 24, strides=[11520, 576, 24, 1], requires_grad=1, device=cpu) = aten::relu(%73) # /mnt/xarfuse/uid-20293/a90d1698-seed-nspid4026533681_cgpid21128615-ns-4026533618/torch/nn/functional.py:1406:0
  return (%61)
# =====================================================

# PyTorch cuda works
model = copy.deepcopy(model)
model.to("cuda")
y = model(x)
# torchscript cpu works
y = ts(x)
# torchscript cuda doesn't work
ts = ts.to("cuda")
y = ts(x)

# =====================================================
RuntimeError: Input type (torch.FloatTensor) and weight type (torch.cuda.FloatTensor) should be the same or input should be a MKLDNN tensor and weight is a dense tensor
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-4-2aece3ad6c9a> in <module>
      7 # torchscript cuda doesn't work
      8 ts = ts.to("cuda")
----> 9 y = ts(x)
/mnt/xarfuse/uid-20293/a90d1698-seed-nspid4026533681_cgpid21128615-ns-4026533618/torch/nn/modules/module.py in _call_impl(self, *input, **kwargs)
   1108         if not (self._backward_hooks or self._forward_hooks or self._forward_pre_hooks or _global_backward_hooks
   1109                 or _global_forward_hooks or _global_forward_pre_hooks):
-> 1110             return forward_call(*input, **kwargs)
   1111         # Do not call functions when jit is used
   1112         full_backward_hooks, non_full_backward_hooks = [], []
RuntimeError: The following operation failed in the TorchScript interpreter.
# =====================================================

# One solution is scripting the casting instead of tracing it, the folloing code demonstrate how to do it. We need to use mixed scripting/tracing
torch.jit.script_if_tracing
def cast_device_like(src: torch.Tensor, dst: torch.Tensor) -> torch.Tensor:
    return src.to(dst.device)

class MyModel2(nn.Module):
    def __init__(self):
        super().__init__()

        self.conv1 = nn.Conv2d(3, 20, 5)
        self.conv2 = nn.Conv2d(20, 20, 5)

    def forward(self, x):
        # cast the input to the same device as this model, this makes it possible to
        # take a cpu tensor as input when the model is on GPU.
        x = cast_device_like(x, self.conv1.weight)

        x = F.relu(self.conv1(x))
        return F.relu(self.conv2(x))

# export the model by tracing
model = MyModel2()
x = torch.zeros([1, 3, 32, 32])
ts = torch.jit.trace(model, x)
print(ts.graph)

# =====================================================
graph(%self.1 : __torch__.MyModel2,
      %x : Float(1, 3, 32, 32, strides=[3072, 1024, 32, 1], requires_grad=0, device=cpu)):
  %conv2 : __torch__.torch.nn.modules.conv.___torch_mangle_5.Conv2d = prim::GetAttr[name="conv2"](%self.1)
  %conv1 : __torch__.torch.nn.modules.conv.___torch_mangle_4.Conv2d = prim::GetAttr[name="conv1"](%self.1)
  %conv1.1 : __torch__.torch.nn.modules.conv.___torch_mangle_4.Conv2d = prim::GetAttr[name="conv1"](%self.1)
  %weight.5 : Tensor = prim::GetAttr[name="weight"](%conv1.1)
  %14 : Function = prim::Constant[name="cast_device_like"]()
  %input.1 : Tensor = prim::CallFunction(%14, %x, %weight.5)
  %68 : Tensor = prim::CallMethod[name="forward"](%conv1, %input.1)
  %input.5 : Float(1, 20, 28, 28, strides=[15680, 784, 28, 1], requires_grad=1, device=cpu) = aten::relu(%68) # /mnt/xarfuse/uid-20293/a90d1698-seed-nspid4026533681_cgpid21128615-ns-4026533618/torch/nn/functional.py:1406:0
  %69 : Tensor = prim::CallMethod[name="forward"](%conv2, %input.5)
  %55 : Float(1, 20, 24, 24, strides=[11520, 576, 24, 1], requires_grad=1, device=cpu) = aten::relu(%69) # /mnt/xarfuse/uid-20293/a90d1698-seed-nspid4026533681_cgpid21128615-ns-4026533618/torch/nn/functional.py:1406:0
  return (%55)
# =====================================================

# PyTorch cuda works
model = copy.deepcopy(model)
model.to("cuda")
y = model(x)
# torchscript cpu works
y = ts(x)
# Note that now torchscript cuda works
ts = ts.to("cuda")
y = ts(x)
print(y.device)

# =====================================================
cuda:0
# =====================================================
```

For D2 (https://github.com/facebookresearch/detectron2/commit/11528ce083dc9ff83ee3a8f9086a1ef54d2a402f), this diff creates a `move_tensor_device_same_as_another(A, B)` function to replace `A.to(B.device)`. This diff updates the `rcnn.py` and all its utils.

For D2 (https://github.com/facebookresearch/detectron2/commit/11528ce083dc9ff83ee3a8f9086a1ef54d2a402f)Go, since the exported model will become device-agnostic, we can remove the "_gpu" from predictor-type.

Differential Revision: D35367772

### DIFF
--- a/detectron2/modeling/anchor_generator.py
+++ b/detectron2/modeling/anchor_generator.py
@@ -6,7 +6,7 @@ import torch
 from torch import nn
 
 from detectron2.config import configurable
-from detectron2.layers import ShapeSpec
+from detectron2.layers import ShapeSpec, move_device_like
 from detectron2.structures import Boxes, RotatedBoxes
 from detectron2.utils.registry import Registry
 
@@ -36,13 +36,17 @@ class BufferList(nn.Module):
         return iter(self._buffers.values())
 
 
-def _create_grid_offsets(size: List[int], stride: int, offset: float, device: torch.device):
+def _create_grid_offsets(
+    size: List[int], stride: int, offset: float, target_device_tensor: torch.Tensor
+):
     grid_height, grid_width = size
-    shifts_x = torch.arange(
-        offset * stride, grid_width * stride, step=stride, dtype=torch.float32, device=device
+    shifts_x = move_device_like(
+        torch.arange(offset * stride, grid_width * stride, step=stride, dtype=torch.float32),
+        target_device_tensor,
     )
-    shifts_y = torch.arange(
-        offset * stride, grid_height * stride, step=stride, dtype=torch.float32, device=device
+    shifts_y = move_device_like(
+        torch.arange(offset * stride, grid_height * stride, step=stride, dtype=torch.float32),
+        target_device_tensor,
     )
 
     shift_y, shift_x = torch.meshgrid(shifts_y, shifts_x)
@@ -167,7 +171,7 @@ class DefaultAnchorGenerator(nn.Module):
         # buffers() not supported by torchscript. use named_buffers() instead
         buffers: List[torch.Tensor] = [x[1] for x in self.cell_anchors.named_buffers()]
         for size, stride, base_anchors in zip(grid_sizes, self.strides, buffers):
-            shift_x, shift_y = _create_grid_offsets(size, stride, self.offset, base_anchors.device)
+            shift_x, shift_y = _create_grid_offsets(size, stride, self.offset, base_anchors)
             shifts = torch.stack((shift_x, shift_y, shift_x, shift_y), dim=1)
 
             anchors.append((shifts.view(-1, 1, 4) + base_anchors.view(1, -1, 4)).reshape(-1, 4))
@@ -314,7 +318,7 @@ class RotatedAnchorGenerator(nn.Module):
     def _grid_anchors(self, grid_sizes):
         anchors = []
         for size, stride, base_anchors in zip(grid_sizes, self.strides, self.cell_anchors):
-            shift_x, shift_y = _create_grid_offsets(size, stride, self.offset, base_anchors.device)
+            shift_x, shift_y = _create_grid_offsets(size, stride, self.offset, base_anchors)
             zeros = torch.zeros_like(shift_x)
             shifts = torch.stack((shift_x, shift_y, zeros, zeros, zeros), dim=1)
 
--- a/detectron2/modeling/poolers.py
+++ b/detectron2/modeling/poolers.py
@@ -1,6 +1,6 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import math
-from typing import List
+from typing import List, Optional
 import torch
 from torch import nn
 from torchvision.ops import RoIPool
@@ -58,6 +58,16 @@ def assign_boxes_to_levels(
     return level_assignments.to(torch.int64) - min_level
 
 
+# script the module to avoid hardcoded device type
+@torch.jit.script_if_tracing
+def _convert_boxes_to_pooler_format(boxes: torch.Tensor, sizes: torch.Tensor) -> torch.Tensor:
+    sizes = sizes.to(device=boxes.device)
+    indices = torch.repeat_interleave(
+        torch.arange(len(sizes), dtype=boxes.dtype, device=boxes.device), sizes
+    )
+    return cat([indices[:, None], boxes], dim=1)
+
+
 def convert_boxes_to_pooler_format(box_lists: List[Boxes]):
     """
     Convert all boxes in `box_lists` to the low-level format used by ROI pooling ops
@@ -83,11 +93,21 @@ def convert_boxes_to_pooler_format(box_lists: List[Boxes]):
     """
     boxes = torch.cat([x.tensor for x in box_lists], dim=0)
     # __len__ returns Tensor in tracing.
-    sizes = shapes_to_tensor([x.__len__() for x in box_lists], device=boxes.device)
-    indices = torch.repeat_interleave(
-        torch.arange(len(box_lists), dtype=boxes.dtype, device=boxes.device), sizes
-    )
-    return cat([indices[:, None], boxes], dim=1)
+    sizes = shapes_to_tensor([x.__len__() for x in box_lists])
+    return _convert_boxes_to_pooler_format(boxes, sizes)
+
+
+@torch.jit.script_if_tracing
+def _create_zeros(
+    batch_target: Optional[torch.Tensor],
+    channels: int,
+    height: int,
+    width: int,
+    like_tensor: torch.Tensor,
+) -> torch.Tensor:
+    batches = batch_target.shape[0] if batch_target is not None else 0
+    sizes = (batches, channels, height, width)
+    return torch.zeros(sizes, dtype=like_tensor.dtype, device=like_tensor.device)
 
 
 class ROIPooler(nn.Module):
@@ -214,9 +234,7 @@ class ROIPooler(nn.Module):
             x[0].size(0), len(box_lists)
         )
         if len(box_lists) == 0:
-            return torch.zeros(
-                (0, x[0].shape[1]) + self.output_size, device=x[0].device, dtype=x[0].dtype
-            )
+            return _create_zeros(None, x[0].shape[1], *self.output_size, x[0])
 
         pooler_fmt_boxes = convert_boxes_to_pooler_format(box_lists)
 
@@ -227,14 +245,10 @@ class ROIPooler(nn.Module):
             box_lists, self.min_level, self.max_level, self.canonical_box_size, self.canonical_level
         )
 
-        num_boxes = pooler_fmt_boxes.size(0)
         num_channels = x[0].shape[1]
         output_size = self.output_size[0]
 
-        dtype, device = x[0].dtype, x[0].device
-        output = torch.zeros(
-            (num_boxes, num_channels, output_size, output_size), dtype=dtype, device=device
-        )
+        output = _create_zeros(pooler_fmt_boxes, num_channels, output_size, output_size, x[0])
 
         for level, pooler in enumerate(self.level_poolers):
             inds = nonzero_tuple(level_assignments == level)[0]
--- a/detectron2/structures/boxes.py
+++ b/detectron2/structures/boxes.py
@@ -144,12 +144,14 @@ class Boxes:
         Args:
             tensor (Tensor[float]): a Nx4 matrix.  Each row is (x1, y1, x2, y2).
         """
-        device = tensor.device if isinstance(tensor, torch.Tensor) else torch.device("cpu")
-        tensor = torch.as_tensor(tensor, dtype=torch.float32, device=device)
+        if not isinstance(tensor, torch.Tensor):
+            tensor = torch.as_tensor(tensor, dtype=torch.float32, device=torch.device("cpu"))
+        else:
+            tensor = tensor.to(torch.float32)
         if tensor.numel() == 0:
             # Use reshape, so we don't end up creating a new tensor that does not depend on
             # the inputs (and consequently confuses jit)
-            tensor = tensor.reshape((-1, 4)).to(dtype=torch.float32, device=device)
+            tensor = tensor.reshape((-1, 4)).to(dtype=torch.float32)
         assert tensor.dim() == 2 and tensor.size(-1) == 4, tensor.size()
 
         self.tensor = tensor
--- a/tests/test_export_torchscript.py
+++ b/tests/test_export_torchscript.py
@@ -1,10 +1,13 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
+import copy
+import glob
 import json
 import os
 import random
 import tempfile
 import unittest
+import zipfile
 import torch
 from torch import Tensor, nn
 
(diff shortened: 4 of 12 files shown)


detectron2/modeling/box_regression.py
```diff
@@ -86,7 +86,7 @@ def apply_deltas(self, deltas, boxes):
             boxes (Tensor): boxes to transform, of shape (N, 4)
         """
         deltas = deltas.float()  # ensure fp32 for decoding precision
-        boxes = boxes.to(deltas.dtype)
+        boxes = move_tensor_device_same_as_another(boxes, deltas)
```
ppwwyyxx, line 89: Is this intended? The old code does not change device.

ppwwyyxx (commented):
Apart from making modeling code more complicated, another potential downside of this approach is that device-specific python branches would stop working. Sometimes these branches are created for performance optimization, e.g. the branch in https://github.com/pytorch/vision/pull/4990 and the one in `paste_masks_in_image` are optimizations specialized for CPU/GPU. After moving the model to a different device, they could become performance regression instead.

IIRC we used to just produce torchscript models for both CPU & GPU and deploy them separately. What's causing the new need of generating a "universal" model?

(Related: https://github.com/pytorch/pytorch/issues/47010 . According to it tracing is never guaranteed to work across devices)

ppwwyyxx:
In case it's helpful, below is a replacement of `model.to(device)` for traced model. (Not sure how reliable it is). 
```
import shutil
import glob, os
from torch import nn
import torch
from torch.nn import functional as F
import zipfile

def ts_to_device(model, from_device, to_device):
    model = model.to(device=to_device)
    tmp = "/tmp/"
    basename = "model"
    jitfile = f"/tmp/{basename}.jit"
    torch.jit.save(model, jitfile)
    zipfile.ZipFile(jitfile).extractall(tmp)

    for f in glob.glob(tmp + f"{basename}/code/**.py"):
        content = open(f).read()
        new_content = content.replace(
            'torch.device("{}")'.format(from_device),
            'torch.device("{}")'.format(to_device)
        )
        if content != new_content:
            with open(f, "w") as ff:
                ff.write(new_content)

    newjit = "/tmp/newmodel"
    shutil.make_archive(newjit, "zip", root_dir=tmp,base_dir=basename)
    ret = torch.jit.load(newjit + ".zip")
    return ret


class MyModel(nn.Module):
    def __init__(self):
        super().__init__()

        self.conv1 = nn.Conv2d(3, 20, 5)
        self.conv2 = nn.Conv2d(20, 20, 5)

    def forward(self, x):
        # cast the input to the same device as this model, this makes it possible to
        # take a cpu tensor as input when the model is on GPU.
        x = x.to(self.conv1.weight.device)

        x = F.relu(self.conv1(x))
        return F.relu(self.conv2(x))

# export the model by tracing
model = MyModel()
x = torch.zeros([1, 3, 32, 32])
ts = torch.jit.trace(model, x)
print(ts.graph)
y = ts(x)

# ts = ts.to("cuda")
newts = ts_to_device(ts, "cpu", "cuda")
print(newts.graph)
y = newts(x)  # works!
```

detectron2/modeling/box_regression.py
wat3rBro: The new need comes from inference on multiple GPUs, but the model exported on `cuda:0` won't work for `cuda:1`. In theory we can export one model per GPU, but this is not vert convenient. I remember previously we have issue moving `pixel_mean` and `pixel_std` as well, but now it's not an issue anymore.

Regarding the performance optimization, I don't think this PR changes the behavior of any currently runnable model (`paste_masks_in_image` is already marked as `script_if_tracing`) so performance wise there's no impact. Maybe I missed something?

ppwwyyxx: About the line of code here:

* The original intention of code here is to make sure both deltas & boxes are in float32 instead of float16. No device casting is required from the code logic, although it's implicitly inserted.
* It seems that even `deltas.float()` at Line88 will be translated to `deltas.to(float, device="cpu")` in tracing and will have hard-coded device. So I think the workaround here is not sufficient.
* This whole class is scriptable. So perhaps a proper solution is to just make sure the method is scripted in tracing. The following should work:
```
def apply_deltas(self, deltas, boxes):
  @script_if_tracing
  def _do_apply_deltas(deltas, boxes, weights, clamp):
      ...

  return _do_apply_deltas(deltas, boxes, self.weights, self.scale_clamp
```

detectron2/modeling/poolers.py
```diff
@@ -83,10 +90,13 @@ def convert_boxes_to_pooler_format(box_lists: List[Boxes]):
     """
     boxes = torch.cat([x.tensor for x in box_lists], dim=0)
     # __len__ returns Tensor in tracing.
-    sizes = shapes_to_tensor([x.__len__() for x in box_lists], device=boxes.device)
+    sizes = shapes_to_tensor([x.__len__() for x in box_lists])
+    sizes = move_tensor_device_same_as_another(sizes, boxes)
     indices = torch.repeat_interleave(
-        torch.arange(len(box_lists), dtype=boxes.dtype, device=boxes.device), sizes
+        move_tensor_device_same_as_another(torch.arange(len(box_lists), dtype=boxes.dtype), boxes),
```
ppwwyyxx, line 96: Code like this might still have performance hit or maybe don't even run at all, because after tracing it will be translated to
```
a = torch.arange(..., device='old_device').to(device='new_device')
# instead of a = torch.arange(device='new_device')
```
If old_device=cuda0 and new_device=cuda1, this line triggers a gpu-gpu copy that's unnecessary.
If old_device=cuda and new_device=cpu, the traced model doesn't work on a pure-CPU machine at all.

My hacky `ts_to_device` script probably works for this one.

detectron2/structures/boxes.py
```diff
@@ -144,12 +144,14 @@ def __init__(self, tensor: torch.Tensor):
...
-        tensor = torch.as_tensor(tensor, dtype=torch.float32, device=device)
+        if not isinstance(tensor, torch.Tensor):
+            tensor = torch.as_tensor(tensor, dtype=torch.float32, device=torch.device("cpu"))
+        else:
+            tensor = tensor.to(torch.float32)
         if tensor.numel() == 0:
             # Use reshape, so we don't end up creating a new tensor that does not depend on
             # the inputs (and consequently confuses jit)
-            tensor = tensor.reshape((-1, 4)).to(dtype=torch.float32, device=device)
+            tensor = tensor.reshape((-1, 4)).to(dtype=torch.float32)
```
ppwwyyxx, line 154: I think this change does not help: `to(dtype=float32)` will still hardcode a device in the `to` operation if I'm not mistaken.

ppwwyyxx (commented):
Thanks for the explanation! makes sense. I pointed out one place where the approach might have performance problem (or may not work).

I'd also recommend using a shorter name such as `cast_to_device`, `move_device` etc.

(https://github.com/facebookresearch/detectron2/issues/4083 looks very promising for GPU inference, btw)

detectron2/modeling/anchor_generator.py
```diff
@@ -36,13 +36,17 @@ def __iter__(self):
...
 
-def _create_grid_offsets(size: List[int], stride: int, offset: float, device: torch.device):
+def _create_grid_offsets(
+    size: List[int], stride: int, offset: float, target_device_tensor: torch.Tensor
+):
     grid_height, grid_width = size
-    shifts_x = torch.arange(
-        offset * stride, grid_width * stride, step=stride, dtype=torch.float32, device=device
+    shifts_x = move_device_like(
+        torch.arange(offset * stride, grid_width * stride, step=stride, dtype=torch.float32),
```
ppwwyyxx, line 44: Here it also uses `arange` so it has the same issue: `arange` is created on old device and moved to new device, instead of created on the new device directly.

ppwwyyxx: To find all problems like this, maybe consider a unittest that traces the model on GPU, and then run the traced model on CPU in a separate process whose `CUDA_VISIBLE_DEVICES=""`. Such a test is supposed to fail for this line (unless pytorch has some smart optimizations).

wat3rBro: My impression is for all torch's functions that doesn't take any tensor as input (eg. `torch.arange`) without specifying the device, it'll always be on `cpu` even when running on GPU device. So this code should work for tracing on one GPU and running on another GPU or CPU. Nonetheless, I'm gonna write some GPU tests covering this.

Performance wise I'm not sure how large is the difference between creating one on CPU then moving to GPU versus creating on GPU directly. Suppose creating tensor on CPU is negligible.

detectron2/structures/boxes.py
wat3rBro: Originally I was thinking `.to(dtype=float32)` will hardcode a device, so that's why changed `boxes = boxes.to(deltas.dtype)`, but somehow it doesn't seem necessary. I reverted all the changed in `box_regression.py` and it still worked.

One thing I'm sure is that `tensor_a.to(tensor_b)` will cause hardcoded device, although it's an all-tensor interface.

tests/test_export_torchscript.py
```diff
@@ -149,13 +174,25 @@ def _test_model(self, config_path, inference_func, batch=1):
...
-            outputs = inference_func(model, *inputs)
-            traced_outputs = wrapper.outputs_schema(traced_model(*inputs))
-        if batch > 1:
-            for output, traced_output in zip(outputs, traced_outputs):
-                assert_instances_allclose(output, traced_output, size_as_tensor=True)
-        else:
-            assert_instances_allclose(outputs, traced_outputs, size_as_tensor=True)
+        # save and load back the model in order to show traceback of TorchScript
+        tmp = "/tmp"
+        basename = f"model_{uuid.uuid4().hex[:8]}"
```
ppwwyyxx, line 179: use `tempfile` so it's cleaned up.

wat3rBro:
Now all the CIs can pass, thanks to https://github.com/facebookresearch/iopath/commit/8079e60d52458ec39258625e4b6ce91e00fa57e2.

PR #4232, opened 2022-05-12
support postprocess on the output of scripted models

wat3rBro:
Summary:
The `get_fields()` in the `ScriptedInstances` will convert `Boxes` to the Tensor,
https://github.com/facebookresearch/detectron2/blob/main/detectron2/export/torchscript_patch.py#L270-L271. This will cause issue when we call `detector_postprocess` on the output of scripted model.

Differential Revision: D36336655

### DIFF
--- a/detectron2/modeling/postprocessing.py
+++ b/detectron2/modeling/postprocessing.py
@@ -2,7 +2,7 @@
 import torch
 from torch.nn import functional as F
 
-from detectron2.structures import Instances, ROIMasks
+from detectron2.structures import Boxes, Instances, ROIMasks
 
 
 # perhaps should rename to "resize_instance"
@@ -44,6 +44,11 @@ def detector_postprocess(
         output_height_tmp / results.image_size[0],
     )
     results = Instances(new_size, **results.get_fields())
+    # .get_fields() returns Tensor, need to cast it back to Boxes.
+    if results.has("pred_boxes") and isinstance(results.pred_boxes, torch.Tensor):
+        results.pred_boxes = Boxes(results.pred_boxes)
+    if results.has("proposal_boxes") and isinstance(results.proposal_boxes, torch.Tensor):
+        results.proposal_boxes = Boxes(results.proposal_boxes)
 
     if results.has("pred_boxes"):
         output_boxes = results.pred_boxes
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
This looks reasonable - but IIRC, we had wrappers (is it the mobile_cv predictor?) around torchscript models to convert its input/output format to match the format of the python model. Such wrapper must be needed for tracing anyway, so for scripting, format conversion (converting scripted `Instances` to real `Instances`) can also be done there, right?

A kind-of related issue is that, last time I checked, the scripted model CANNOT input/output custom classes (e.g. Instances) in C++ even if it's scriptable. That's why there is a wrapper [here](https://github.com/facebookresearch/detectron2/blob/42c44b622eb7b179576bdac218c84c107fff8eab/tools/deploy/export_model.py#L82-L85). This limitation also mandates a "format conversion" wrapper around any scripted model.

wat3rBro:
> This looks reasonable - but IIRC, we had wrappers (is it the mobile_cv predictor?) around torchscript models to convert its input/output format to match the format of the python model. Such wrapper must be needed for tracing anyway, so for scripting, format conversion (converting scripted `Instances` to real `Instances`) can also be done there, right?
> 
> A kind-of related issue is that, last time I checked, the scripted model CANNOT input/output custom classes (e.g. Instances) in C++ even if it's scriptable. That's why there is a wrapper [here](https://github.com/facebookresearch/detectron2/blob/42c44b622eb7b179576bdac218c84c107fff8eab/tools/deploy/export_model.py#L82-L85). This limitation also mandates a "format conversion" wrapper around any scripted model.

The wrapper is actually calling the `detector_postprocess` (eg. https://github.com/facebookresearch/d2go/blob/0c258baca427379b4792c19c34aa6843e190ae51/d2go/modeling/meta_arch/rcnn.py#L471). Currently we are using the same wrapper for tracing and scripting (the tracing one has another layer of `TracingAdapter`), and for tracing the `outputs` is `Instances` and for scripting it's `ScriptedInstances `. Yeah I think we can also do a special handling to convert `ScriptedInstances` to `Instances` outside of `detector_postprocess`.

ppwwyyxx:
Thanks, it seems to make sense to add the conversion in the wrapper.
https://github.com/facebookresearch/detectron2/blob/728ee8b0cfe045301a2ca479354b2fb969390e8e/detectron2/utils/testing.py#L71-L83 is the conversion function.

PR #4205, opened 2022-05-02
Add CUDA 11.3 + PyTorch 1.11.0 and 1.12.1

thiagocrepaldi:
Currently CI supports only CUDA 11.1, which is sufficient for PyTorch versions up to 1.10.
For PyTorch 1.11+ and nightly builds, the minimum CUDA version is 11.3.

This PR adds 3 new pipelines:
* Linux CUDA 11.3 which installs PyTorch, TorchVision, Detectron2 nightly builds.
* Linux CUDA 11.3 which installs PyTorch 1.11 and 1.12 stable release with matching TorchVision and Detectron2 builds.
* Linux CPU which installs PyTorch, TorchVision, Detectron2 nightly builds.
* Windows CPU which install PyTorch, TorchVision, Detectron2 nightly builds.

The new pipelines are important to test end-to-end ONNX export tests, in which PyTorch inference results are numerically compared with the ONNX Runtime's . A second benefit is help identifying ONNX export issues earlier in the development cycle.

ps: Nightly builds are only installed in the new pipelines (aka `windows_cpu_build_pytorch_master`, `linux_cuda113_tests_pytorch_master_python39`, `linux_cpu_tests_pytorch_master`. All existing pipelines are intact.

PR #4237, opened 2022-05-16
Fix ONNX graph missing input and with extra output constants

thiagocrepaldi:
Currently, when models are exported to ONNX although they are semantically correct, all model inputs are exported as Constants, which results in user input to be ignored during the model execution by a Runtime.

The root cause is actually the way the model was created as opposed opposed to a bug on the ONNX converter. This PR rewrites parts of the model to replace some "python" APIs by their "PyTorch" counterparts, that can handle Tensors tracing properly.


thiagocrepaldi:
@garymm I left as draft until #4120  #4205 are merged, but this is a merge candidate in case you want to review it

FrancescoMandru (approved):

detectron2/export/flatten.py
```diff
@@ -307,6 +322,18 @@ def forward(self, *args: torch.Tensor):
...
+                if len(flattened_output_tensors) > len(flattened_outputs) and\
+                        self.__is_in_torch_onnx_export:
+                    warnings.warn("PyTorch ONNX export (`torch.onnx.export`) detected!"
+                                  " To prevent extra outputs in the ONNX graph, the original"
+                                  " model output cannot be reconstructed through"
+                                  " `adapter.outputs_schema(flattened_outputs)`."
+                                  " For results evaluation, use `torch.jit.trace` instead.")
+                    flattened_outputs = flattened_outputs[:len(outputs)]
+
                 if self.outputs_schema is None:
```
FrancescoMandru, line 337: Does it means that the final output will need to be processed to be well formatted and ready to use?

tests/test_export_onnx.py
```diff
@@ -0,0 +1,149 @@
...
+    @unittest.skipIf(not min_torch_version("1.11.1"), "PyTorch 1.11.1+ is the minimum version")
+    def testMaskRCNNFPN_with_postproc(self):
+        def inference_func(model, image):
+            inputs = [{"image": image, "height": image.shape[1], "width": image.shape[2]}]
+            return model.inference(inputs, do_postprocess=True)[0]["instances"]
+
+        self._test_model_zoo_from_config_path(
+            "COCO-InstanceSegmentation/mask_rcnn_R_50_FPN_3x.yaml", inference_func, opset_version=16
+        )
+
```
thiagocrepaldi, line 92: @FrancescoMandru does that answer the question about the post processing?

FrancescoMandru: Yes @thiagocrepaldi, thank you!

PR #4258, opened 2022-05-22
Fix dictionary member access bug in new_baselines models

khoinguyen19k8:
In the new baselines config for R50-FPN Mask RCNN model. The train config is a dictionary. Hence, I changed member accesses from dotting to normal dictionary indexing.


ppwwyyxx:
This is not a bug. 

> The train config is a dictionary.

If you follow the documented instructions to use `LazyConfig.load` to load config files, then the train config is **not** just a dictionary.

Issue #4005, opened 2022-03-01
Update black to 22.1.0


BruegelN:
## 🚀 Feature

Update black, the code formatter, to version 22.1.0.
I've already did the changes over here: https://github.com/facebookresearch/detectron2/compare/main...BruegelN:update-black

## Motivation & Examples

Version 22.1.0 is the first stable release of black: https://github.com/psf/black/releases/tag/22.1.0

With this version they provide a [stability policy](https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy).

## Note

When including detectron2 into another project which is also using black it's nice to share the same (stable) version of black. The fist non-beta release of black is most likely used by some projects or they're willing to upgrade.

BUT: Updating black will also include a huge diff due to the changed codestyle guidelines: https://github.com/BruegelN/detectron2/commit/844538858bb6867a78e49d128346732d5cc3ea0d
I've already did all the required changes over here: https://github.com/facebookresearch/detectron2/compare/main...BruegelN:update-black

I'm aware this is an opinionated proposal, please let me know what you think.

jbcpollak:
this is a big deal, the pinned version of black is causing conflicts with other dependencies in our project. Please change the black dependency to use the `>=` operator.

aawor:
we face the same problem. please update.

lpasselin:
We are also having the same problem. Is black really required for detectron2?

PR #4293, opened 2022-05-31
Remove black version constraint, except for dev dependencies

lpasselin:
fix https://github.com/facebookresearch/detectron2/issues/4005

Currently, having a requirements.txt containing detectron2 and a specific version of black is conflicting.

````
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
detectron2 0.6+cu113 requires black==21.4b2, but you have black 22.3.0 which is incompatible.
````

This PR moves the strict black version to dev dependencies
It also removes the version specification for the regular dependencies.

detectron2 does require a specific version of black for development. 
detectron2 does **not** require a specific at runtime since only `black.format_str` is used. This function is available in all versions of black (even in the pre-release versions).


ppwwyyxx (approved):

wat3rBro:
I don't know why changing the requirement of `black` also changes `protobuf` version, which causes CI failing. We need to "downgrade protobuf package to 3.20.x or lower".

ppwwyyxx:
The protobuf problem is caused by the recent protobuf release https://github.com/protocolbuffers/protobuf/issues/10051. 
Setting an env var like in https://github.com/facebookresearch/detectron2/commit/1f7fa2c9a0cc85f6b938303dbbd794087e314552 can fix the issue (CI passes there)

ppwwyyxx:
I took the liberty to add the protobuf fix to @lpasselin 's branch. @wat3rBro could you merge this?

lpasselin:
@wat3rBro is there anything I can do to help you merge this?

samwaterbury:
I would love to get this merged. I consider this a bug — including a formatter in the installation requirements is a packaging mistake, and it makes `detectron2` harder to use.

@wat3rBro can this be merged?

ppwwyyxx:
Fixed in #4678

PR #4324, opened 2022-06-09
use python >=3.7

lyttonhao:
Summary:
X-link: https://github.com/fairinternal/detectron2/pull/571

fix CircleCI Test Failure as fairscale requires 3.7+

Differential Revision: D37022594

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ version_parameters: &version_parameters
       default: "https://download.pytorch.org/whl/torch_stable.html"
     python_version:  # NOTE: only affect linux
       type: string
-      default: '3.6.8'
+      default: '3.7.9'
 
   environment:
     PYTORCH_VERSION: << parameters.pytorch_version >>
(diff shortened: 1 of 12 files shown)


docker/Dockerfile
```diff
@@ -14,7 +14,7 @@ USER appuser
 WORKDIR /home/appuser
 
 ENV PATH="/home/appuser/.local/bin:${PATH}"
-RUN wget https://bootstrap.pypa.io/pip/3.6/get-pip.py && \
```
ppwwyyxx, line 17: This change is incorrect. The python in this docker container is 3.6.

lyttonhao: Thanks Yuxin for comment! So we should keep this Dockerfile unchanged? What about the `deploy.Dockerfile` file

ppwwyyxx: Best option: make `timm/fairscale` optional so that D2 still works with python 3.6; keep both dockerfile unchanged.

fairscale actually requires python 3.8 now: https://github.com/facebookresearch/fairscale/commit/ca18b50cbe0a48ad7a0ca3f1962e5f88f79d717b

lyttonhao: Let's land this PR to fix the CI first and then make fairscale optional make it more flexible

PR #4372, opened 2022-06-30
replace batch_norm calculation

FindHao:
Hi,

From pytorch profiler's reports, I found the batch_norm operation in the if-branch is inefficient. After changing it to `torch.batch_norm`, there is 1.03X speedup for the model detectron2_fasterrcnn_r_50_fpn in pytorch/benchmark. 

The previous comments says `F.batch_norm will use extra memory` but I didn't see obvious increase in pytorch profiler's reports. 

Could you please review this PR?

### DIFF
--- a/detectron2/layers/batch_norm.py
+++ b/detectron2/layers/batch_norm.py
@@ -43,7 +43,17 @@ class FrozenBatchNorm2d(nn.Module):
 
     def forward(self, x):
         if x.requires_grad:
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
The memory overhead of this change should be pretty significant. IIRC this will result in 20%-30% extra memory usage for a simple ResNet50 MaskRCNN, measured by the training log printed by detectron2.

FindHao:
@ppwwyyxx Sorry, I made some mistake when check the memory usage. Yes, there is a lot of memory increase. Thank you.

Issue #4394, opened 2022-07-11
A bug about dockerfile


peiyingxin:
## Instructions To Reproduce the 🐛 Bug:
1. Full runnable code or full changes you made:
```
docker build -t detectron2:v0 .
```
2. What exact command you run:     docker build -t detectron2:v0 .
3. __Full logs__ or other relevant observations:
```
ERROR: Package 'detectron2' requires a different Python: 3.6.9 not in '>=3.7'
```
4. please simplify the steps as much as possible so they do not require additional resources to
   run, such as a private dataset.

## Expected behavior:

it will build successful

## Environment:

Provide your environment information using the following command:
```
Failed to load OpenCL runtime
No CUDA runtime is found, using CUDA_HOME='/usr/local/cuda'
---------------------  --------------------------------------------------------------------------
sys.platform           linux
Python                 3.6.9 (default, Mar 15 2022, 13:55:28) [GCC 8.4.0]
numpy                  1.13.3
detectron2             0.6 @/home/appuser/detectron2_repo/detectron2
detectron2._C          not built correctly: No module named 'detectron2._C'
Compiler ($CXX)        c++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
DETECTRON2_ENV_MODULE  <not set>
PyTorch                1.10.0+cu111 @/home/appuser/.local/lib/python3.6/site-packages/torch
PyTorch debug build    False
GPU available          No: torch.cuda.is_available() == False
Pillow                 8.4.0
torchvision            0.11.1+cu111 @/home/appuser/.local/lib/python3.6/site-packages/torchvision
fvcore                 0.1.5
iopath                 0.1.10
cv2                    3.2.0
---------------------  --------------------------------------------------------------------------
PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) Math Kernel Library Version 2020.0.0 Product Build 20191122 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v2.2.3 (Git Hash 7336ca9f055cf1bfa13efb658fe15dc9b41f0740)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - LAPACK is enabled (usually provided by MKL)
  - NNPACK is enabled
  - CPU capability usage: AVX512
[...]

Issue #4335, opened 2022-06-15
Dockerfile Python 3.6 doesn't match the new 3.7 version requirement


bpetit2019:
You've updated the requirement of Python's version to 3.7 but in the Dockerfile, the 3.6 version is still install.

bpetit2019:
Thanks @mahesh-musashi  I'll test it.

MatthiasCami8:
@bpetit2019 what were the other modifications, as I seem to have an issue getting it to work

mahesh-musashi:
@MatthiasCami8 What's the issue?

bpetit2019:
@MatthiasCami8  on my side there were some missing keys, that I added

bpetit2019:
@mahesh-musashi did you succeed to make this image run with Nvidia driver?

leikahing:
Don't know if this will help anyone else, but I wrote a new `Dockerfile` based on Nvidia'a CentOS CUDA image (rather than their Ubuntu image or Pytorch's image) because my machine is `arm64` and the other images only supported `amd64`.


```
FROM nvidia/cuda:11.1.1-cudnn8-devel-centos8

RUN cd /etc/yum.repos.d/ && \
    sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-* && \
    sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-* && \
    dnf check-update; dnf install -y ca-certificates python38 python38-devel git sudo which gcc-c++ mesa-libGL && \
    dnf clean all

RUN alternatives --set python /usr/bin/python3 && alternatives --install /usr/bin/pip pip /usr/bin/pip3 1

# create a non-root user
ARG USER_ID=1000
RUN useradd -m --no-log-init --system  --uid ${USER_ID} appuser -g wheel
RUN echo '%wheel ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
USER appuser
WORKDIR /home/appuser

ENV PATH="/home/appuser/.local/bin:${PATH}"

# install dependencies
# See https://pytorch.org/ for other options if you use a different version of CUDA
ARG CXX="g++"
RUN pip install --user tensorboard ninja cmake opencv-python opencv-contrib-python  # cmake from apt-get is too old
[...]

PR #4405, opened 2022-07-16
upgrade python from 3.6.9 to 3.7.x in docker

farukcankaya:
Fixes https://github.com/facebookresearch/detectron2/issues/4394 https://github.com/facebookresearch/detectron2/issues/4335 by upgrading Python version from 3.6 to 3.7 inside `nvidia/cuda:11.1.1-cudnn8-devel-ubuntu18.04` docker image.


eracle:
Can you please provide some context as to why the development on this Pull Request has stopped? Are there any blockers or issues that need to be addressed before it can be resumed? Thank you.

PR #4500, opened 2022-08-23
Fix visualizer.py error.

sukjunhwang:
For certain cases, the floating value exceeds 1.0, which causes the following ValueError
```
raise ValueError("RGBA values should be within 0-1 range")
```
from matplotlib/colors.py.

The following code clips the value so it is kept within the range [0, 1].

### DIFF
--- a/detectron2/utils/visualizer.py
+++ b/detectron2/utils/visualizer.py
@@ -875,6 +875,7 @@ class Visualizer:
             font_size = self._default_font_size
 
         # since the text background is dark, we don't want the text to be dark
+        color = [min(max(c, 0.0), 1.0) for c in color]
         color = np.maximum(list(mplc.to_rgb(color)), 0.2)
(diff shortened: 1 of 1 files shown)


ppwwyyxx (approved):

ppwwyyxx:
Sorry, this PR is incorrect, please pause landing. @wat3rBro

detectron2/utils/visualizer.py
```diff
@@ -875,6 +875,7 @@ def draw_text(
             font_size = self._default_font_size
 
         # since the text background is dark, we don't want the text to be dark
+        color = [min(max(c, 0.0), 1.0) for c in color]
```
ppwwyyxx, line 878: As the documentation above says, `color` is not necessarily a list of float. Supported formats are https://matplotlib.org/stable/tutorials/colors/colors.html.

It's better to find where the out-of-bounds values are created and fix it there instead.

PR #4511, opened 2022-08-29
Bugfix: when deploying a Detectron2 model, make sure that the created tensors are located on the right device

JorritWillaert:
Earlier, when you try to deploy a Detectron2 model on for example a Triton server, the following error occured:
`RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!` (see also [here](https://github.com/triton-inference-server/server/issues/2024)). This has been resolved by sending the newly created tensor to the right device.

### DIFF
--- a/detectron2/structures/image_list.py
+++ b/detectron2/structures/image_list.py
@@ -86,7 +86,8 @@ class ImageList(object):
 
         image_sizes = [(im.shape[-2], im.shape[-1]) for im in tensors]
         image_sizes_tensor = [shapes_to_tensor(x) for x in image_sizes]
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
The `max_size` tensor is expected to always be on CPU.

Division of `max_size` and a constant should succeed regardless of which device `max_size` is on. According to reports in https://github.com/triton-inference-server/server/issues/2024 the division can fail in triton and is reported to work fine without triton. 

Unless there is more evidence to suggest otherwise, I believe it is a bug of triton.

Issue #4354, opened 2022-06-23
Panoptic Segmentation Can Not be Exported to ONNX


TankAbraham:
## Instructions To Reproduce the 🐛 Bug:

1. Full runnable code or full changes you made:
```
I use the original repo and do not change anything.
```

2. What exact command do you run:
```bash
export DETECTRON2_DATASETS=/data/datasets/

python3 export_model.py \
    --config-file ../../configs/COCO-PanopticSegmentation/panoptic_fpn_R_50_1x.yaml \
    --output /data/output/ \
    --export-method caffe2_tracing \
    --format onnx \
    MODEL.WEIGHTS /data/model/COCO-PanopticSegmentation/panoptic_fpn_R_50_1x/139514544/model_final_dbfeb4.pkl \
    MODEL.DEVICE cpu
```

I have already prepared the COCO dataset in `/data/datasets/`.

4. __Full logs__ or other relevant observations:
```
[06/23 06:22:33 detectron2]: Command line arguments: Namespace(format='onnx', export_method='caffe2_tracing', config_file='../../configs/COCO-PanopticSegmentation/panoptic_fpn_R_50_1x.yaml', sample_image=None, run_eval=False, output='/data/output/', opts=['MODEL.WEIGHTS', '/data/model/COCO-PanopticSegmentation/panoptic_fpn_R_50_1x/139514544/model_final_dbfeb4.pkl', 'MODEL.DEVICE', 'cpu'])
[W init.cpp:759] Warning: Use _jit_set_fusion_strategy, bailout depth is deprecated. Setting to (STATIC, 1) (function operator())
[06/23 06:22:35 d2.data.datasets.coco]: Loaded 5000 images in COCO format from /data/datasets/coco/annotations/instances_val2017.json
[06/23 06:22:35 d2.data.datasets.coco]: Loaded 5000 images with semantic segmentation from /data/datasets/coco/val2017
[06/23 06:22:35 d2.data.build]: Distribution of instances among all 80 categories:
|   category    | #instances   |   category   | #instances   |   category    | #instances   |
|:-------------:|:-------------|:------------:|:-------------|:-------------:|:-------------|
|    person     | 10777        |   bicycle    | 314          |      car      | 1918         |
|  motorcycle   | 367          |   airplane   | 143          |      bus      | 283          |
|     train     | 190          |    truck     | 414          |     boat      | 424          |
| traffic light | 634          | fire hydrant | 101          |   stop sign   | 75           |
[...]

PR #4520, opened 2022-08-31
Add support to PanopticFPN to ONNX exporter

thiagocrepaldi:
This PR extends `tools/deploy/export_model.py` with code to enable `PanopticFPN` to be exported to ONNX using tracing

It also improves `compatibility.md` and `deployment.md` with extra information about ONNX compatibility, while it also highlights the deprecation of Caffe2

Fixes https://github.com/facebookresearch/detectron2/issues/4354

### DIFF
--- a/docs/tutorials/deployment.md
+++ b/docs/tutorials/deployment.md
@@ -6,12 +6,8 @@ A few basic concepts about this process:
 __"Export method"__ is how a Python model is fully serialized to a deployable format.
 We support the following export methods:
 
-* `tracing`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it\
-This method produces a standardized ONNX graph that can be used by any ONNX runtime, such as \
-[ONNX Runtime](https://onnxruntime.ai/)
-* `scripting`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it\
-This method produces a standardized ONNX graph that can be used by any ONNX runtime, such as \
-[ONNX Runtime](https://onnxruntime.ai/)
+* `tracing`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it
+* `scripting`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it
 * `caffe2_tracing`: replace parts of the model by caffe2 operators, then use tracing.
 This method produces a Caffe2-specific ONNX graph that can be used only by the [Caffe2](https://caffe2.ai/).
 Caffe2 support has been ***DEPRECATED*** by PyTorch, therefore, we don't plan to support it, but contributions are welcome.
@@ -27,33 +23,33 @@ A runtime is often tied to a specific format
 We currently support the following combination and each has some limitations:
 
 ```eval_rst
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
-|       Export Method        |   tracing   |  scripting  | tracing/scripting | caffe2_tracing (DEPRECATED) |
-+============================+=============+=============+===================+=============================+
-| **Formats**                | TorchScript | TorchScript |       ONNX        | Caffe2, TorchScript, ONNX   |
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
-| **Runtime**                | PyTorch     | PyTorch     |    ONNX Runtime   | Caffe2, PyTorch             |
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
-| C++/Python inference       | ✅           | ✅           | ✅                 | ✅                           |
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
-| Dynamic resolution         | ✅           | ✅           | ✅                 | ✅                           |
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
-| Batch size requirement     | Constant    | Dynamic     | Dynamic           | Batch inference unsupported |
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
-| Extra runtime deps         | torchvision | torchvision | torchvision       | Caffe2 ops (usually already |
-|                            |             |             |                   |                             |
-|                            |             |             |                   | included in PyTorch)        |
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
-| Faster/Mask/Keypoint R-CNN | ✅           | ✅           | ✅                 | ✅                           |
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
-| RetinaNet                  | ✅           | ✅           | ✅                 | ✅                           |
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
-| PointRend R-CNN            | ✅           | ❌           | ✅                 | ❌                           |
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
-| Cascade R-CNN              | ✅           | ❌           | ✅                 | ❌                           |
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
-| PanopticFPN                | ✅           | N/A         | ✅                 | N/A                         |
-+----------------------------+-------------+-------------+-------------------+-----------------------------+
++----------------------------+-----------------------+-----------------------+-----------------------------+
+|       Export Method        |   tracing             |  scripting            | caffe2_tracing (DEPRECATED) |
++============================+=======================+=======================+=============================+
+| **Formats**                | TorchScript, ONNX     | TorchScript, ONNX     | Caffe2, TorchScript, ONNX   |
++----------------------------+-----------------------+-----------------------+-----------------------------+
+| **Runtime**                | PyTorch, ONNX Runtime | PyTorch, ONNX Runtime | Caffe2, PyTorch             |
++----------------------------+-----------------------+-----------------------+-----------------------------+
+| C++/Python inference       | ✅                     | ✅                     | ✅                           |
++----------------------------+-----------------------+-----------------------+-----------------------------+
+| Dynamic resolution         | ✅                     | ✅                     | ✅                           |
++----------------------------+-----------------------+-----------------------+-----------------------------+
+| Batch size requirement     | Constant              | Dynamic               | Batch inference unsupported |
++----------------------------+-----------------------+-----------------------+-----------------------------+
+| Extra runtime deps         | torchvision           | torchvision           | Caffe2 ops (usually already |
+|                            |                       |                       |                             |
+|                            |                       |                       | included in PyTorch)        |
++----------------------------+-----------------------+-----------------------+-----------------------------+
+| Faster/Mask/Keypoint R-CNN | ✅                     | ✅                     | ✅                           |
++----------------------------+-----------------------+-----------------------+-----------------------------+
+| RetinaNet                  | ✅                     | ✅                     | ✅                           |
++----------------------------+-----------------------+-----------------------+-----------------------------+
+| PointRend R-CNN            | ✅                     | ❌                     | ❌                           |
++----------------------------+-----------------------+-----------------------+-----------------------------+
+| Cascade R-CNN              | ✅                     | ❌                     | ❌                           |
++----------------------------+-----------------------+-----------------------+-----------------------------+
+| PanopticFPN                | ✅                     | ❌                     | ❌                           |
++----------------------------+-----------------------+-----------------------+-----------------------------+
 
 ```
 
--- a/docs/tutorials/deployment.md
+++ b/docs/tutorials/deployment.md
@@ -6,64 +6,74 @@ A few basic concepts about this process:
 __"Export method"__ is how a Python model is fully serialized to a deployable format.
 We support the following export methods:
 
-* `tracing`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it
-* `scripting`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it
+* `tracing`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it\
+This method produces a standardized ONNX graph that can be used by any ONNX runtime, such as \
+[ONNX Runtime](https://onnxruntime.ai/)
+* `scripting`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it\
+This method produces a standardized ONNX graph that can be used by any ONNX runtime, such as \
+[ONNX Runtime](https://onnxruntime.ai/)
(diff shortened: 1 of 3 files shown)


thiagocrepaldi:
@ppwwyyxx @wat3rBro @mcimpoi could you help reviewing this one when you get some time?

thiagocrepaldi:
> 

gentle ping

malfet (approved):

docs/tutorials/deployment.md
```diff
@@ -6,64 +6,74 @@ A few basic concepts about this process:
...
-| Faster/Mask/Keypoint R-CNN | ✅          | ✅          | ✅                          |
-+----------------------------+-------------+-------------+-----------------------------+
-| RetinaNet                  | ✅          | ✅          | ✅                          |
-+----------------------------+-------------+-------------+-----------------------------+
-| PointRend R-CNN            | ✅          | ❌          | ❌                          |
-+----------------------------+-------------+-------------+-----------------------------+
-| Cascade R-CNN              | ✅          | ❌          | ❌                          |
-+----------------------------+-------------+-------------+-----------------------------+
++----------------------------+-------------+-------------+-------------------+-----------------------------+
+|       Export Method        |   tracing   |  scripting  | tracing/scripting | caffe2_tracing (DEPRECATED) |
```
ppwwyyxx, line 31: Unless this is implemented in `export_model.py` I don't think it's correct to way "scripting + ONNX" is supported.

docs/tutorials/deployment.md
```diff
@@ -6,64 +6,74 @@ A few basic concepts about this process:
...
++============================+=============+=============+===================+=============================+
+| **Formats**                | TorchScript | TorchScript |       ONNX        | Caffe2, TorchScript, ONNX   |
++----------------------------+-------------+-------------+-------------------+-----------------------------+
+| **Runtime**                | PyTorch     | PyTorch     |    ONNX Runtime   | Caffe2, PyTorch             |
++----------------------------+-------------+-------------+-------------------+-----------------------------+
+| C++/Python inference       | ✅           | ✅           | ✅                 | ✅                           |
++----------------------------+-------------+-------------+-------------------+-----------------------------+
+| Dynamic resolution         | ✅           | ✅           | ✅                 | ✅                           |
++----------------------------+-------------+-------------+-------------------+-----------------------------+
+| Batch size requirement     | Constant    | Dynamic     | Dynamic           | Batch inference unsupported |
```
ppwwyyxx, line 41: I think Tracing + ONNX will have the same limitation that batch size has to be constant.

docs/tutorials/deployment.md
```diff
@@ -6,64 +6,74 @@ A few basic concepts about this process:
...
-* `scripting`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it
+* `tracing`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it\
+This method produces a standardized ONNX graph that can be used by any ONNX runtime, such as \
+[ONNX Runtime](https://onnxruntime.ai/)
+* `scripting`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it\
+This method produces a standardized ONNX graph that can be used by any ONNX runtime, such as \
+[ONNX Runtime](https://onnxruntime.ai/)
 * `caffe2_tracing`: replace parts of the model by caffe2 operators, then use tracing.
+This method produces a Caffe2-specific ONNX graph that can be used only by the [Caffe2](https://caffe2.ai/).
+Caffe2 support has been ***DEPRECATED*** by PyTorch, therefore, we don't plan to support it, but contributions are welcome.
```
ppwwyyxx, line 17: Might be a good idea to delete everything about caffe2 from this doc to simplify things.

ppwwyyxx (changes requested):

docs/tutorials/deployment.md
thiagocrepaldi: `torch.onnx.export(input_names=['input0',...],dynamic_axes={'input0': {0: 'batch', ...}})` would replace any numerical value by strings and make these axis dynamic for onnx, for both scripting and tracing

docs/tutorials/deployment.md
thiagocrepaldi: Not sure if I followed, in theory, wherever [scripted_module = torch.jit.script(model)` succeeds](https://github.com/facebookresearch/detectron2/blob/0e29b7ab8b77860b397aa02c2a164e1a8edcea8b/tests/test_export_torchscript.py), `torch.onnx.export(scripted_module)` succeeds by extension. The ONNX export does not need to do anything extra because of scripting or tracing.

docs/tutorials/deployment.md
thiagocrepaldi: I am OK in removing any reference to Caffe2 ONNX export, but my previous PR already did the bulk of the work to make it work.

Last I heard, Meta still uses Caffe2 internally, so the removal might break their thing

docs/tutorials/deployment.md
ppwwyyxx: Does that actually work for scripting detectron2 models, though? The scripted model has complex IR including custom classes/list/dict/loops.
If it works, please add `--format=onnx` support to `export_scripting` in `export_model.py`.

docs/tutorials/deployment.md
ppwwyyxx: I doubt it can automatically "make" the model dynamic - it would only works for simple models. 
The "batch" dimension in the model may not be in any axis: for an intermediate tensor, the "batch" dimension may not exist and should be constructed by a for loop. 
For example, a model that's traced with a bs=2 might have the following IR:
```
tensor = tensor(N, C)
tensor_0 = tensor[0]
tensor_1 = tensor[1]
output0 = operations(tensor_0)  # operations that cannot be batched
output1 = operations(tensor_1)
output = other_operation(output_0, output_1)
```
So it would only work for the same bs=2 in inference. And the "true" dynamic version of this model should contain a for loop over `N`. Is onnx able to automagically translate the above IR into a for loop?

docs/tutorials/deployment.md
ppwwyyxx: I'm only suggesting to remove docs since probably no one externally is using them. But either way is fine.

thiagocrepaldi:
@wat3rBro could you take a quick look at this?

wat3rBro:
@thiagocrepaldi Follow up with @ppwwyyxx 's comment, have you tested scripting+ONNX? Maybe we need to have separate columns for tracing and scripting when using onnx format.

thiagocrepaldi:
> @thiagocrepaldi Follow up with @ppwwyyxx 's comment, have you tested scripting+ONNX? Maybe we need to have separate columns for tracing and scripting when using onnx format.

I actually got a little confused by @ppwwyyxx comment on this. It seems it is asks whether torchscript really works on detectron2. There are tests at https://github.com/facebookresearch/detectron2/blob/main/tests/test_export_torchscript.py which shows it does.

On which regards to ONNX exporting, once TorchScript has succeeded, either tracing or scripting APIs, the TorchScript -> ONNX translation process is the same. The major export issue is that sometimes the user model can only run in tracing mode, leading into a TorchScript IR that inline loops/conditions. The resulting ONNX would be equally inlined, because it only translates nodes from one IR (TorchScript) to another (ONNX). In this case, the user model must be able to export in TorchScript script mode, to preserve loops and branches - this has nothing to do with ONNX. If the model is scriptable, then ONNX will be able to export it too.

`torch.onnx.export(dynamic_axes=...)` does not make the model dynamic... The model itself must be dynamic when loops and control flows are used through TorchScript script. If they are, it is just a 1:1 mapping to ONNX. AFAIK, the TorchScript unit tests prove D2 is scriptable. The `dynamic_axes` flag of the exporter merely replaces numerical shapes oming from user input with string names that are propagated throughout the ONNX graph. That makes those dimensions to assume different values other than the ones provided as input to the exporter

Does it make sense?

ppwwyyxx:
I agree, and it's indeed not related to ONNX. Tracing and scripting already has their limitations (as indicated in the existing table), and ONNX will not make these limitations disappear.

That's why I think the table in the PR is misleading. For example, tracing + ONNX will have the limitation of constant batch size (because tracing in detectron2 already has it). But the table in the PR doesn't suggest this.

One way to make it more clear is to add two columns: `tracing+format=onnx`, and `scripting+format=onnx`. They will have different "Batch size requirement" and different models that they are able to support, similar to the `tracing+format=torchscript` and `scripting+format=torchscript` columns.

docs/tutorials/deployment.md
```diff
@@ -6,64 +6,74 @@ A few basic concepts about this process:
...
 We support the following export methods:
 
-* `tracing`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it
-* `scripting`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it
+* `tracing`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it\
+This method produces a standardized ONNX graph that can be used by any ONNX runtime, such as \
+[ONNX Runtime](https://onnxruntime.ai/)
+* `scripting`: see [pytorch documentation](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) to learn about it\
+This method produces a standardized ONNX graph that can be used by any ONNX runtime, such as \
+[ONNX Runtime](https://onnxruntime.ai/)
```
ppwwyyxx, line 14: This makes it sound like tracing/scripting only works with ONNX graph format.. please remove them here and relevant info below to "Format" or "Runtime" section if needed.

thiagocrepaldi:
> I agree, and it's indeed not related to ONNX. Tracing and scripting already has their limitations (as indicated in the existing table), and ONNX will not make these limitations disappear.
> 
> That's why I think the table in the PR is misleading. For example, tracing + ONNX will have the limitation of constant batch size (because tracing in detectron2 already has it). But the table in the PR doesn't suggest this.
> 
> One way to make it more clear is to add two columns: `tracing+format=onnx`, and `scripting+format=onnx`. They will have different "Batch size requirement" and different models that they are able to support, similar to the `tracing+format=torchscript` and `scripting+format=torchscript` columns. If table gets crowded I think it's better to remove caffe2 documentation.

Oh, I got what you meant. My assumption was that the TorchScript + ONNX graph will have the same limitation as the TorchScript graph, only that in a different IR. Having two columns to make the TorchScript limitations more visible makes sense

docs/tutorials/deployment.md
ppwwyyxx: I just tried adding `torch.onnx.export(scripted_module, ...)` to `export_model.py:export_scripting()` and it does not work for me. Maybe I'm just using a bad pytorch version or I lack enough onnx knowledge.. but anyway please implement it first before claiming it in the docs.

ppwwyyxx: The error I had when exporting a maskrcnn with scripting+onnx:
```
  File "/home/USERNAME//detectron2/tools/deploy/./export_model.py", line 222, in <module>
    exported_model = export_scripting(torch_model)
  File "/home/USERNAME//detectron2/tools/deploy/./export_model.py", line 102, in export_scripting
    torch.onnx.export(ts_model, ({"image":image},), f, opset_version=STABLE_ONNX_OPSET_VERSION)
  File "/home/USERNAME/.local/lib/python3.10/site-packages/torch/onnx/utils.py", line 504, in export
    _export(
  File "/home/USERNAME/.local/lib/python3.10/site-packages/torch/onnx/utils.py", line 1506, in _export
    with exporter_context(model, training, verbose):
  File "/usr/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/home/USERNAME/.local/lib/python3.10/site-packages/torch/onnx/utils.py", line 176, in exporter_context
    with select_model_mode_for_export(
  File "/usr/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/home/USERNAME/.local/lib/python3.10/site-packages/torch/onnx/utils.py", line 116, in select_model_mode_for_export
    model.train(False)
  File "/home/USERNAME/.local/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1905, in train
    self.training = mode
  File "/home/USERNAME/.local/lib/python3.10/site-packages/torch/jit/_script.py", line 794, in __setattr__
    self._c.setattr(attr, value)
RuntimeError: Can't set constant 'training' which has value:False
```

This looks related to one of the hacks we did to make scripting works.. I doubt there is an easy fix to this. Maybe just remove scripting from this PR. Tracing has been more useful and reliable anway.

docs/tutorials/deployment.md
thiagocrepaldi: Refactored the support table to separate tracing and scripting for ONNX format.

docs/tutorials/deployment.md
thiagocrepaldi: Tracing and scripting columns for ONNX were separated. Now it should be clearer that ONNX + tracing or ONNX + scripting didn't create any new limitation or capability that wasn't there for tracing or scripting alone

thiagocrepaldi:
@ppwwyyxx see what you think about the new table with ONNX deployment description

docs/tutorials/deployment.md
ppwwyyxx: ONNX + scripting doesn't seem to work according to my comment above.. in that case ONNX+scripting should be removed from doc

thiagocrepaldi: please provide a full repro script

ppwwyyxx:
```diff
diff --git i/tools/deploy/export_model.py w/tools/deploy/export_model.py
index f507dffe..ff991195 100755
--- i/tools/deploy/export_model.py
+++ w/tools/deploy/export_model.py
@@ -72,7 +72,6 @@ def export_scripting(torch_model):
         "pred_keypoints": torch.Tensor,
         "pred_keypoint_heatmaps": torch.Tensor,
     }
-    assert args.format == "torchscript", "Scripting only supports torchscript format."
 
     class ScriptableAdapterBase(nn.Module):
         # Use this adapter to workaround https://github.com/pytorch/pytorch/issues/46944
@@ -97,9 +96,10 @@ def export_scripting(torch_model):
                 return [i.get_fields() for i in instances]
 
     ts_model = scripting_with_instances(ScriptableAdapter(), fields)
-    with PathManager.open(os.path.join(args.output, "model.ts"), "wb") as f:
-        torch.jit.save(ts_model, f)
-    dump_torchscript_IR(ts_model, args.output)
+    if args.format == "onnx":
+        with PathManager.open(os.path.join(args.output, "model.onnx"), "wb") as f:
+            image = torch.ones((100, 100, 3))
+            torch.onnx.export(ts_model, ({"image":image},), f, opset_version=STABLE_ONNX_OPSET_VERSION)
     # TODO inference in Python now missing postprocessing glue code
     return None
 
```

```
./export_model.py --config-file ../../configs/quick_schedules/mask_rcnn_R_50_FPN_inference_acc_test.yaml --output ./mrcnn_scripting_onnx --export-method scripting --format onnx    MODEL.DEVICE cpu
```

thiagocrepaldi:
@ppwwyyxx @wat3rBro gentle ping

thiagocrepaldi:
@ppwwyyxx @wat3rBro @malfet gentle ping

PR #4526, opened 2022-09-06
Add support for DW folder structure

yangbo02:
Summary: As title.

Reviewed By: wenliangzhao2018

Differential Revision: D38968928


ppwwyyxx:
It looks like you can achieve the feature by just setting a different `self.input_file_to_gt_file` and `self.sem_seg_loading_fn` that know about the new file structure.

yangbo02:
Will have a new implementation that does not require change in here.

PR #4541, opened 2022-09-14
ensure mask size is based on integers

willembressers:
Thanks for your contribution!


willembressers:
this should fix issue: https://github.com/facebookresearch/detectron2/pull/4541

ppwwyyxx:
This is not a good fix. The best action should be to figure out why in the original issue, shapes are represented as float in the first place and fix it at the root.

willembressers:
yeah i know. it was a quick and dirty fix.. i'll dive into it when i have the time

PR #4626, opened 2022-10-28
make the offload function customizable for DatasetFromList

wat3rBro:
Summary:
Previously we use `serialize: bool` to control if we want to offload the `DatasetFromList` storage to numpy. This diff generalize the "serialize" to "offload", and make the "offload function" customizable so that we can switch between implementations.

The setting of `offload function` is done by context manager in order to avoid passing this argument all the way down.

Differential Revision: D40818736

### DIFF
--- a/detectron2/data/common.py
+++ b/detectron2/data/common.py
@@ -1,16 +1,20 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
+import contextlib
 import copy
 import itertools
 import logging
 import numpy as np
 import pickle
 import random
+from typing import Callable, Union
 import torch.utils.data as data
 from torch.utils.data.sampler import Sampler
 
 from detectron2.utils.serialize import PicklableWrapper
 
 __all__ = ["MapDataset", "DatasetFromList", "AspectRatioGroupedDataset", "ToIterableDataset"]
+
+logger = logging.getLogger(__name__)
 
 
 def _shard_iterator_dataloader_worker(iterable):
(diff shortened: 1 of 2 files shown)


wat3rBro:
See https://github.com/facebookresearch/d2go/pull/405 for the benchmarking results.

detectron2/data/common.py
```diff
@@ -106,56 +110,96 @@ def __getitem__(self, idx):
...
     """
 
-    def __init__(self, lst: list, copy: bool = True, serialize: bool = True):
+    def __init__(
+        self,
+        lst: list,
+        copy: bool = True,
+        offload: bool = True,
+        _offload_func: Optional[Callable] = None,
+    ):
```
ppwwyyxx, line 177: I think there is no need to change the API here. You can do this:
```diff
--- i/detectron2/data/common.py
+++ w/detectron2/data/common.py
@@ -108,19 +108,20 @@ class MapDataset(data.Dataset):

 class DatasetFromList(data.Dataset):
     """
-    Wrap a list to a torch Dataset. It produces elements of the list as data.
+    Wrap a list-like object to a torch Dataset. It produces elements of the list as data.
     """

     def __init__(self, lst: list, copy: bool = True, serialize: bool = True):
         """
         Args:
-            lst (list): a list which contains elements to produce.
+            lst (list): a index-able object which contains elements to produce.
             copy (bool): whether to deepcopy the element when producing it,
                 so that the result can be modified in place without affecting the
                 source in the list.
             serialize (bool): whether to hold memory using serialized objects, when
                 enabled, data loader workers can use shared RAM from master
-                process instead of making a copy.
+                process instead of making a copy. Effective only if lst is a
+                python list.
         """
         self._lst = lst
         self._copy = copy
@@ -130,24 +131,11 @@ class DatasetFromList(data.Dataset):
             buffer = pickle.dumps(data, protocol=-1)
             return np.frombuffer(buffer, dtype=np.uint8)

-        if self._serialize:
-            logger = logging.getLogger(__name__)
-            logger.info(
-                "Serializing {} elements to byte tensors and concatenating them all ...".format(
-                    len(self._lst)
-                )
-            )
-            self._lst = [_serialize(x) for x in self._lst]
-            self._addr = np.asarray([len(x) for x in self._lst], dtype=np.int64)
-            self._addr = np.cumsum(self._addr)
-            self._lst = np.concatenate(self._lst)
-            logger.info("Serialized dataset takes {:.2f} MiB".format(len(self._lst) / 1024**2))
+        if self._serialize and isinstance(lst, list):
+            self._lst = NumpySerializedList(lst)

     def __len__(self):
-        if self._serialize:
-            return len(self._addr)
-        else:
-            return len(self._lst)
+        return len(self._lst)

     def __getitem__(self, idx):
         if self._serialize:
```

This way, the `lst=` argument is generalized to support either regular python list, or `NumpySerializedList/SharedList`.

ppwwyyxx: To be more clear,  `DatasetFromList(lst, offload=True, _offload_func=f)` can just become `DatasetFromList(f(lst))` under the old API, so the API is kept unchanged.

----

When `lst` is a plain list, `DatasetFromList` internally can still use `_DEFAUL_DATASET_FROM_LIST_OFFLOAD_FUNC` if a global context is needed. To not introduce globals, something like this might also be useful:
```
oldf = DatasetCatalog.get("dataset_name")
DatasetCatalog.register("dataset_name", NumpySerializedList(oldf()), overwrite=True)
```

ppwwyyxx: `if isinstance(dataset, list)` in `build_detection_{train,test}_loader` might be changed to `if not isinstance(dataset, torch.data.Dataset)` if we're generalizing the concept of list.

wat3rBro: @ppwwyyxx It feels a bit redundant to have a 'serialize' option in `DatasetFromList` and support "serialized' input `lst`, is it mainly for keeping BC? I haven't seen many people using `serialize=False` internally, but not sure about OSS. Also, are you considering forking `SharedList` in D2 (the only common dependency between D2 and mobile_cv is fvcore, but fvcore doesn't have `comm`, so we probably have to fork it)? maybe it would be related to `DatasetFromList` interface.

> DatasetCatalog.register("dataset_name", NumpySerializedList(oldf()), overwrite=True)

This way it prevents concatenating datasets, or say the concatenated datasets is not numpy-serialized.

ppwwyyxx: BC is one reason. Besides, it's nice to have a default serialization method.

Another approach could be to keep `lst` unchanged but generalize `serialize: Union[None, Callable, bool] = NumpySerializedList` where the bool is only there for BC.

>  Also, are you considering forking SharedList in D2 

No I don't have a preference on where `SharedList` lives. Having it in mobile_cv is fine as I can still import and use it if I need to.

wat3rBro: I renamed `offload` back to `serialize` to keep BC in the new commit.

I found it difficult to change default value if writing it as `serialize: Union[None, Callable, bool] = NumpySerializedList` or `serialze_method: Callable = NumpySerializedList` without using patch. When using global variable, if I write it as `DefaultSerializeMethod = NumpySerializedList; ... serialze_method: Callable = DefaultSerializeMethod`, the default value is recorded at import time, even later `DefaultSerializeMethod` is changed, it'll still use `NumpySerializedList`. So I ended up still write it as `serialze_method: Optional[Callable] = None` and `None` means use the default value from global variable.

ppwwyyxx: You can do `serialize: Union[Callable, bool] = True`, where `True` means `_DEFAULT_DATASET_FROM_LIST_SERIALIZE_METHOD`. This way the default is still modifyable later.

detectron2/data/common.py
```diff
@@ -106,56 +110,96 @@ def __getitem__(self, idx):
...
-                    len(self._lst)
-                )
-            )
-            self._lst = [_serialize(x) for x in self._lst]
-            self._addr = np.asarray([len(x) for x in self._lst], dtype=np.int64)
-            self._addr = np.cumsum(self._addr)
-            self._lst = np.concatenate(self._lst)
-            logger.info("Serialized dataset takes {:.2f} MiB".format(len(self._lst) / 1024**2))
+            if _serialize_method is None:
+                _serialize_method = _DEFAUL_DATASET_FROM_LIST_SERIALIZE_METHOD
```
ppwwyyxx, line 194: DEFAUL -> DEFAULT

detectron2/data/common.py
```diff
@@ -106,56 +110,101 @@ def __getitem__(self, idx):
                 )
 
 
+class NumpySerializedList(object):
+    """
+    A list-like object whose items are serialized and stored in a Numpy Array. When
+    passing this object to loader workers, they can use shared RAM from master process
+    instead of making a copy.
```
ppwwyyxx, line 117: I think "when passing this object to loader workers" is not clear: if you pass it as a `mp.Processs` argument or through a `mp.Queue`, they will not share memory. Change it to something like:

```
When forking a process that has NumpySerializedList, subprocesses can read the same list without 
triggering copy-on-write, therefore they will share RAM for the list. This avoids the issue in https://github.com/pytorch/pytorch/issues/13246
```

detectron2/data/common.py
```diff
@@ -106,56 +110,101 @@ def __getitem__(self, idx):
...
             copy (bool): whether to deepcopy the element when producing it,
                 so that the result can be modified in place without affecting the
                 source in the list.
-            serialize (bool): whether to hold memory using serialized objects, when
-                enabled, data loader workers can use shared RAM from master
-                process instead of making a copy.
+            serialize (bool or callable): whether to serialize the stroage to other
+                backend. If `serialize` is callable, it will be used to offload th
+            _serialize_method (callable): a callable to serialize original list. If `None` is
+                provided, it'll use the default serialize function.
```
ppwwyyxx, line 186: wrong docs.

ppwwyyxx (approved):
Some docs issue. Otherwise LGTM! Thanks for supporting shared memory.

detectron2/data/common.py
wat3rBro: Isn't the dataset passed as a `mp.Process` argument (https://github.com/pytorch/pytorch/blob/d1ba4c3a6d7a5007665419c57988b06d5b87e96e/torch/utils/data/dataloader.py#L1024)? When I benchmarked the newly added shared memory, I found that now even the `numpy` is not shared by subprocess, but I remembered that it was shared when we first added this.

ppwwyyxx: Yeah my bad. Passing it as a mp.Process argument should work (it's also just a fork). But through a mp.Queue would not work.

ppwwyyxx: Commented about numpy benchmark at https://github.com/facebookresearch/d2go/pull/405#issuecomment-1301510320

PR #4667, opened 2022-11-18
add option to load checkpoints to GPU

YanjunChen329:
Summary:
X-link: https://github.com/facebookresearch/d2go/pull/411

Add config option `cfg.LOAD_CKPT_TO_GPU` to load checkpoints to the worker's current GPU

Previously, D2 (https://github.com/facebookresearch/detectron2/commit/11528ce083dc9ff83ee3a8f9086a1ef54d2a402f)go maps checkpoints to CPU before loading them to the model. In large-scale distributed training, many GPU processes may be used to train a model. This means each process will load the model checkpoint to a single CPU, causing the same model checkpoint to be loaded many times. This would cause CPU OOM issue when the model checkpoint size is large.

There're two solutions to this problem. One is to load checkpoints to GPU; the other one is to use share memory for the checkpoint between different GPU processes. This diff implements the first solution, which can support cases where model size + model checkpoint size is smaller than the total GPU memory. The second solution may be revisited for large models that need to offload checkpoints to cpu. Reference diff: D40789062

Differential Revision: D41063306


ppwwyyxx:
> There're two solutions to this problem. One is to load checkpoints to GPU; the other one is to use share memory for the checkpoint between different GPU processes. This diff implements the first solution, which can support cases where model size + model checkpoint size is smaller than the total GPU memory. The second solution may be revisited for large models that need to offload checkpoints to cpu.

There is a third option that's likely simpler: only load checkpoints if the process's `rank == 0`.
This works because DDP will copy weights from rank0 to all other processes.

ppwwyyxx:
Untested, but something like this might solve it (cc @wat3rBro ):
```diff
diff --git i/detectron2/checkpoint/detection_checkpoint.py w/detectron2/checkpoint/detection_checkpoint.py
index 6620a2b5..a341f6ed 100644
--- i/detectron2/checkpoint/detection_checkpoint.py
+++ w/detectron2/checkpoint/detection_checkpoint.py
@@ -30,8 +30,6 @@ class DetectionCheckpointer(Checkpointer):
         self.path_manager = PathManager

     def load(self, path, *args, **kwargs):
-        need_sync = False
-
         if path and isinstance(self.model, DistributedDataParallel):
             logger = logging.getLogger(__name__)
             path = self.path_manager.get_local_path(path)
@@ -44,16 +42,14 @@ class DetectionCheckpointer(Checkpointer):
                     f"Not all workers can read checkpoint {path}. "
                     "Training may fail to fully resume."
                 )
-                # TODO: broadcast the checkpoint file contents from main
-                # worker, and load from it instead.
-                need_sync = True
-            if not has_file:
-                path = None  # don't load if not readable
-        ret = super().load(path, *args, **kwargs)
-
-        if need_sync:
-            logger.info("Broadcasting model states from main worker ...")
-            self.model._sync_params_and_buffers()
+        if comm.get_rank() == 0:
+            ret = super().load(path, *args, **kwargs)
+        if comm.get_world_size() > 1:
+            if isinstance(self.model, DistributedDataParallel):
+                logger.info("Broadcasting model states from main worker ...")
+                self.model._sync_params_and_buffers()
+            # TODO: must also broadcast ret to all workers
+            pass
         return ret
```

YanjunChen329:
@ppwwyyxx Thanks for the suggestion! This is indeed a nice solution for loading large checkpoints to DDP and I will create another commit for it. There's actually another motivation for GPU loading that I didn't mention. We are integrating Fully Sharded Data Parallel (FSDP) and there would be scenarios where each GPU process loads a different sharded checkpoint. In this case, it might be better to load checkpoints to corresponding GPUs to avoid cluttering CPU memory and I/O

ppwwyyxx:
Sounds good, thanks! 

Maybe the checkpointing logic can be automatically determined from the type of `self.model`? e.g.: if it's (fully-replicated) DDP then load on main worker, and with `map_location` set to `self.model.device_type`; if it's fully-sharded, then each load their own piece and `map_location` set to cuda (or whatever device the model parameters live on).

YanjunChen329:
Sounds great! This way we can save a config key so users don't need to decide what to use. Let me add this logic to the commit. Thanks again for the suggestion :)

YanjunChen329:
Hi @ppwwyyxx I explored the approach you suggested and found some issues with it. Loading checkpoints on rank 0 and syncing model weights would work for only loading pretrained weights, but in the case of resuming training, the other checkpointable states like optimizer/ema/scheduler will be lost. Then I tried to only load checkpoints on rank 0 and broadcast the loaded data dict to other ranks. This still incurs high peak memory utilization because the checkpoint data are copied to different ranks instead of shared among ranks. For now I think the best way to unblock is to directly load checkpoints to GPU. What do you think?

ppwwyyxx:
sounds good!

Issue #3972, opened 2022-02-16
RuntimeError: Default process group has not been initialized, please make sure to call init_process_group.


saidineshpola:
I have tried to train detectron2 using LazyConfig on single GPU but I encountered `File "/home/user/.conda/envs/default/lib/python3.9/site-packages/torch/distributed/distributed_c10d.py", line 358, in _get_default_group
    raise RuntimeError("Default process group has not been initialized") ` . can anybody help what to do here for using it on single machine with out using any DistributedDataParallel from pytorch?

## Instructions To Reproduce the Issue:

1. Full runnable code or full changes you made:
```
!python ./tools/lazyconfig_train_net.py --num-gpus 1 --config-file ./configs/new_baselines/mask_rcnn_regnetx_4gf_dds_FPN_200ep_LSJ.py 

<put code or diff here>
```
2. What exact command you run:
3. __Full logs__ or other relevant observations:
```
[02/16 08:16:26 detectron2]: Rank of current process: 0. World size: 1
[02/16 08:16:27 detectron2]: Environment info:
----------------------  -----------------------------------------------------------------------------------------------
sys.platform            linux
Python                  3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 19:23:11) [GCC 9.4.0]
numpy                   1.22.1
detectron2              0.6 @/home/studio-lab-user/.conda/envs/default/lib/python3.9/site-packages/detectron2
Compiler                GCC 7.3
CUDA compiler           CUDA 11.1
detectron2 arch flags   3.7, 5.0, 5.2, 6.0, 6.1, 7.0, 7.5, 8.0, 8.6
DETECTRON2_ENV_MODULE   <not set>
PyTorch                 1.9.0+cu111 @/home/studio-lab-user/.conda/envs/default/lib/python3.9/site-packages/torch
PyTorch debug build     False
GPU available           Yes
GPU 0                   Tesla T4 (arch=7.5)
Driver version          470.57.02
CUDA_HOME               /usr/local/cuda
Pillow                  9.0.0
torchvision             0.10.0+cu111 @/home/studio-lab-user/.conda/envs/default/lib/python3.9/site-packages/torchvision
torchvision arch flags  3.5, 5.0, 6.0, 7.0, 7.5, 8.0, 8.6
fvcore                  0.1.5.post20211023
iopath                  0.1.9
cv2                     4.5.5
----------------------  -----------------------------------------------------------------------------------------------
PyTorch built with:
[...]

PR #4679, opened 2022-11-26
use non-sync BN if world_size==1

ppwwyyxx:
Workaround for issues like #3972. 

Should be safe to do .. unless someone creates process group after model - but that should be rare (I think.....)


ppwwyyxx:
Not needed given https://github.com/pytorch/pytorch/pull/89706

PR #4684, opened 2022-12-01
fix mmdet detector wrapper init

normster:
MMDet Detector wrapper currently doesn't call init

### DIFF
--- a/detectron2/modeling/mmdet_wrapper.py
+++ b/detectron2/modeling/mmdet_wrapper.py
@@ -145,6 +145,7 @@ class MMDetDetector(nn.Module):
 
             detector = build_detector(_to_container(detector))
         self.detector = detector
+        self.detector.init_weights()
         self.size_divisibility = size_divisibility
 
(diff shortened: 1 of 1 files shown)


detectron2/modeling/mmdet_wrapper.py
```diff
@@ -145,6 +145,7 @@ def __init__(
 
             detector = build_detector(_to_container(detector))
         self.detector = detector
+        self.detector.init_weights()
```
ppwwyyxx, line 148: Is this needed? IIRC weights can be initialized the same way as other detectron2 models through the checkpointer.

normster: This is to call random weight initialization methods, not to load checkpoints: https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/base_module.py#L56

ppwwyyxx: Nice! thanks for the clarification.

ppwwyyxx (approved):

PR #4705, opened 2022-12-15
Remove black from install_requires

malcolmgreaves:
Slim down the set of required dependencies. Removes `black` from `install_requires`: it is already listed as an optional dev dependency for detectron2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -184,7 +184,6 @@ setup(
         "dataclasses; python_version<'3.7'",
         "omegaconf>=2.1",
         "hydra-core>=1.1",
-        "black",
         "timm",
         "packaging",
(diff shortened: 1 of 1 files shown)


malcolmgreaves:
Hi! 👋 Thanks for making detectron2 open source! I noticed that the `black` in the `install_requires` was conflicting with a version that I have set up in my normal development workflow. I would like to know if the maintainers would accept this small patch to slim-down the set of non-optional dependencies.

ppwwyyxx:
>  I noticed that the black in the install_requires was conflicting with a version 

It's listed in `install_requires` without any version specifier, so it shouldn't conflict with any other version.

PR #4709, opened 2022-12-16
update to use torch.optim.lr_scheduler.LRScheduler

wat3rBro:
Summary:
https://github.com/pytorch/pytorch/pull/88503 introduces the public version `LRScheduler`, however `isinstance(self.scheduler, torch.optim.lr_scheduler._LRScheduler)` doesn't work anymore because of https://github.com/pytorch/pytorch/blob/1ea11ecb2eea99eb552603b7cf5fbfc59659832d/torch/optim/lr_scheduler.py#L166-L169, if the schedule is inherited from the public version.

It's a bit tricky to make it BC compatible for torch version <= 1.13. V1 of this diff uses try catch block to import the `LRScheduler` and make it available in `detectron2.solver`, then the whole D2 uses this version of `LRScheduler`. There're two drawbacks though:
- it adds a little mental burden to figure out what's D2 (https://github.com/facebookresearch/detectron2/commit/11528ce083dc9ff83ee3a8f9086a1ef54d2a402f)'s `LRScheduler`, previously it's clear that the `LRScheduler`/`_LRScheduler` is from `torch`.
- it has a name collision with `hooks.LRScheduler`, eg. in the `hooks.py` I have to do `LRScheduler as _LRScheduler`.

But I couldn't found a better solution, maybe use try catch block in every file?

Differential Revision: D42111273

### DIFF
--- a/detectron2/solver/lr_scheduler.py
+++ b/detectron2/solver/lr_scheduler.py
@@ -11,6 +11,11 @@ from fvcore.common.param_scheduler import (
     ParamScheduler,
 )
 
+try:
+    from torch.optim.lr_scheduler import LRScheduler
+except ImportError:
+    from torch.optim.lr_scheduler import _LRScheduler as LRScheduler
+
 logger = logging.getLogger(__name__)
 
 
@@ -52,7 +57,7 @@ class WarmupParamScheduler(CompositeParamScheduler):
         )
 
 
-class LRMultiplier(torch.optim.lr_scheduler._LRScheduler):
+class LRMultiplier(LRScheduler):
     """
     A LRScheduler which uses fvcore :class:`ParamScheduler` to multiply the
     learning rate of each param in the optimizer.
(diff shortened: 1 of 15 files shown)


detectron2/solver/lr_scheduler.py
```diff
@@ -132,7 +137,7 @@ def get_lr(self) -> List[float]:
 # MultiStepLR with WarmupLR but the current LRScheduler design doesn't allow it.
 
 
-class WarmupMultiStepLR(torch.optim.lr_scheduler._LRScheduler):
```
ppwwyyxx, line 135: now it should be fine to remove everything below this line. Looks like they are not used any more.

wat3rBro: @ppwwyyxx There're actually many usage internally (including `_get_warmup_factor_at_iter`), it's a bit hard to remove them.

PR #4712, opened 2022-12-20
fix D2 handler's isfile

wat3rBro:
Differential Revision: D42158024

### DIFF
--- a/detectron2/utils/file_io.py
+++ b/detectron2/utils/file_io.py
@@ -1,5 +1,4 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
-import os
 from urllib.parse import urlparse
 from iopath.common.file_io import HTTPURLHandler, OneDrivePathHandler, PathHandler
 from iopath.common.file_io import PathManager as PathManagerBase
@@ -27,16 +26,11 @@ class Detectron2Handler(PathHandler):
         return [self.PREFIX]
 
     def _get_local_path(self, path, **kwargs):
-        parsed_url = urlparse(path)
-        path = parsed_url._replace(query="").geturl()  # remove query from path
         name = path[len(self.PREFIX) :]
         return PathManager.get_local_path(self.S3_DETECTRON2_PREFIX + name, **kwargs)
 
     def _open(self, path, mode="r", **kwargs):
         return PathManager.open(self._get_local_path(path), mode, **kwargs)
-
-    def _isfile(self, path: str, **kwargs):
-        return os.path.isfile(self._get_local_path(path))
 
 
 PathManager.register_handler(HTTPURLHandler())
--- a/detectron2/checkpoint/detection_checkpoint.py
+++ b/detectron2/checkpoint/detection_checkpoint.py
@@ -29,12 +29,15 @@ class DetectionCheckpointer(Checkpointer):
             **checkpointables,
         )
         self.path_manager = PathManager
+        self._parsed_url_during_load = None
 
     def load(self, path, *args, **kwargs):
+        assert self._parsed_url_during_load is None
         need_sync = False
+        logger = logging.getLogger(__name__)
+        logger.info("[DetectionCheckpointer] Loading from {} ...".format(path))
 
         if path and isinstance(self.model, DistributedDataParallel):
-            logger = logging.getLogger(__name__)
             path = self.path_manager.get_local_path(path)
             has_file = os.path.isfile(path)
             all_has_file = comm.all_gather(has_file)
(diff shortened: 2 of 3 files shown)


detectron2/utils/file_io.py
```diff
@@ -35,10 +35,16 @@ def _get_local_path(self, path, **kwargs):
...
         return PathManager.open(self._get_local_path(path), mode, **kwargs)
 
+
+class Detectron2HTTPURLHandler(HTTPURLHandler):
+    """
+    Allow pathmanager.isfile("http://dl.fbaipublicfiles.com/detectron2/path-to-weight.pkl")
+    """
+
     def _isfile(self, path: str, **kwargs):
         return os.path.isfile(self._get_local_path(path))
```
ppwwyyxx, line 45: This doesn't match the semantics of `isfile`: if the path does not exist, this will throw an error instead of return False. 

In general I'm a bit worried that a user might not expect that a `isfile` call will be expensive and has side effects. For example, a user might check `isfile(image_file)` for every file in a dataset and in that case an explicit error is better than silently downloading the entire dataset.

If this change is for the `isfile` call in fvcore's `checkpointer.load`, one way to avoid it is to let `Detectron2Checkpointer` call `path = get_local_path(path)` before calling `super().load()` . This way fvcore's load will always see a local path. WDYT?

wat3rBro: Yeah I feel it's a bit complicated to rely on `isfile`, overwriting the `load` might be a good option (it also avoids the change in `fvcore`).

wat3rBro: @ppwwyyxx I found it still awkward to do this without relying on PathManager's `isfile`. The original version of `super().load()` requires `PathManager.get_local_path` to return a path that can pass `os.path.isfile` check: https://github.com/facebookresearch/fvcore/blob/be6af6cd8f6214c812514ee4ed201bc5e4fd4e5c/fvcore/common/checkpoint.py#L153-L155, therefore I made the change to `Detectron2Handler._get_local_path`, so that `PathManager.get_local_path("detectron2://model.pth?matching_heuristics=True")` returns `CACHE_DIR/model.pth` instead of `CACHE_DIR/model.pth?matching_heuristics=True`. The change of `Detectron2Handler._get_local_path` allows the `isfile` check to pass, but if we call `path = get_local_path(path)`, the extra information will get lost. On the other hand, if we don't change `Detectron2Handler._get_local_path`, i.e. calling `super().load("CACHE_DIR/model.pth?matching_heuristics=True")`, it also won't work because of the `isfile` check inside `super().load()`. I haven't found a way without touching `fvcore`/`iopath` yet :slightly_frowning_face:, any suggestions?

ppwwyyxx: Thanks for looking into this. A path like `CACHE_DIR/model.pth?matching_heuristics=True` does feel less reliable since there is no guarantee that fvcore/iopath will always handle such paths as expected even if they do today.
Since this new syntax only exists in detectron2, it's better to handle it without touching fvcore/iopath.

One idea:

`DetectionCheckpointer` will be responsible for calling `get_local_path` and save a mapping of `{local_path: query params}` (the key should be normalized by `os.path.realpath & os.path.normpath` for uniqueness). Then this mapping will be read in `_load_file`. 

(BTW, I noticed that there is a `if ddp` codepath in `DetectionCheckpointer` that looks currently broken due to `get_local_path`.)

wat3rBro: @ppwwyyxx The query params can't be passed via `.load` or `._load_file` unless we extend `fvcore`, to overcome this, do you mean saving a persistent mapping in the `DetectionCheckpointer` object? But what happens if we load two checkpoints (say `"detectron2://model.pth?matching_heuristics=True"` and `"detectron2://model.pth"`) using the same checkpointer instance? How is it different from using a temporary variable to record the query params using during `.load`?

ppwwyyxx: I meant to save the query param as an attribute inside the object.
Using a temporary attribute (only available during `.load()`) sounds like a better idea as it can handle loading the same path with two different queries (probably not useful, but technically it can happen).

ppwwyyxx (approved):
ic, this already supports loading the same file with different queries.
Using a temporary attribute (that's deleted at the end of `load()`) could be cleaner, but the current status looks great as well.

wat3rBro:
I'll go with using temporary attribute since mapping costs more RAM.

wat3rBro:
add some more tests

PR #4713, opened 2022-12-20
np.bool -> np.bool_

wat3rBro:
Summary: numpy 1.24 removed the np.bool

Differential Revision: D42159481


ppwwyyxx:
`np.bool -> bool` should also work and is simpler

PR #4720, opened 2022-12-23
Support launch() on CPU machines.

ppwwyyxx:


wat3rBro (approved):
we added `backend` as a parameter to `launch` in case user wants to use CPU backend when GPU is available, probably not something common for D2 users though.

wat3rBro (approved):

wat3rBro:
@facebook-github-bot shipit

PR #4721, opened 2022-12-24
NumpySerializedList -> _TorchSerializedList

ppwwyyxx:
I recently did a complete study of the RAM sharing strategies: https://ppwwyyxx.com/blog/2022/Demystify-RAM-Usage-in-Multiprocess-DataLoader.

It shows that torch.Tensor should be preferred over numpy arrays so that sharing can work under spawn/forkserver start_method.

Although detectron2 never explicitly sets the start_method, it's common that some one will set a global start_method. So this change might reduce RAM usage.

(I also provide [another implementation](https://github.com/ppwwyyxx/RAM-multiprocess-dataloader/blob/bb8f67fbd21e1acc2301d722e6955b8178b51fad/serialize.py#L55-L75) of cross-GPU sharing that's simpler than https://github.com/facebookresearch/mobile-vision/pull/120. Feel free to try it out and include it in the future).

### DIFF
--- a/detectron2/data/common.py
+++ b/detectron2/data/common.py
@@ -7,6 +7,7 @@ import numpy as np
 import pickle
 import random
 from typing import Callable, Union
+import torch
 import torch.utils.data as data
 from torch.utils.data.sampler import Sampler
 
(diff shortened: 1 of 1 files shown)


wat3rBro:
Thanks @ppwwyyxx for the really nice write up, especially for the scientific RAM usage measurement tools.

PR #4735, opened 2023-01-03
Color tuple forced to stay within 0.0 - 1.0 range

kanishkanarch:
Earlier, values in the `color` tuple were being greater than 1.0:
image
This was leading to the following error:
image
This minor fix did the trick!

### DIFF
--- a/detectron2/utils/visualizer.py
+++ b/detectron2/utils/visualizer.py
@@ -875,6 +875,9 @@ class Visualizer:
             font_size = self._default_font_size
 
         # since the text background is dark, we don't want the text to be dark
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
Same comment applies: https://github.com/facebookresearch/detectron2/pull/4500#discussion_r954387438

kanishkanarch:
Turns out this is a [Matplotlib error](https://github.com/matplotlib/matplotlib/blob/6091437be9776139d3672cde28a19cbe6c09dcd5/lib/matplotlib/colors.py#L394):
image

PR #4749, opened 2023-01-21
Remove unused onnx.optimizer

kit1980:
onnx.optimizer import is currently unused in this file. Optimizer API was recently removed from ONNX, so this unused import blocks migration to newer ONNX versions.


malfet (approved):
LGTM, though it was not removed recently, but actually a while back

PR #4754, opened 2023-01-24
Add gradient accumulation to trainers

fardinayar:
Hi
Recent deep-learning papers use multiple GPUs for training, which is unavailable for all users. Adding gradient accumulation to Detectron2 helps reproduce results better.

### DIFF
--- a/detectron2/engine/train_loop.py
+++ b/detectron2/engine/train_loop.py
@@ -100,6 +100,8 @@ class TrainerBase:
             By convention the minimum possible value is 0.
 
         max_iter(int): The iteration to end training.
+        
+        accumulate_grad_batches(int): Accumulates grads every k batches. default is 1: No gradient accumulation.
 
         storage(EventStorage): An EventStorage that's opened during the course of training.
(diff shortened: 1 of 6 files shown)


detectron2/engine/train_loop.py
```diff
@@ -279,20 +285,24 @@ def run_step(self):
             losses = sum(loss_dict.values())
 
         """
-        If you need to accumulate gradients or do something similar, you can
-        wrap the optimizer with your custom `zero_grad()` method.
```
ppwwyyxx, line 283: As the comment says, gradient accumulation should be implemented with a custom optimizer.

Erotemic: Gradient accumulation is a pretty fundamental option. It would be nice to have it available out of the box. PytorchLightning does provide this, is there a reason detectron2 can't?

wishforgood: Would you please show a working example? I tried to do this but failed to make it work.

PR #4771, opened 2023-02-02
fix(projects): missing abc when using python 3.10

FateScript:
This PR is a small  fix of `AttributeError` when using python3.10
User could reproduce by using the following code
```
import importlib
type(importlib.abc.MetaPathFinder)
```
PTAL @ppwwyyxx

### DIFF
--- a/detectron2/projects/__init__.py
+++ b/detectron2/projects/__init__.py
@@ -1,5 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
-import importlib
+from importlib import abc, util
 from pathlib import Path
 
 _PROJECTS = {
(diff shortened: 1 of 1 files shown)


ppwwyyxx (approved):

detectron2/projects/__init__.py
```diff
@@ -1,5 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
-import importlib
+from importlib import abc, util
```
wat3rBro, line 2: `abc` and `util` can confused with other modules, it's better to avoid using them without package name.
```suggestion
import importlib.abc
import importlib.util
# or from importlib.abc import MetaPathFinder, ...
```

ppwwyyxx: Yeah; related to this, adding a `__all__ = []` to the file would be helpful as well

FateScript: Fixed this as your suggestion. PTAL.

wat3rBro (approved):

PR #4773, opened 2023-02-02
Let Detectron2Handler.open open the URL directly without downloading

ppwwyyxx:
This is currently a no-op, because `S3_DETECTRON2_PREFIX` is on http and `open()` of http URL will download anyway.

However, our internal version uses an env module (similar to `detecton2.fb.env:setup_environment()`) that sets `S3_DETECTRON2_PREFIX` to something not on http. In that case this change would help.


wat3rBro:
ah lint...

ppwwyyxx:
.. fixed

PR #4811, opened 2023-02-21
added indexing to meshgrid in anchor_generator

yiftachbeer:
Otherwise a torch gives the following warning (will need to specify it in the future because the default is going to change).

Thanks for your contribution!

### DIFF
--- a/detectron2/modeling/anchor_generator.py
+++ b/detectron2/modeling/anchor_generator.py
@@ -49,7 +49,7 @@ def _create_grid_offsets(
         target_device_tensor,
     )
 
-    shift_y, shift_x = torch.meshgrid(shifts_y, shifts_x)
+    shift_y, shift_x = torch.meshgrid(shifts_y, shifts_x, indexing='ij')
     shift_x = shift_x.reshape(-1)
     shift_y = shift_y.reshape(-1)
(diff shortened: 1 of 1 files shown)


ppwwyyxx (approved):

yiftachbeer:
Apparently, the "indexing" feature was only added to pytorch in 1.10 (so it cannot be added on previous versions, but is required to be added in later versions to not raise a warning).
How are such version differences usually handled in detectron2?

ppwwyyxx:
Typically that's done by ending support for older versions. 

I remember there were some PRs upgrading the pytorch versions in the CI. This fix probably has to come after those.

wat3rBro:
@yiftachbeer If you would like to merge this PR before upgrading the PyTorch versions in CI, I'm fine with disabling the affected tests for 1.8 & 1.9 before that.

yiftachbeer:
> @yiftachbeer If you would like to merge this PR before upgrading the PyTorch versions in CI, I'm fine with disabling the affected tests for 1.8 & 1.9 before that.

Sure, how do I do that?

wat3rBro:
@yiftachbeer something similar to https://fburl.com/code/cmcwnx96

PR #4840, opened 2023-03-02
allow print non loss metrics.

newstzpz:
Summary:
allow print non loss metrics.
* If the metric name contains `[metric]`.
* Non-loss metrics have to be passed as numbers (not tensors) and they will not be used to compute the overall loss.

Differential Revision: D43531600

### DIFF
--- a/detectron2/utils/events.py
+++ b/detectron2/utils/events.py
@@ -260,7 +260,7 @@ class CommonMetricPrinter(EventWriter):
         # NOTE: max_mem is parsed by grep in "dev/parse_results.sh"
         self.logger.info(
             str.format(
-                " {eta}iter: {iter}  {losses}  {avg_time}{last_time}"
+                " {eta}iter: {iter}  {losses}  {non_losses}  {avg_time}{last_time}"
                 + "{avg_data_time}{last_data_time} lr: {lr}  {memory}",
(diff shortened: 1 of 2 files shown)


detectron2/engine/train_loop.py
```diff
@@ -325,7 +325,7 @@ def _write_metrics(
 
     @staticmethod
     def write_metrics(
-        loss_dict: Mapping[str, torch.Tensor],
+        loss_dict: Mapping[str, Union[torch.Tensor, float]],
```
ppwwyyxx, line 328: You can already write metrics from inside the model like this: https://github.com/facebookresearch/detectron2/blob/80e2673da161f57afe37ef769836a61976108ef1/detectron2/modeling/roi_heads/fast_rcnn.py#L114
With that, the trainer does not need to be changed to include "non-loss" in "loss dict".

newstzpz: Thanks, this works. But one issue with this is that it will only show item that contains the name "loss", but some of the items we want to print are not actually losses. Do you have a recommend way to support this?

ppwwyyxx: For that I think it's good to change MetricsPrinter (e.g. add an argument or hard code "[metrics]"). But no need to change `train_loop.py` then.

PR #4850, opened 2023-03-09
Fixed error in matplotlib when color > 1.0 due to numerical errors

Nerei:
Fixed rare error below that comes from matplotlib. In my case color was 1.001.

```ValueError: RGBA values should be within 0-1 range```

### CODE REVIEW

File: detectron2/utils/visualizer.py [hunk 1]
```diff
@@ -726,6 +726,7 @@ def overlay_instances(
 
                 height_ratio = (y1 - y0) / np.sqrt(self.output.height * self.output.width)
                 lighter_color = self._change_color_brightness(color, brightness_factor=0.7)
+                lighter_color = tuple(np.clip(lighter_color, 0.0, 1.0))
```
ppwwyyxx, line 729: Thanks, could you add it into _change_color_brightness instead?

ppwwyyxx (approved):

detectron2/utils/visualizer.py
Nerei: Done. Also signed CLA

ppwwyyxx (approved):

PR #4903, opened 2023-04-09
Allow async write metrics to speed up training

tglik:
Summary:
This change is a follow up on D43098985 which reduce the frequency on metrics collection calls due to the issue that its involve gather logic which slows down training.
The downside of that solution is that it impacts the granularity of metrics and loss graphs which reduce visablity and debuggibility of the training progress.

This diff provide a solution that will avoid the slowness of gather logic in the training loop while keep the fine granulariy of the metrics and graphs.
The solution moves the write_metrics call (that include the copy from gpu to cpu, the gather, and the storate event write) to be async in a separate thread.
We introduce a thread pool executor with a single worker (thread) that will execute all the logic that will be submitted to its queue in order.

Reviewed By: sf-wind

Differential Revision: D44305165

### DIFF
--- a/detectron2/engine/train_loop.py
+++ b/detectron2/engine/train_loop.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright (c) Facebook, Inc. and its affiliates.
-
+import concurrent.futures
 import logging
 import numpy as np
 import time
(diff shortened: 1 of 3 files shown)


stefan-it:
Hi @tglik and @danielm322,

this unfortunately introduces breaking change in PyTorch Lightning example, where `write_metrices()` from `SimpleTrainer` is called with two arguments (and `cur_iter` is then missing). I could fix that by passing `self.storage.iter`.

PR #4913, opened 2023-04-16
use threshold 0.5 to compute mask accuracy

stephenyan1231:
Summary: Use threshold 0.5 (vs 0) to compute mask accuracy. This will not affect the actual model training and eval, but improves the logging in the event storage.

Differential Revision: D45027814


ppwwyyxx:
this is before sigmoid, so 0.0 is correct.

PR #4938, opened 2023-05-08
Support Moduleted Deform Conv on CPU via torchvision

Jmennius:
Just like #2258 but for modulated deformable convolutions - let `torchvision` do the heavylifting.
Support for modulated deform conv in `torchvision` exists since `v0.9.0`, so it doesn't seem to influence dependency requirements at the moment.
Updated tests to let most of them run on CPU and extended other tests to cover modulated variant as well.


Jmennius:
Builds fail due to `protobuf` incompatibility with `pytorch` (too recent protobuf is installed for proto files in pytorch).
It should be pinned to `~=3.20` (or `<4`). For now it seems that pytorch doesn't update to support `4.*` python protobuf (pytorch/pytorch#78362 and protobuf versions are pinned in CI).

PR #4956, opened 2023-05-16
Distributed training with pytorch lightning discards data

CJcool06:
I noticed that, when using detectron2 with pytorch lightning (2.0.2) to perform lightning's distributed data parallel training (4 gpus), the `build_detection_train_loader` and `build_detection_test_loader` functions would build a dataloader that would skip half of the dataset.

I found the problem to be due to dataset sharding. The detectron2 distributed samplers first perform their sharding during init, then the pytorch dataloader would _also_ perform sharding downstream, thus essentially discarding a portion of the data (proportional to the number of gpus).

With the current state of the functions, there isn't a simple way to use something other than the implemented distributed samplers. The only way to do this would be foregoing the functions themselves and re-implementing them in project code with only minor changes. My proposed changes maintain the original integrity of the functions and allow the use of pytorch sampler implementations.

It's my understanding that pytorch lightning is not officially supported with detectron2 as of yet (beta according to [tools/lightning_train_net.py](https://github.com/facebookresearch/detectron2/blob/main/tools/lightning_train_net.py)). This would be a good and simple step toward that direction.


ppwwyyxx:
It is my understanding that pytorch dataloader does not do any sharding, other than what's done in the sampler. Therefore `build_detection_train_loader` is correct and no need to change it. 

Maybe lightning has some weird behaviors, but I don't think it's worthwhile to change a correct function to adapt to it.

PR #4975, opened 2023-05-25
Replace numpy aliases since they are deprecated

stefanklut:
AttributeError: module 'numpy' has no attribute 'int'.
`np.int` was a deprecated alias for the builtin `int`. To avoid this error in existing code, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations. Did you mean: 'inf'?

### DIFF
--- a/detectron2/data/build.py
+++ b/detectron2/data/build.py
@@ -169,11 +169,11 @@ def print_instances_class_histogram(dataset_dicts, class_names):
     """
     num_classes = len(class_names)
     hist_bins = np.arange(num_classes + 1)
-    histogram = np.zeros((num_classes,), dtype=np.int)
+    histogram = np.zeros((num_classes,), dtype=np.int32)
     for entry in dataset_dicts:
         annos = entry["annotations"]
         classes = np.asarray(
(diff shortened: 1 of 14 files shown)


ppwwyyxx:
> To avoid this error in existing code, use int by itself. Doing this will not modify any behavior and is safe. 

Please just use `int` and `float` directly.
There is also `np.bool` that should be changed to `bool`.

stefanklut:
> Please just use `int` and `float` directly. There is also `np.bool` that should be changed to `bool`.

Implemented this suggestion. Replace all aliases I could find from this list https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations. I did not find any use of np.bool

ppwwyyxx (approved):

PR #4986, opened 2023-06-01
Fix is_compiling import error

hasbegun:
Fix is_compiling import error from torch._dynamo.
Summary: It seems like Torch 2.0 doesn't have is_compiling module and it causes an import error.


jmariscalharana (approved):

wat3rBro:
Already merged as part of in https://github.com/facebookresearch/detectron2/pull/5274.

Issue #3983, opened 2022-02-18
Depreciation warning when running inference w/Pytorch 1.10.1


rtarquini:
## Details

detectron2/detectron2/structures/image_list.py:88: UserWarning: __floordiv__ is deprecated, and its behavior will change in a future version of pytorch. It currently rounds toward 0 (like the 'trunc' function NOT 'floor'). This results in incorrect rounding for negative values. To keep the current behavior, use torch.div(a, b, rounding_mode='trunc'), or for actual floor division, use torch.div(a, b, rounding_mode='floor').
  max_size = (max_size + (stride - 1)) // stride * stride

## Environment:

----------------------  ----------------------------------------------------------------
sys.platform            linux
Python                  3.9.6 (default, Feb 18 2022, 13:15:09) [GCC 7.5.0]
numpy                   1.22.1
detectron2              0.6 @/opt/thirdparty/detectron2/detectron2
Compiler                GCC 7.5
CUDA compiler           CUDA 10.2
detectron2 arch flags   5.0, 6.0, 6.1, 7.0, 7.5
DETECTRON2_ENV_MODULE   <not set>
PyTorch                 1.10.1+cu102 @/usr/local/lib/python3.9/site-packages/torch
PyTorch debug build     False
GPU available           Yes
GPU 0,1,2,3             NVIDIA GeForce GTX 1080 (arch=6.1)
Driver version          470.103.01
CUDA_HOME               /usr/local/cuda
TORCH_CUDA_ARCH_LIST    5.0;6.0;6.1;7.0;7.5
Pillow                  9.0.1
torchvision             0.11.2+cu102 @/usr/local/lib/python3.9/site-packages/torchvision
torchvision arch flags  3.5, 5.0, 6.0, 7.0, 7.5
fvcore                  0.1.5.post20220212
iopath                  0.1.9
cv2                     4.4.0
----------------------  ----------------------------------------------------------------
PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) Math Kernel Library Version 2020.0.0 Product Build 20191122 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v2.2.3 (Git Hash 7336ca9f055cf1bfa13efb658fe15dc9b41f0740)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - LAPACK is enabled (usually provided by MKL)
  - NNPACK is enabled
  - CPU capability usage: AVX2
[...]

PR #5004, opened 2023-06-19
Remove pytorch floordiv deprecation messages 

picturedots:
This change should remove the following pytorch deprecation message:
```
detectron2/structures/keypoints.py:220: UserWarning: __floordiv__ is deprecated, and its behavior will change in a future version of pytorch. It currently rounds toward 0 (like the 'trunc' function NOT 'floor'). This results in incorrect rounding for negative values. To keep the current behavior, use torch.div(a, b, rounding_mode='trunc'), or for actual floor division, use torch.div(a, b, rounding_mode='floor'). 
```

This change is presuming that the value of `y_int` is positive, hence of use of `rounding_method="floor" ` instead of `rounding_method="trunc"`

This specific fix is mentioned in https://github.com/facebookresearch/detectron2/issues/3983


SergiyShebotnov:
Will this be still fixed?

Issue #1296, opened 2020-04-26
Getting the output metrics from Detectron2's COCOEvaluator


kierangilliam:
## ❓ How to use Detectron2

The metrics from standard out are much more useful than the outputs written in the output folder.

`Standard out`

[![ ][1]][1]

`metrics.json` and `coco_instances_results.json`

[![enter image description here][2]][2]

    def test(instance, cfg, trainer, test_instance):  
      cfg.MODEL.WEIGHTS = os.path.join(cfg.OUTPUT_DIR, "model_final.pth")
      cfg.MODEL.ROI_HEADS.SCORE_THRESH_TEST = 0.7   # TODO What is this
      cfg.DATASETS.TEST = (test_instance, )
      predictor = DefaultPredictor(cfg)
      evaluator = COCOEvaluator(test_instance, cfg, False, output_dir=cfg.OUTPUT_DIR)
      val_loader = build_detection_test_loader(cfg, test_instance)
      result = inference_on_dataset(trainer.model, val_loader, evaluator)

How can I get the details like the Average Percision for ioU=0.5 or per-category bbox AP?

  [1]: https://i.stack.imgur.com/xtQ7t.png
  [2]: https://i.stack.imgur.com/3N4qC.png

ppwwyyxx:
> How can I get the details like the Average Percision for ioU=0.5 or per-category bbox AP?

Add `print(result)` to your code or save it to a file.

kierangilliam:
> > How can I get the details like the Average Percision for ioU=0.5 or per-category bbox AP?
> 
> Add `print(result)` to the code in your post or save it to a file.

This is the output of printing result

`OrderedDict([('bbox', {'AP': 0.5943767883084139, 'AP50': 1.4017789571077026, 'AP75': 0.4708867402112728, 'APs': 0.46667355809273686, 'APm': 1.2510921596135547, 'APl': 0.5088786109768504, 'AP-tomato_fruit_unripe': 1.1100818533272692, 'AP-tomato_fruit': 1.8462554691484383, 'AP-tomato_seedling': 0.01276231363919374, 'AP-tomato_young_plant': 0.0027843054271681393, 'AP-tomato_flower': 0.0})])`

It contains the evaluation results for bbox and per-category bbox AP, but does not contain the AR, or any of the stats from above the line "Evaluation results for bbox".

ppwwyyxx:
> does not contain the AR,

[...]

PR #5175, opened 2023-12-06
Coco evaluator ar

carandraug:
This PR adds AR metrics to the results of `COCOEvaluator`.  This was requested on the discussion of issue #1296 to which the reply was to submit a PR:

_Originally posted by @ppwwyyxx in https://github.com/facebookresearch/detectron2/issues/1296#issuecomment-619626547_:
>> does not contain the AR,
>
> AR is not commonly used, and is therefore not included in results. You can send a PR to add it.
>
>> or any of the stats from above the line "Evaluation results for bbox".
>
> everything except for AR are included.

Here's that PR.

I split this into two commits. The first commit adds the "global" AR metrics only (effectively the value displayed by `COCOEval.summarize()`). The second commit adds the per-class AR.

### DIFF
--- a/detectron2/evaluation/coco_evaluation.py
+++ b/detectron2/evaluation/coco_evaluation.py
@@ -326,18 +326,54 @@ class COCOEvaluator(DatasetEvaluator):
 
         Args:
             coco_eval (None or COCOEval): None represents no predictions from model.
-            iou_type (str):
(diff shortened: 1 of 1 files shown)


carandraug:
The failed checks appear to be due to some issues with the CI configuration (indeed the last pushed commit that passed all the checks was 9 months ago)

Issue #5249, opened 2024-04-03
🐛 Minor Bug: PointsVisualizer() throws error when passed floating coordinate values 


soumya997:
1. **Issue:**
I can't share my code but below is a minimal code sample to reproduce the error. My code calls the function `DensePoseDataPointsVisualizer()` and internally it calls `PointsVisualizer()`, at the 108 line `cv2.circle()` gets called, but as the centre coordinates passed in that function is in float, cv2 throws an error, as `cv2.circle()` expects integer pixel coordinates as centre. 

_Here is the [google colab notebbok](https://colab.research.google.com/drive/1AWetIKwwNyrztHMQciT53YkW3-dJQnkM?usp=sharing) that reproduces the error._

```python3
height, width = 400, 400
radius = 50
image = np.random.randint(0, 256, (height, width, 3), dtype=np.uint8)
centers = [(100.09, 100.5), (300.6, 300.89)]

viz_func = PointsVisualizer()
viz_func.visualize(image, centers)
```

2. **Error Log:**
```
---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
[<ipython-input-3-fe6e38084d89>](https://localhost:8080/#) in <cell line: 2>()
      1 viz_func = PointsVisualizer()
----> 2 viz_func.visualize(image, centers)

[/content/detectron2/projects/DensePose/densepose/vis/base.py](https://localhost:8080/#) in visualize(self, image_bgr, pts_xy, colors_bgr, rs)
    106             color_bgr = colors_bgr[j] if colors_bgr is not None else self.color_bgr
    107             r = rs[j] if rs is not None else self.r
--> 108             cv2.circle(image_bgr, (x, y), r, color_bgr, -1)
    109         return image_bgr
    110 

error: OpenCV(4.8.0) :-1: error: (-5:Bad argument) in function 'circle'
> Overload resolution failed:
>  - Can't parse 'center'. Sequence item with index 0 has a wrong type
>  - Can't parse 'center'. Sequence item with index 0 has a wrong type
```

3. **Changes Made:**
changed,
`x, y = pt_xy`
to 
`x, y = [int(val) for val in pt_xy]`

https://github.com/soumya997/detectron2/blob/0f3f9920d3e98e4db7442c19262d5f0eade25581/projects/DensePose/densepose/vis/base.py#L95C1-L109C25

4. **output of `git rev-parse HEAD; git diff`**
`0f3f9920d3e98e4db7442c19262d5f0eade25581`

PR #5250, opened 2024-04-03
minor fix: made centre coordinates integer for cv2.circle()

soumya997:
PR regarding the issue #5249 

Its a minor error, please look into it and let me if you if I need to make any change.
There is deliverable that stuck because of this issue. If this gets merged it would be very helpful.


soumya997:
Hi @ppwwyyxx can you please look into this PR.

Issue #5228, opened 2024-03-02
Detectron doesn't include torch as a formal dependency


mdvthu:
## Instructions To Reproduce the 🐛 Bug:
1. Full runnable code or full changes you made:
```
cd new-python-project-requiring-detectron2
echo 'detectron2 @ git+https://github.com/facebookresearch/detectron2.git#egg=detectron2@v0.6' >> requirements.txt
python3.10 -m venv venv
. venv/bin/activate
python -m pip install -r requirements.txt
```

The above will fail as torch is not already available and not installed as a dependency

## Desired behavior:

If `detectron2` included torch as a stated dependency, it would make fresh install more straightforward.  It currently crashes without a useful error message due to the `import torch` statement in `setup.py`

This limits the usefulness of using clean virtual environments (with repeatable requirements.txt) as a manual install of pytorch is required first

Programmer-RD-AI:
An PR #5296 has been setup :)

PR #5296, opened 2024-05-27
Fix torch dependency requirement and code formatting in setup.py

Programmer-RD-AI:
This PR addresses issue #5228 reported by @mdvthu. The changes ensure that the torch dependency requirement in the setup.py file of Detectron2 is correctly specified, aligning with actual version compatibility. Additionally, the code formatting and readability in the setup.py file have been improved for better consistency.

Before submitting this PR, I have run `dev/linter.sh` to ensure that the code adheres to the project's linting standards.

Thank you for considering this contribution!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,9 @@ assert torch_ver >= [1, 8], "Requires PyTorch >= 1.8"
 
 
 def get_version():
-    init_py_path = path.join(path.abspath(path.dirname(__file__)), "detectron2", "__init__.py")
+    init_py_path = path.join(
+        path.abspath(path.dirname(__file__)), "detectron2", "__init__.py"
+    )
     init_py = open(init_py_path, "r").readlines()
     version_line = [l.strip() for l in init_py if l.startswith("__version__")][0]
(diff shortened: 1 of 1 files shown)


sliedes:
torch seems to also be a build requirement. Adding it in setup.py as a dependency still doesn't seem to fix everything.

I got detectron2 to install with `poetry add $url` by just creating a `pyproject.toml ` with these contents:

```toml
[build-system]
requires = ["setuptools", "torch"]
```

Though in addition to that it should be added in `setup.py` like this PR does.

mostlyuseful:
> torch seems to also be a build requirement. Adding it in setup.py as a dependency still doesn't seem to fix everything.
> 
> I got detectron2 to install with `poetry add $url` by just creating a `pyproject.toml ` with these contents:
> 
> ```toml
> [build-system]
> requires = ["setuptools", "torch"]
> ```
> 
> Though in addition to that it should be added in `setup.py` like this PR does.

Tried it in a [fork](https://github.com/mostlyuseful/detectron2), that was the missing piece for my setup. Thanks so much!

PR #5310, opened 2024-06-17
Installation fails with recent version of setuptools, document working version

Telcrome:
Installation fails with setuptools>=70.0.0 with ImportError about packaging, which cannot be imported from pkg_resources

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,7 +1,7 @@
 ## Installation
 
 ### Requirements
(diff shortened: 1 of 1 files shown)


Programmer-RD-AI (changes requested):
Hi,
Adding these setuptools to the `setup.py` would help automate the process and will ensure that the recent version of setuptools will be installed...
Thank you

PR #5330, opened 2024-07-20
Update `detectron2/engine/defaults.py` to make `DefaultPredictor` able to accept batch predictions

irhawks:
Modify  `DefaultPredictor` 's `__call__` method to accept list of `original_image` at the same time keep this API compatible with previous code. (already checked with `dev/linter.sh`)

### DIFF
--- a/detectron2/engine/defaults.py
+++ b/detectron2/engine/defaults.py
@@ -16,6 +16,7 @@ import sys
 import weakref
 from collections import OrderedDict
 from typing import Optional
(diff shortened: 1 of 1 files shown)


Programmer-RD-AI (approved):

ywuenthought:
Does Detectron support real batch under the hood? Yes, you are feeding a batch of images to the predictor, but technically speaking, it still performs prediction one by one?

PR #5348, opened 2024-08-14
fix accuracy test errors

wat3rBro:
Summary:
Some accuracy tests started to fail in between Jun 11 and Jun 17:
- ❌ mask_rcnn_R_50_FPN_inference_acc_test
- ✅ keypoint_rcnn_R_50_FPN_inference_acc_test
- ✅ fast_rcnn_R_50_FPN_inference_acc_test
- ❌ panoptic_fpn_R_50_inference_acc_test
- ✅ retinanet_R_50_FPN_inference_acc_test
- ❌ rpn_R_50_FPN_inference_acc_test
- ✅ semantic_R_50_FPN_inference_acc_test
- ❌ cascade_mask_rcnn_R_50_FPN_inference_acc_test

V1: update the yaml to reflect the new scores.

Differential Revision: D61301698

### DIFF
--- a/configs/quick_schedules/cascade_mask_rcnn_R_50_FPN_inference_acc_test.yaml
+++ b/configs/quick_schedules/cascade_mask_rcnn_R_50_FPN_inference_acc_test.yaml
@@ -5,3 +5,4 @@ DATASETS:
   TEST: ("coco_2017_val_100",)
 TEST:
   EXPECTED_RESULTS: [["bbox", "AP", 50.18, 0.02], ["segm", "AP",  43.87, 0.02]]
+FLOAT32_PRECISION: "highest"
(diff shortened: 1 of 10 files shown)


wat3rBro:
@ppwwyyxx do you know if there were changes recently that lead to difference in accuracy metrics?

ppwwyyxx:
When did this start to happen?

It could also be a change of CUDA version / precision. For example, are these now running on Ampere cards with TF32 enabled?

wat3rBro:
@ppwwyyxx Great call! The last successful run (in June) was on Volta, and the first failed run was on Ampere. Setting `torch.backends.cudnn.allow_tf32` to `False` makes the accuracy tests pass (`torch.backends.cuda.matmul.allow_tf32` is `False` by default).

PR #5500, opened 2025-11-15
Small fix about refcycle

ppwwyyxx:
The comment about `__del__` applies to Python 2 and is no longer relevant.


ppwwyyxx:
@wat3rBro could you help review this?

PR #5538, opened 2026-08-19
Preserve holes in instance mask overlays

### LINKED ISSUE #1771: Potential bug on drawing segmentation mask with holes

tandede:
## Summary

- preserve holes when `overlay_instances` renders binary instance masks
- reuse the existing raster mask path only for masks with holes, while retaining polygon rendering for regular masks
- add a pixel-level regression test that verifies the mask is colored and its hole remains unchanged

## Root cause

`GenericMask` returns both outer and inner contours for masks with holes. `overlay_instances` drew every contour as an independently filled polygon, so the inner contour filled the hole with the instance color. Dropping the inner contour would not be sufficient because the filled outer polygon still covers the hole.

This change routes masks with holes through `draw_binary_mask`, which already represents holes as transparent pixels. Masks without holes continue to use the existing polygon path.

Fixes #1771.

## Validation

- `MPLBACKEND=Agg python -m pytest tests/test_visualizer.py -q` (15 passed)
- `flake8 .`
- `isort -c -sp .`
- `black -l 100 --check detectron2/utils/visualizer.py tests/test_visualizer.py`
- `git diff --check`


tandede:
Hi maintainers, I’m following up on this PR in case it was missed. The Meta CLA and import checks have passed, while the Import Status check has remained queued. The change is ready for review, and I’d be happy to address any feedback. Thank you!

PR #5540, opened 2026-08-30
Preserve holes in instance masks

Marchematics:
Preserve holes in instance masks.


Marchematics:
Avoided duplicate contour extraction for masks with holes and made the regression test deterministic.

